PR #253, opened 2015-01-01
fix #216, support large jars. host a embedded server to store large jars...

### LINKED ISSUE #216: Support large jars
Currently,

We are using akka messages to pass the jars, there maybe size limitation.
We need to find a robust way to distribute the jars.

clockfly:
..., also allow to use distributed file system to store jars so that when appmaster and master both crashes, we can still recover.

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/253)

### DIFF
--- a/core/src/main/scala/org/apache/gearpump/cluster/Master.scala
+++ b/core/src/main/scala/org/apache/gearpump/cluster/Master.scala
@@ -18,6 +18,7 @@
 
 package org.apache.gearpump.cluster
 
+import akka.actor.Actor.Receive
 import akka.actor._
 import akka.remote.DisassociatedEvent
 import com.typesafe.config.Config
@@ -28,7 +29,9 @@ import org.apache.gearpump.cluster.MasterToAppMaster._
 import org.apache.gearpump.cluster.MasterToWorker._
 import org.apache.gearpump.cluster.WorkerToMaster._
 import org.apache.gearpump.cluster.scheduler.Scheduler.ApplicationFinished
+import org.apache.gearpump.jarstore.{JarStore, JarFileContainer}
 import org.apache.gearpump.util.{LogUtil, Util, ActorUtil, Constants}
+import org.apache.hadoop.fs.Path
 import org.slf4j.{Logger, LoggerFactory}
 
 import scala.annotation.tailrec
@@ -50,9 +53,12 @@ private[cluster] class Master extends Actor with Stash {
 
   LOG.info("master is started at " + ActorUtil.getFullPath(context) + "...")
 
+  val jarStore = context.actorOf(JarStore.props(systemConfig.getString(Constants.GEAR_APP_JAR_STORE_ROOT_PATH)))
+
   override def receive : Receive = workerMsgHandler orElse
     appMasterMsgHandler orElse
     clientMsgHandler orElse
+    jarStoreService orElse
     terminationWatch orElse
     disassociated orElse
     ActorUtil.defaultMsgHandler(self)
@@ -76,6 +82,11 @@ private[cluster] class Master extends Actor with Stash {
       LOG.info(s"Register Worker $id....")
     case resourceUpdate : ResourceUpdate =>
       scheduler forward resourceUpdate
+  }
+
+  def jarStoreService : Receive = {
+    case GetJarFileContainer =>
+      jarStore forward GetJarFileContainer
   }
 
   def appMasterMsgHandler : Receive = {
@@ -147,4 +158,5 @@ private[cluster] class Master extends Actor with Stash {
 
 object Master{
   case class WorkerTerminated(workerId : Int)
+
 }
--- a/core/src/main/scala/org/apache/gearpump/cluster/client/ClientContext.scala
+++ b/core/src/main/scala/org/apache/gearpump/cluster/client/ClientContext.scala
@@ -23,11 +23,16 @@ import java.util.concurrent.TimeUnit
 
 import akka.actor.{ActorSystem, Props}
 import akka.util.Timeout
+import org.apache.gearpump.cluster.ClientToMaster.GetJarFileContainer
 import org.apache.gearpump.cluster._
+import org.apache.gearpump.jarstore.JarFileContainer
 import org.apache.gearpump.transport.HostPort
 import org.apache.gearpump.util.Constants._
 import org.apache.gearpump.util.{LogUtil, Configs, Util}
 import org.slf4j.{LoggerFactory, Logger}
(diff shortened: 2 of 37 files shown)


clockfly:
The configuration format is changed, we include our customization of akka config in a new section called "base", the reason is:

1. There is a new error "command line too long" when adding more dependant libraries. It is because the class path is too long, so I use lib/* for classpath.
2. After step 1, the classloader "reference.con" resolution order is un-determined. It is possible that con o akka module will override our customized configuration reference.conf under core/ and streaming/. 
3. After step2, we need a way to make sure our configuration in core/ and streaming/ will take precedence. So I add a new section called base, then I can load the configuration as follow:
```
config.getConfig("base").withFallback(config.withOutpath("base"))
```

clockfly:
/**
 *
 * File Override rule:
 *
 * System Property > Custom configuration(by using -Dconfig.file)
 * > gear.conf > reference.conf in (resource folder) > akka reference.conf
 *
 * Section Override rule:
 *
 * For master daemon: GEARPUMP_CONFIGS > MASTER > BASE > REMAIN
 * (REMAIN: is defined as config excluding section GEARPUMP_CONFIGS, MASTER, BASE)
 * For worker daemon: GEARPUMP_CONFIGS > WORKER > BASE > REMAIN
 * For App(neither Master or Worker): GEARPUMP_CONFIGS > BASE > REMAIN
 *
 * We will first use "File override rule" to get a full config, then use
 * "Section Override rule" to determine configuration for master, worker, executor,
 * and etc..
 *
 */

core/src/main/scala/org/apache/gearpump/cluster/Master.scala
```diff
@@ -78,6 +84,11 @@ private[cluster] class Master extends Actor with Stash {
       scheduler forward resourceUpdate
   }
 
+  def jarStoreService : Receive = {
```
clockfly: The JarContainer is created by Master and passed back to client, so that we can client can code transparently without knowing where the jar is stored. This is very flexible.

For example we may first use HDFS to store the jars, then HDFS cluster need a service upgrade, we cannot use it anymore, we can easily change this to use ftp to store it without upgrading client side binary.

core/src/main/scala/org/apache/gearpump/util/FileServer.scala
```diff
@@ -0,0 +1,138 @@
...
+
+import scala.util.{Failure, Success, Try}
+
+/**
+ *
+ * Should not use this to server too big files(more than 100MB), otherwise OOM may happen.
+ *
+ * port: set port to 0 if you want to bind to random port
+ */
+class FileServer(rootDir: File, host: String, port : Int) extends Actor with Stash {
```
clockfly: Currently it is still difficult for this FileServer to store/get file larger than 1GB. 

We still need to study spary to find way to support it.

project/Build.scala
```diff
@@ -129,7 +135,7 @@ object Build extends sbt.Build {
                         ),
         packExclude := Seq(fsio.id, examples_kafka.id, sol.id, wordcount.id),
         packResourceDir := Map(baseDirectory.value / "conf" -> "conf"),
-        packExpandedClasspath := true,
+        packExpandedClasspath := false,
```
clockfly: The classpath should NOT be expanded.
Otherwise, the classpath maybe too long. On windows, it will report "command line too long."

core/src/main/scala/org/apache/gearpump/cluster/client/ClientContext.scala
```diff
@@ -69,15 +80,14 @@ class ClientContext(masters: Iterable[HostPort]) {
...
-    val buf = ListBuffer[Byte]()
-    var b = fis.read()
-    while (b != -1) {
-      buf.append(b.byteValue)
-      b = fis.read()
+
+    val uploadFile = (master ? GetJarFileContainer).asInstanceOf[Future[JarFileContainer]]
+      .map {container =>
+      container.copyFromLocal(new java.io.File(jarPath))
+      AppJar(jarFile.getName, container)
```
whjiang: It is better to rename the jar file in jar store to avoid multiple clients submitting the same jar file or different version of same-name jar files.

core/src/main/scala/org/apache/gearpump/jarstore/JarFileContainer.scala
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * The JarStore is a distributed store
+ * User can use this jar store to submit application jars
+ */
+trait JarFileContainer extends Serializable {
+
+  def copyFromLocal(localFile : File) : Unit
+
+  def copyToLocalFile(localPath : File) : Unit
```
whjiang: when to delete the used jar files?

If it is part of application history, it is better to link this JarFileContainer with application info (e.g. appId or AppManager).

core/src/main/scala/org/apache/gearpump/cluster/client/ClientContext.scala
clockfly: It will be renamed in JarStore. On File Sever, it will pick a random name.

On executor, it will use name: $jarName +  random String

core/src/main/scala/org/apache/gearpump/jarstore/JarFileContainer.scala
clockfly: Good point. The expected behavior is that JarFileContainer wil be cleaned when the application is cleaned.

This will be tracked by https://github.com/intel-hadoop/gearpump/issues/257

clockfly: In Master, JarFileContainer will be binded with a appId(and persisted with HA) when application submission succeed.

core/src/main/scala/org/apache/gearpump/cluster/client/ClientContext.scala
```diff
@@ -69,15 +80,14 @@ class ClientContext(masters: Iterable[HostPort]) {
...
     }
-    AppJar(jarFile.getName, buf.toArray)
+
+    Await.result(uploadFile, Duration(15, TimeUnit.SECONDS))
   }
```
huafengw: Is Await avoidless here?

clockfly: ClientContext  is used in client side, which is not performance sensitive. It is OK to block for simplicity.

whjiang:
+1.

PR #254, opened 2015-01-01
fix #184, fix sbt package settings so that we can have a clean build.

### LINKED ISSUE #184: examples' classpath need not include core and streaming jars
As they are already on gearpump's classpath, there is no need to include core and streaming jars (also all their dependencies) on example modules' classpath. We should have something like maven's **provided** scope. I've tried `dependsOn(streaming % "provided")` but that doesn't work.

clockfly:
To  build the project

sbt assembly pack

To build to single zip package, 

sbt assembly pack-archive # or abt assembly packArchive

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/254)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,16 +34,16 @@ There is a 20 pages technical paper on typesafe blog, with technical highlights 
 
   We support [Master HA](https://github.com/intel-hadoop/gearpump/wiki/Run-Examples#master-ha) and allow master to start on multiple nodes. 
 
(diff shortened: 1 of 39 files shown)


clockfly:
Only need to review 
https://github.com/clockfly/gearpump/commit/465e8a474f46b9cdf66bc8e2439d7b2d537bcdc7

manuzhang:
+1

PR #255, opened 2015-01-03
fix #244, add more UT

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/255)


clockfly:
Only need to review https://github.com/clockfly/gearpump/commit/473416932b0865225663fa87867c5ab053216ddf
in this PR, as it depends on some old PR.

core/src/main/resources/reference.conf
```diff
@@ -1,128 +1,141 @@
...
-    scheduling {
-      scheduler-class = "org.apache.gearpump.cluster.scheduler.PriorityScheduler"
-    }
+  ### This path follows the hadoop path schema
+  ### For HDFS, use hdfs://host:port/path/
+  ### For FTP, use ftp://host:port/path
+  ### If you want to store on master nodes, then use local directory,
+  ### jarstore.rootpath = "jarstore/" will points to relative directory where master is started.
+  ### jarstore.rootpath = "/jarstore/" will points to absolute directory on master server
+  jarstore.rootpath = "jarstore/"
```
huafengw: When we pack-archive gearpump and unzip the tar file, there is only gear.conf under the conf folder, so ```jarstore.rootpath``` is not configurable for a user, is it acceptable?

core/src/main/scala/org/apache/gearpump/cluster/Application.scala
```diff
@@ -29,8 +29,53 @@ case class AppJar(name: String, container: JarFileContainer)
...
+/**
+ * This interface should keep stable, so that we can remain
+ * compatible with old versions
+ */
+trait ExecutorContextInterface {
+  val executorId : Int
+  val workerId : Int
+  val appId : Int
+  val appMaster : ActorRef
+  val startClock : TimeStamp
```
huafengw: I think startClock field is mainly related to the streaming type application, so is it necessary for this general interface?

core/src/main/scala/org/apache/gearpump/transport/netty/Context.scala
```diff
@@ -63,9 +63,9 @@ import org.apache.gearpump.transport.netty.Context._
   }
 
 
-  def bind(name: String, lookupActor : ActorLookupById): Int = {
+  def bind(name: String, lookupActor : ActorLookupById, deserializeFlag : Boolean = true): Int = {
     val taskDispatcher = system.settings.config.getString("gearpump.task-dispatcher")
```
huafengw: Is this flag an application-level configuration? If so you can move it to the user configuration.

core/src/test/scala/org/apache/gearpump/cluster/TestUtil.scala
```diff
@@ -53,4 +53,14 @@ object TestUtil{
 
     def shutDown() = system.shutdown()
   }
-}
+
+  def findFreePort: Try[Int] = {
```
huafengw: Duplicated function in core/src/main/scala/org/apache/gearpump/util/Util.scala

whjiang:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/intel-hadoop/gearpump/255)







---

<sup>**[core/src/test/scala/org/apache/gearpump/cluster/main/MainSpec.scala, line 45 \[r2\]](https://reviewable.io:443/reviews/intel-hadoop/gearpump/255#-JesuCjsujypxkYg-f3s)** ([raw file](https://github.com/intel-hadoop/gearpump/blob/a0e2f3ee2a7e8f0e4a271e3e6f7646a2bb883721/core/src/test/scala/org/apache/gearpump/cluster/main/MainSpec.scala#L45)):</sup>
it seems the comment is not consistent with the real code (2000ms vs 10s).

---

<sup>**[core/src/test/scala/org/apache/gearpump/cluster/main/MainSpec.scala, line 121 \[r2\]](https://reviewable.io:443/reviews/intel-hadoop/gearpump/255#-Jet17RPFXo3lOTCNv80)** ([raw file](https://github.com/intel-hadoop/gearpump/blob/a0e2f3ee2a7e8f0e4a271e3e6f7646a2bb883721/core/src/test/scala/org/apache/gearpump/cluster/main/MainSpec.scala#L121)):</sup>
can we group the creation of mock master into a function to avoid duplication?

---

<sup>**[examples/fsio/src/main/scala/org/apache/gearpump/streaming/examples/fsio/SeqFileStreamProcessor.scala, line 50 \[r2\]](https://reviewable.io:443/reviews/intel-hadoop/gearpump/255#-Jet2NasO3PfkEGz4C-0)** ([raw file](https://github.com/intel-hadoop/gearpump/blob/a0e2f3ee2a7e8f0e4a271e3e6f7646a2bb883721/examples/fsio/src/main/scala/org/apache/gearpump/streaming/examples/fsio/SeqFileStreamProcessor.scala#L50)):</sup>
better rename taskContext argument. It is quite confusing with the constructor param. 

maybe a better name is startTime. Why we name the type as NewStartTime? Does it mean that onStart is not called only once per task actor?

---

<sup>**[streaming/src/main/scala/org/apache/gearpump/streaming/task/TaskContext.scala, line 24 \[r2\]](https://reviewable.io:443/reviews/intel-hadoop/gearpump/255#-Jet254jp_GJFx4nNh50)** ([raw file](https://github.com/intel-hadoop/gearpump/blob/a0e2f3ee2a7e8f0e4a271e3e6f7646a2bb883721/streaming/src/main/scala/org/apache/gearpump/streaming/task/TaskContext.scala#L24)):</sup>
It's great we have TaskContext for each task now!!!!

---

core/src/main/resources/reference.conf
clockfly: There is also a section "jarstore.rootpath" in gear.conf
which will override reference.conf.

clockfly:
@whjiang, good points, I have made the changes.

core/src/main/scala/org/apache/gearpump/transport/netty/Context.scala
clockfly: good point!

I now add a TODO comment in the source code.
I am not sure whether we want to allow a application to override this config for now.

core/src/main/scala/org/apache/gearpump/cluster/Application.scala
clockfly: good point.

clockfly:
I have fixed all the comments. Please check 
https://github.com/clockfly/gearpump/commit/89f2d2299aa5a145d4e865777db30d5685028c48
 for changes.

clockfly:
Please suggest if there is other concern.

huafengw:
+1

PR #260, opened 2015-01-05
fix #259, fix stream replay api and impl, and add UTs

### LINKED ISSUE #259: fix stream replay api and impl
fix issues in #251

manuzhang:
this pr intends to address the Sean and Huafeng's comments from #251 

1. add doc/examples to transaction public api 
2. add UTs 
3. fix bugs 

The external directory structure remains  "external/kafka". The concern is that future KafkaSink may want to share some libs with KafkaSink which is not easy to do with two sub-folders "external/datasource/kafka" and "external/datasink/kafka"

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/260)


clockfly:
+1

PR #261, opened 2015-01-06
fix #217 Add Distributed Shell

### LINKED ISSUE #217: Write a small custom AppMaster example, which can run distributed shell.
Requirements:

1. An appMaster, which accept commands and response with comand line outputs dynamically.

huafengw:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/261)

### DIFF
--- a/core/src/main/scala/org/apache/gearpump/cluster/ClusterMessage.scala
+++ b/core/src/main/scala/org/apache/gearpump/cluster/ClusterMessage.scala
@@ -65,6 +65,7 @@ object AppMasterToMaster {
   case class RequestResource(appId: Int, request: ResourceRequest)
   case class SaveAppData(appId: Int, key: String, value: Any)
   case class GetAppData(appId: Int, key: String)
+  case object GetAllWorkers
 }
 
 object MasterToAppMaster {
@@ -88,6 +89,7 @@ object MasterToAppMaster {
   case class AppMasterDataDetail(appId: Int, appDescription: Application)
   case class ReplayFromTimestampWindowTrailingEdge(appId: Int)
   case class GetAppDataResult(key: String, value: Any)
+  case class WorkerList(workers: List[Int])
   case object AppDataReceived
 }
 
--- a/experiments/distributedshell/src/main/scala/org/apache/gearpump/experiments/cluster/appmaster/AbstractAppMaster.scala
+++ b/experiments/distributedshell/src/main/scala/org/apache/gearpump/experiments/cluster/appmaster/AbstractAppMaster.scala
@@ -1,0 +1,142 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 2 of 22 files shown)


experiments/distributedshell/src/main/scala/org/apache/gearpump/experiments/cluster/appmaster/AbstractAppMaster.scala
```diff
@@ -0,0 +1,142 @@
...
+        val result2 = cancelSuicide.cancel()
+        result1 && result2
+      }
+
+      def isCancelled: Boolean = {
+        cancelSend.isCancelled && cancelSuicide.isCancelled
+      }
+    }
+  }
+}
```
kkasravi: Extending AppMaster currently seems to be fairly involved and not easily done. We should have a simpler way to do this including extending the message protocol.

core/src/main/scala/org/apache/gearpump/cluster/ClusterMessage.scala
```diff
@@ -88,6 +89,7 @@ object MasterToAppMaster {
   case class AppMasterDataDetail(appId: Int, appDescription: Application)
   case class ReplayFromTimestampWindowTrailingEdge(appId: Int)
   case class GetAppDataResult(key: String, value: Any)
+  case class WorkerList(workers: List[Int])
   case object AppDataReceived
 }
```
kkasravi: Right now we extend the core protocol (AppMaster -> Master) but it's only used within experimental. We should have an easier way to extend the protocol for custom appmasters.

clockfly: @kkasravi 

Yeah, you are right. We should do that. 

Before that, we want to add more experiment projects, so that we can abstract out a more general AppMaster.

experiments/distributedshell/src/main/scala/org/apache/gearpump/experiments/cluster/appmaster/AbstractAppMaster.scala
clockfly: what you mean by  extending the message protocol.?

clockfly:
+1

clockfly:
Huafeng,

you need to add some UT for 
...a/org/apache/gearpump/experiments/cluster/appmaster/AbstractAppMaster.scala	56	0	56	0	0	0%
...cala/org/apache/gearpump/experiments/cluster/executor/DefaultExecutor.scala	18	0	18	0	0	0%
...la/org/apache/gearpump/experiments/cluster/appmaster/ExecutorLauncher.scala	16	0	16	0	0	0%

huafengw:
Got a problem here, I added the project "distributedshell" but when I run ```sbt test```, no UT under this project were executed. Is something wrong with my Build.scala?

experiments/distributedshell/src/main/scala/org/apache/gearpump/experiments/cluster/appmaster/AbstractAppMaster.scala
kkasravi: It should be easier to define the API into an AppMasters functionality. Right now adding a new message from Client to AppMaster is very involved since it must traverse master, appmanager.

clockfly:
+1

PR #262, opened 2015-01-07
fix #244 more UTs

huafengw:


core/src/test/scala/org/apache/gearpump/cluster/AppManagerSpec.scala
```diff
@@ -7,7 +7,7 @@
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *      http://www.apache.org/licenses/LICENSE-2.0
```
manuzhang: unwanted changes here.

PR #271, opened 2015-01-09
fix #244, add more UT, refactor appmanager

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/271)


core/src/main/scala/org/apache/gearpump/cluster/master/AppManager.scala
```diff
@@ -0,0 +1,256 @@
...
+  (masterHA ? GetMasterState).asInstanceOf[Future[GetMasterStateResult]].map(self ! _)
+  context.become(waitForMasterState)
+
+  def waitForMasterState: Receive = {
+    case MasterState(state) =>
+      val set = state.map(_.appId)
+      if (set.isEmpty) {
+        this.appId = 0
+      } else {
+        this.appId = state.map(_.appId).max + 1
```
whjiang: Maybe it is better to store max appId separately. Otherwise, we may generate duplicate application ID if the max appId used but the app ended and appmanager restarted.

core/src/main/scala/org/apache/gearpump/cluster/master/AppManager.scala
```diff
@@ -0,0 +1,256 @@
...
+
+  def clientMsgHandler: Receive = {
+    case SubmitApplication(app, jar, username) =>
+      LOG.info(s"AppManager Submiting Application $appId...")
+      val client = sender
+      val appLauncher = context.actorOf(launcher.props(appId, executorId, app, jar, username, context.parent, Some(client)), s"launcher${appId}_${Util.randInt}")
+
+      val appState = new ApplicationState(appId, 0, app, jar, username, null)
+      masterHA ! UpdateMasterState(appState)
+      appId += 1
```
whjiang: How to reply Client that an application is submitted successfully or not?

core/src/main/scala/org/apache/gearpump/cluster/master/AppManager.scala
clockfly: It is a future feature to serialize the master state to disk.

Currently, we assume the master cluster is always running.

core/src/main/scala/org/apache/gearpump/cluster/master/AppManager.scala
clockfly: Thanks for finding this.

the code to reply to client is part of AppMasterLauncher.

I have just add the UT in AppMasterLauncherSpec.

core/src/main/scala/org/apache/gearpump/cluster/master/AppManager.scala
clockfly: Good point, fixed in new PR.

core/src/main/scala/org/apache/gearpump/cluster/master/AppMasterLauncher.scala
```diff
@@ -0,0 +1,140 @@
...
+    case ResourceAllocated(allocations) =>
+      LOG.info(s"Resource allocated for appMaster $appId")
+      val ResourceAllocation(resource, worker, workerId) = allocations(0)
+      val appMasterContext = AppMasterContext(appId, username, executorId, resource, jar, null, AppMasterRuntimeInfo(worker))
+
+      LOG.info(s"Try to launch a executor for app Master on ${worker} for app $appId")
+      val name = ActorUtil.actorNameForExecutor(appId, executorId)
+      val selfPath = ActorUtil.getFullPath(context.system, self.path)
+
+      val jvmSetting = Util.resolveJvmSetting(app.conf, systemConfig).appMater
```
whjiang: appMater? It is typo for appMaster?

core/src/main/scala/org/apache/gearpump/cluster/master/AppMasterLauncher.scala
```diff
@@ -0,0 +1,140 @@
...
+      worker ! LaunchExecutor(appId, executorId, resource, executorJVM)
+      context.become(waitForActorSystemToStart(worker, appMasterContext, app.conf))
+  }
+
+  def waitForActorSystemToStart(worker : ActorRef, appContext : AppMasterContext, user : UserConfig) : Receive = {
+    case ExecutorLaunchRejected(reason, resource, ex) =>
+      LOG.error(s"Executor Launch failed reason：$reason", ex)
+      LOG.info(s"reallocate resource $resource to start appmaster")
+      master ! RequestResource(appId, ResourceRequest(resource))
+      context.become(waitForResourceAllocation)
```
whjiang: will this become an infinite loop? executor launch rejected -> resource allocatioin -> executor launch rejected ....

may be add a counter for retry better.

whjiang:
+1

PR #302, opened 2015-01-14
fix #244 add AppMasterSpec and complete other UT

huafengw:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/302)


core/src/test/scala/org/apache/gearpump/cluster/MasterHarness.scala
```diff
@@ -40,7 +40,7 @@ trait MasterHarness {
   private var systemAddress: Address = null
   private var host: String = null
   private var port: Int = 0
-  val PROCESS_BOOT_TIME = Duration(10, TimeUnit.SECONDS)
+  val PROCESS_BOOT_TIME = Duration(20, TimeUnit.SECONDS)
```
manuzhang: maybe we can make this a conf rather than modifying it every time in code

clockfly:
+1

PR #305, opened 2015-01-14
fix #301, fix UserConfig ClassNotFoundException by storing everything in...

clockfly:
... Config as String

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/305)

PR #307, opened 2015-01-14
fix #306, ClassNotFound for customized Application type

### LINKED ISSUE #306: ClassNotFound when submitting a customized type of appmaster
Currently,

User need to extend trait Application to define its own application type, and then this application type is submitted to Master to create application.

When the user define some new type of application, like Graph, then the Master will not be able to respove the classpath of the new application type, then ClassNotFound will be thrown.

clockfly:
It depends on #305 

The change after #305 is here https://github.com/clockfly/gearpump/commit/b55556471fa00345531edc95df218a3d5d1b0a5b

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/307)


kkasravi:
Looks like AppManagerSpec failed on build.

core/src/main/scala/org/apache/gearpump/cluster/Application.scala
```diff
@@ -23,70 +23,63 @@ import org.apache.gearpump._
...
+ * @param appMaster: The class name of AppMaster Actor
+ *    The AppMaster must have a constructor this(appContext : AppMasterContext, app : Application)
  */
-trait Application {
-  val name : String
-  val appMaster : String
-  val conf: UserConfig
-}
+
+final case class Application(name : String, appMaster : String, conf: UserConfig)
```
kkasravi: Extending from case class is not recommended and may be deprecated. See http://www.scala-lang.org/old/node/1582

clockfly: We are enforcing using final.

So Application will not be extendable.

clockfly: Application will not have any child class.

So, I don't think sealed trait is appliable here.

clockfly: The reason why we can not extend  Application is that,
Application instance will be sent to Master daemon. If we define a sub-type of Application, then Master will not be able to find the sub-type in classpath.

clockfly: Maybe I should use normal class, instead of case class.

core/src/main/scala/org/apache/gearpump/cluster/UserConfig.scala
```diff
@@ -75,6 +83,12 @@ import org.apache.gearpump.cluster.UserConfig._
    * requires an implicit ActorSystem.
    *
    * @see [[http://doc.akka.io/docs/akka/snapshot/scala/serialization.html#A_Word_About_Java_Serialization]]
+=======
+   *
+   * Need to set a serializer,
+   *
+   * implicit serializer: akka.serialization.Serializer = new JavaSerializer(actorSystem)
+>>>>>>> fix #301, fix UserConfig ClassNotFoundException by storing everything in
    */
```
huafengw: need to clean up some conflict messages

huafengw:
+1

PR #331, opened 2015-01-21
fix #330, Allow user to output the metrics to log file instead of graphi...

### LINKED ISSUE #330: Allow user to output the metrics to log file instead of graphite

clockfly:
...te

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/331)

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -17,13 +17,29 @@ gearpump {
   ### If you want to use metrics, please change
   ###########################
 
-  ### Flag to enable metrics
-  metrics.enabled = false
 
-  ## Graphite host settings
-  metrics.graphite.host = "192.168.1.54"
-  metrics.graphite.port = 2003
-  metrics.sample.rate = 1
+  ### Flag to enable metrics
+  metrics {
+    enabled = false
+
+    # We will take one metric out of ${sample.rate}
+    sample-rate = 10
+
+    report-interval-ms = 15000
+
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/gearpump/metrics/Metrics.scala
```diff
@@ -57,11 +57,14 @@ object Metrics extends ExtensionId[Metrics] with ExtensionIdProvider {
 
     val metricsEnabled = system.settings.config.getBoolean("gearpump.metrics.enabled")
     LOG.info(s"Metrics is enabled...,  $metricsEnabled")
-    val sampleRate = system.settings.config.getInt("gearpump.metrics.sample.rate")
+    val sampleRate = system.settings.config.getInt("gearpump.metrics.sample-rate")
```
huafengw: Move these config key into Constants

core/src/main/resources/reference.conf
```diff
@@ -17,13 +17,29 @@ gearpump {
...
-  ## Graphite host settings
-  metrics.graphite.host = "192.168.1.54"
-  metrics.graphite.port = 2003
-  metrics.sample.rate = 1
+    # reporter = "logfile"
+    reporter = "graphite"
+
+    graphite {
+      ## Graphite host settings
+      host = "192.168.1.54"
```
manuzhang: use 127.0.0.1 for default ?

huafengw:
+1

PR #342, opened 2015-01-22
fix #299 fix bug in StreamingTestUtil.createEchoForTaskActor

### LINKED ISSUE #299: Some UT may fail randomly, most because of the expectMsg time out

huafengw:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/342)


clockfly:
Why you think this will fix the problem?

huafengw:
The TaskActor will register itself to AppMaster after registering to the Express. When the mock appmaster received both two RegisterTask messages, each Express's local actor is ready.

clockfly:
It should work,  +1

another way is to do #345.

PR #348, opened 2015-01-22
fix #338, fix kafka leader not available exception

### LINKED ISSUE #338: TopicOrPartitionUnknown exception thrown by KafkaStorage
The cause is that topic for kafka storage is created at runtime and kafka leader has not been elected or metadata not propagated to all brokers.

manuzhang:
1. check for topic exists before creating topic
2. client tries to fetch topic metadata from kafka with configurable timeout after creating topic

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/348)

### DIFF
--- a/conf/kafka.conf
+++ b/conf/kafka.conf
@@ -18,6 +18,7 @@ kafka {
     producer.type = "sync"
     request.required.acks = "1"
     emit.batch.size = 100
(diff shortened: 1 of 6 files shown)


huafengw:
+1

PR #357, opened 2015-01-23
fix #356 filter the empty Config

### LINKED ISSUE #356: User's application.conf can not be loaded
When launch a Executor, ExecutorWatcher will set ```Constants.GEARPUMP_CUSTOM_CONFIG_FILE``` to the arguments, but when the config file is empty, this argument should not be set.

huafengw:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/357)


clockfly:
+1

PR #376, opened 2015-01-28
fix #372, fix DAG subgraph build

### LINKED ISSUE #372: TaskContext.dag.tasks is empty in a single node DAG
the cause is in creating a subGraph for a task, where the tasks map is built from the `newGraph` which is empty when there are no outgoing edges. 

```scala
  def subGraph(task : Int) = {
    var newGraph = Graph.empty[Int, Partitioner]
    graph.edgesOf(task).foreach { edge =>
      val (node1, partitioner, node2) = edge
      newGraph.addEdge(node1, partitioner, node2)
    }
    val newMap = newGraph.vertex.foldLeft(Map.empty[Int, TaskDescription]){ (map, vertex) =>
      val task = tasks.get(vertex).get

      //clean out other in-degree and out-degree tasks' data except the task parallelism
      map + (vertex -> task.copy(null, task.parallelism))
    }
    new DAG(newMap, newGraph)
  }
```

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/376)

### DIFF
--- a/core/src/main/scala/org/apache/gearpump/util/Graph.scala
+++ b/core/src/main/scala/org/apache/gearpump/util/Graph.scala
@@ -34,33 +34,33 @@ import scala.language.implicitConversions
 class Graph[N, E](private[Graph] val graph : DefaultDirectedGraph[N, Edge[E]]) extends Serializable{
   import org.apache.gearpump.util.Graph._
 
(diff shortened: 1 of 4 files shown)


clockfly:
Can you add UT for this?

clockfly:
+1

PR #379, opened 2015-01-28
fix #378, construct the serializer explicitly, instead of implicitly.

### LINKED ISSUE #378: Error: Cannot find serializer for Message
For message that is passed between tasks, we require the user to define a explicit kryo serializer, for the benefit of performance.

But sometimes, there may have ocassional error that the platform cannot find the serializer for Message.

clockfly:
Before this fix, we don't know what config are is really used to initialize the KryoSerializer. As com.romix.akka.serialization.kryo.KryoSerializationExtension call GearpumpSerialization.custom() without passing any configuration as arguments.

After this fix, we will call GearpumpSerailization.custom(config)

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/379)

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -51,13 +51,6 @@ gearpump {
   log.application.dir = "logs"
 
   serializers {
-    ## Follow this format when adding new serializer for new message types
-    ##    "org.apache.gearpump.Message" = "org.apache.gearpump.streaming.MessageSerializer"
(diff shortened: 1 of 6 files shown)


clockfly:
Can someone review this?

huafengw:
+1

kkasravi:
I'll get to it either later tonight or tomorrow morning.
Thanks

PR #381, opened 2015-01-28
fix #380, set the context class loader as the URL class loader for Actor...

### LINKED ISSUE #380: ClassNotFound at when client side ActorSystem received a custom message
ActorSystem will use context class loader in precedence.

As we are using URLClassloader to load the application jar, we also need to set the context classloader to the URL class loader, so that ActorSystem will be able to use the URL class loader to deserialize unknown mesage type.

clockfly:
...System.

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/381)


whjiang:
+1.

PR #384, opened 2015-01-30
Fixes #358

### LINKED ISSUE #358: Add back functionality of AppMasterDataDetailRequest, AppMasterDataDetail
PR for fix #244 removed this functionality which deleted the REST API that allowed a DAG to be returned  for the REST API call /appmaster/<id>?detail=true

kkasravi:
AppMaster now responds directly to client (using forward which preserves original sender).
uPickle simplifies JSON output, using a customized implicit writer defined in DAG.
RestServices is much simplier but can probably be parametertized on what JSON types are emitted.
Work will be continued in followon PR's to add the DAG widget, dashboard, etc.

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/384)


clockfly:
I made several comments, but then the comments is gone.

clockfly:
I made several comments, but then the comments is gone.

rest/src/main/scala/org/apache/gearpump/services/AppMasterService.scala
```diff
@@ -18,54 +18,52 @@
...
+                 Option(application) match {
+                   case Some(app) =>
+                     complete(write(dag(Option(application.userConfig))))
+                   case None =>
+                     complete(StatusCodes.InternalServerError, "UserConfig is null")
+                 }
+               case Failure(ex) => complete(StatusCodes.InternalServerError, s"An error occurred: ${ex.getMessage}")
+             }
+           case false =>
+             onComplete((master ? AppMasterDataRequest(appId)).asInstanceOf[Future[AppMasterData]]) {
```
clockfly: Since you are using uPickle for AppMasterDataDetail, maybe we should use uPickle consistently for all response of REST, like AppMasterData.

rest/src/main/scala/org/apache/gearpump/services/RestServices.scala
```diff
@@ -20,59 +20,48 @@ package org.apache.gearpump.services
...
-  val appMasterService = new AppMasterService {
-    val master = masters
-    def actorRefFactory = context
-  }
+trait RestServices extends AppMastersService with AppMasterService {
+  val route = appMastersRoute ~ appMasterRoute ~
+    get {
+      pathPrefix("") {
+        pathEndOrSingleSlash {
+          getFromResource("index.html")
```
clockfly: getFromResource() means  we are loading index.html from resource of claspath?

Do we still can modify the index.html dynamically without restarting rest?

rest/src/main/scala/org/apache/gearpump/services/RestServices.scala
```diff
@@ -20,59 +20,48 @@ package org.apache.gearpump.services
...
 
-  override def preStart(): Unit = {
-    context.actorOf(Props(classOf[AppMastersServiceActor], masters), "appMastersService")
-    context.actorOf(Props(classOf[AppMasterServiceActor], masters), "appMasterService")
+class RestServicesActor(masters: ActorRef) extends Actor with RestServices {
+  val master = masters
+  def actorRefFactory = context
+  implicit val system = context.system
+  implicit val executionContext: ExecutionContext = context.dispatcher
+  def dag(config: Option[UserConfig]): DAG = {
```
clockfly: This is not a safe call.

We should first the Appication type explicily, to find out whether this is a streaming application.
Then, maybe we need to add a implicit conversion in compaion of AppDescription to convert Application to streaming AppDescription.

streaming/src/main/scala/org/apache/gearpump/streaming/DAG.scala
```diff
@@ -51,6 +52,16 @@ object DAG {
       taskDescription.equals(node)
     }.get._1
   }
+
+  implicit val writer = upickle.Writer[DAG] {
```
clockfly: If some other object contains the DAG, like AppDescription(name, DAG), will uPickle automatically detect the writer you defined here when serializing AppDescription using uPickle?

rest/src/main/scala/org/apache/gearpump/services/AppMasterService.scala
kkasravi: Yes, I think I can get rid of AppMasterProtocol in general. I'll update.

rest/src/main/scala/org/apache/gearpump/services/RestServices.scala
kkasravi: It depends where we point the getFromResource to - it can be from a filesystem. Right now I believe index.html isn't functional since it will actually read from a 'dashboard' index.html in upcoming PR.

rest/src/main/scala/org/apache/gearpump/services/RestServices.scala
kkasravi: We could do something like 
        val dag = DAG(_config.getValue[Graph[TaskDescription, Partitioner]](AppDescription.DAG).getOrElse(Graph.empty[TaskDescription, Partitioner]))

since in general UserConfig.getValue[T] should be well behaved.

streaming/src/main/scala/org/apache/gearpump/streaming/DAG.scala
kkasravi: Yes, uPickle will look for implicit writer[T] within the companion object.

kkasravi: I should add that no explicit writer is needed for AppDescription since the other field (name) is handled implicitly. In the case of DAG we are sort of rolling our own JSON. We may want to have an implicit conversion (other than writer[T]) which is more JSON friendly. Will look at this more closely when the DAG widget evolves.

rest/src/main/scala/org/apache/gearpump/services/AppMasterService.scala
kkasravi: Additional note. uPickle is scala.js friendly should we go that route.

rest/src/main/scala/org/apache/gearpump/services/RestServices.scala
clockfly: I mean you cannot assure there is a DAG in the UserConfig. So, the code should reflect that you have already checked that it is a streaming application.

rest/src/test/scala/org/apache/gearpump/services/AppMasterServiceSpec.scala
```diff
@@ -44,6 +46,16 @@ class AppMasterServiceSpec extends FlatSpec with ScalatestRouteTest with AppMast
       null
   }
 
+  def dag(config: Option[UserConfig]): DAG = {
+    config match {
```
clockfly: It is not OK to define this operation here. 
As we set the DAG field of UserConfig in AppDescription, you should also define this under AppDescription. 

A better way is to define a implicit conversion from Application to AppDescription in companion of AppDescription.

like this:

object AppDescription　｛
　def ApplicationToAppDescripton: AppDescription(app: Application) {
     val dag = _config.getValue[Graph[TaskDescription, Partitioner]](AppDescription.DAG).get)
    AppDescription(name, mainClass, dag)
  }
｝

kkasravi: We would need an implicit ActorSystem
```  def getValue[T](key: String)(implicit system: ActorSystem): Option[T]  = {
```
So your approach wouldn't work. I agree we shouldn't need to repeat the dag operation within the Spec class.

rest/src/main/scala/org/apache/gearpump/services/RestServices.scala
kkasravi: Yes. We could avoid a call to userConfig if we determine that Application is not a 'streaming' Application. But this is not very elegant or extensible if we have to make similar calls in order to determine if userConfig contains other useful values. I think we need something different. For now just having userConfig return null allows easy refactoring in the future.

rest/src/main/scala/org/apache/gearpump/services/AppMasterService.scala
```diff
@@ -18,54 +18,58 @@
...
+  import upickle._
   val master:ActorRef
+  implicit val system: ActorSystem
 
-  val routes = readRoute 
+  def dag(config: Option[UserConfig]): DAG = {
+    config match {
+      case Some(_config) =>
+        DAG(_config.getValue[Graph[TaskDescription, Partitioner]](AppDescription.DAG).getOrElse(Graph.empty[TaskDescription, Partitioner]))
+      case None =>
```
clockfly: Can we move this to AppDescription, by adding an implicit conversion?

I asked the same question before.

You said it won't work because it requires an implicit actor system. Why it is a problem? Cannot we provide an implicit actor system?

clockfly: Generally, this PR is very good!
I am +1 after this small fix.

kkasravi:
Sample output for wordcount
```
{"name":"wordCount","dag":{"vertices":["org.apache.gearpump.streaming.examples.wordcount.Split","org.apache.gearpump.streaming.examples.wordcount.Sum"],"edges":[["org.apache.gearpump.streaming.examples.wordcount.Split","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.wordcount.Sum"]]}}
```

Sample output for complexdag:
```
{"name":"dag","dag":{"vertices":["org.apache.gearpump.streaming.examples.complexdag.Node_2","org.apache.gearpump.streaming.examples.complexdag.Sink_3","org.apache.gearpump.streaming.examples.complexdag.Node_3","org.apache.gearpump.streaming.examples.complexdag.Node_4","org.apache.gearpump.streaming.examples.complexdag.Sink_0","org.apache.gearpump.streaming.examples.complexdag.Node_0","org.apache.gearpump.streaming.examples.complexdag.Sink_4","org.apache.gearpump.streaming.examples.complexdag.Sink_1","org.apache.gearpump.streaming.examples.complexdag.Source_0","org.apache.gearpump.streaming.examples.complexdag.Node_1","org.apache.gearpump.streaming.examples.complexdag.Sink_2","org.apache.gearpump.streaming.examples.complexdag.Source_1"],"edges":[["org.apache.gearpump.streaming.examples.complexdag.Source_0","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.complexdag.Sink_2"],["org.apache.gearpump.streaming.examples.complexdag.Source_0","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.complexdag.Sink_0"],["org.apache.gearpump.streaming.examples.complexdag.Node_0","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.complexdag.Sink_3"],["org.apache.gearpump.streaming.examples.complexdag.Node_1","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.complexdag.Node_4"],["org.apache.gearpump.streaming.examples.complexdag.Source_0","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.complexdag.Node_3"],["org.apache.gearpump.streaming.examples.complexdag.Node_1","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.complexdag.Sink_3"],["org.apache.gearpump.streaming.examples.complexdag.Source_1","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.complexdag.Node_0"],["org.apache.gearpump.streaming.examples.complexdag.Source_0","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.complexdag.Node_1"],["org.apache.gearpump.streaming.examples.complexdag.Node_2","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.complexdag.Node_3"],["org.apache.gearpump.streaming.examples.complexdag.Node_3","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.complexdag.Sink_3"],["org.apache.gearpump.streaming.examples.complexdag.Source_1","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.complexdag.Sink_4"],["org.apache.gearpump.streaming.examples.complexdag.Node_4","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.complexdag.Sink_3"],["org.apache.gearpump.streaming.examples.complexdag.Source_0","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.complexdag.Node_2"],["org.apache.gearpump.streaming.examples.complexdag.Source_0","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.complexdag.Sink_1"],["org.apache.gearpump.streaming.examples.complexdag.Node_1","org.apache.gearpump.partitioner.HashPartitioner","org.apache.gearpump.streaming.examples.complexdag.Node_3"]]}}
```

clockfly:
+1
This is great!

clockfly:
By the way, next time, can you add some text description in the commit comments? It will be easier when browsering through the commit logs.

clockfly:
By the way, we also need to add UT for DetailAppmasterData.

Maybe, create a new issue for this.

kkasravi:
Yes, I'll do that, thanks!

PR #385, opened 2015-02-01
fix #370, refactor streaming appmaster

### LINKED ISSUE #370: refactor the streaming appmaster
There are several problems currently:

1. MasterHA is not transparent to appmaster. AppMaster need to retry connection to connect to active master.
2. There are a lot of boilplate code which provision the jar to remote node, and start an actor system. 
3. The AppMaster is too big, and not easy to test.

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/385)


core/src/main/scala/org/apache/gearpump/cluster/scheduler/PriorityScheduler.scala
```diff
@@ -47,34 +47,34 @@ class PriorityScheduler extends Scheduler{
     var allocated = Resource.empty
     val totalResource = resourcesSnapShot.foldLeft(Resource.empty){ (totalResource, workerWithResource) =>
       val (_, (_, resource)) = workerWithResource
-      totalResource.add(resource)
+      totalResource.+(resource)
```
manuzhang: I think we can use infix notation here and other `Resource` operations as well.

clockfly: Thanks, fixed

core/src/main/scala/org/apache/gearpump/cluster/appmaster/AppMasterRuntimeEnvironment.scala
```diff
@@ -0,0 +1,160 @@
...
+import org.apache.gearpump.cluster.AppMasterToMaster.RegisterAppMaster
+import org.apache.gearpump.cluster.appmaster.AppMasterRuntimeEnvironment._
+import org.apache.gearpump.cluster.appmaster.ExecutorSystemScheduler.{Session, StartExecutorSystems}
+import org.apache.gearpump.cluster.master.MasterProxy
+import org.apache.gearpump.cluster.{AppMasterContext, Application}
+import org.apache.gearpump.transport.HostPort
+import org.apache.gearpump.util.LogUtil
+import org.apache.gearpump.cluster.appmaster.MasterConnectionKeeper.MasterConnectionStatus._
+
+class AppMasterRuntimeEnvironment(
```
manuzhang: Could you add some comments to explain the function of this class ?

core/src/main/scala/org/apache/gearpump/cluster/appmaster/AppMasterRuntimeEnvironment.scala
```diff
@@ -0,0 +1,160 @@
...
+
+  private val master = context.actorOf(masterFactory(appId, context.actorOf(Props(new MasterProxy(masters)))))
+  private var appContext = appContextInput.copy(masterProxy = master)
+
+  //create appMaster proxy to receive command and forward to appmaster
+  private val appMaster = context.actorOf(appMasterFactory(appContext, app))
+  context.watch(appMaster)
+
+  private val registerAppMaster = RegisterAppMaster(appMaster, appContext.registerData)
+  private val masterConnectionKeeper =  context.actorOf(masterConnectionKeeperFactory(master, registerAppMaster, self))
```
manuzhang: It seems that this actor is never used

core/src/main/scala/org/apache/gearpump/cluster/appmaster/AppMasterRuntimeEnvironment.scala
```diff
@@ -0,0 +1,160 @@
...
+  masters: Iterable[ActorPath],
+  masterFactory: (AppId, MasterActorRef) => Props,
+  appMasterFactory: (AppMasterContext, Application)=> Props,
+  masterConnectionKeeperFactory: (MasterActorRef, RegisterAppMaster, ListenerActorRef) => Props) extends Actor {
+
+  val appId = appContextInput.appId
+  private val LOG = LogUtil.getLogger(getClass, app = appId)
+
+  private val master = context.actorOf(masterFactory(appId, context.actorOf(Props(new MasterProxy(masters)))))
+  private var appContext = appContextInput.copy(masterProxy = master)
```
manuzhang: this could be a `val`

core/src/main/scala/org/apache/gearpump/cluster/appmaster/AppMasterRuntimeEnvironment.scala
```diff
@@ -0,0 +1,160 @@
...
+  //create appMaster proxy to receive command and forward to appmaster
+  private val appMaster = context.actorOf(appMasterFactory(appContext, app))
+  context.watch(appMaster)
+
+  private val registerAppMaster = RegisterAppMaster(appMaster, appContext.registerData)
+  private val masterConnectionKeeper =  context.actorOf(masterConnectionKeeperFactory(master, registerAppMaster, self))
+
+  def receive: Receive = {
+    case MasterConnected =>
+      LOG.error(s"master is connected, start AppMaster ${appId}...")
```
manuzhang: this is an "error" ?

core/src/main/scala/org/apache/gearpump/cluster/appmaster/ExecutorSystemLauncher.scala
```diff
@@ -0,0 +1,121 @@
...
+
+      worker.ref ! launch
+      context.become(waitForActorSystemToStart(sender, launch, worker))
+  }
+
+  def waitForActorSystemToStart(replyTo: ActorRef, launch: LaunchExecutor, worker: WorkerInfo) : Receive = {
+    case RegisterActorSystem(systemPath) =>
+      import launch._
+      timeout.cancel()
+      LOG.info(s"Received RegisterActorSystem $systemPath for app master")
```
manuzhang: "for app master"?

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/ClockService.scala
```diff
@@ -16,54 +16,91 @@
...
-    dag.tasks.foreach { taskIdWithDescription =>
-      val (taskGroupId, description) = taskIdWithDescription
-      val taskClocks = new Array[TimeStamp](description.parallelism).map(_ => Long.MaxValue)
-      val taskGroupClock = new TaskGroupClock(taskGroupId, Long.MaxValue, taskClocks)
-      taskGroupClocks.add(taskGroupClock)
-      taskGroupLookup.put(taskGroupId, taskGroupClock)
+    LOG.info("Initializing Clock service, get snapshotted StartClock ....")
+
+    store.get(START_CLOCK).asInstanceOf[Future[TimeStamp]].map { clock =>
+      val startClock = Option(clock).getOrElse(0L)
```
manuzhang: the minClock will never get updated if its initial value is 0. Also, I don't think ClockServiceSpec has well tested this

clockfly: Can you elaborate more about the scenario why clock service is not working?

clockfly: Yes, it indeed has issues.

Still checking it.

clockfly: Fixed/ There is a bug in TaskManager.

clockfly:
UT for TaskManager is not added in this PR. I prefer to do it in followup PR, as the TaskManager still need large refactory, we need to seperate the task scheduling part to a seperate module. So that we can add a sub-graph dynamically to a running graph.

clockfly:
For now, the test coverage for Taskmanager can be reached by AppMasterSpec.

manuzhang:
Frankly, I cannot comprehend or verify all the bits in this pr as this is a big change but I'm +1 for merging  this as long as all the UTs pass. You may also want to update the Actor Hierarchy graph afterwards.

clockfly:
It is OK, 

let's keep it open for several days, so that everyone get a chance to review it.

clockfly:
The Actor hierachy graph is still the same. This PR is type of implementation level improvements.

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/ClockService.scala
huafengw: For some self-driven TaskActor, like Split in Wordcount, they may send self a Message without Timestamp, so the clock will not update because the minimal number is always zero.

core/src/main/scala/org/apache/gearpump/cluster/appmaster/AppMasterRuntimeEnvironment.scala
```diff
@@ -0,0 +1,199 @@
...
+ */
+private[appmaster]
+class AppMasterRuntimeEnvironment (
+    appContextInput: AppMasterContext,
+    app: Application,
+    masters: Iterable[ActorPath],
+    masterFactory: (AppId, MasterActorRef) => Props,
+    appMasterFactory: (AppMasterContext, Application)=> Props,
+    masterConnectionKeeperFactory: (MasterActorRef, RegisterAppMaster, ListenerActorRef) => Props)
+  extends Actor {
```
huafengw: Maybe rename this class a better name because intuitively the "environment" should be static, not an actor.

clockfly: This is like what a JRE start a java application.

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/ClockService.scala
clockfly: Currently, for the example code, they are using system clock when sending self messages.

PR #390, opened 2015-02-03
fix #389, fix fetch thread bug

### LINKED ISSUE #389: kafka fetch thread exits when there are no more messages
The fetch thread should sleep rather than exit when there are no more messages such that it could continue consuming data when new messages available.

manuzhang:
PR summary,

1. FetchThread sleeps rather than exits when there are no more messages
2. update KafkaStreamProducerSpec to cover the case

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/390)


clockfly:
+1

PR #409, opened 2015-02-07
fix #408, handle Ctrl+C(sigint) gracefully.

### LINKED ISSUE #408: ActorSystem should handle Ctrl+C (SIGINT) gracefully
On Linux, I found the SIGINT will be passed from parent process to child process, and will shutdown the child process directly. When this happen, ActorSystem don't get the chance to cleanup itself, like closing the connection, closing the file handler.

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/409)

### DIFF
--- a/core/src/main/scala/org/apache/gearpump/util/ActorSystemBooter.scala
+++ b/core/src/main/scala/org/apache/gearpump/util/ActorSystemBooter.scala
@@ -56,7 +56,20 @@ object ActorSystemBooter  {
 
     LogUtil.loadConfiguration(config, ProcessType.APPLICATION)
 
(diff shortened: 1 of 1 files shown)


manuzhang:
hi, in which case will a user hit `Ctrl+C` to interrupt an `ActorSystemBooter` ?

clockfly:
For example, when user starts a worker, and then he may ctrl+C to stop the worker.

clockfly:
In this case, all child processes of worker will receive a Ctrl+C on linux. Please check #408, there are description about how Ctrl+c can be propagated to child processes.

manuzhang:
+1

PR #442, opened 2015-02-11
fix  #423, refactor task by seperating TaskActor and Task

### LINKED ISSUE #423: Refactor task implementation
Goal of this refactory:

1. Seperate Task from heavy TaskActor, so that it is easier to test Tasks.
2. When there is a temporary network down, and some message is lost but recovered immediately, we should not clean in-memory state. 

This issue should be considered the first step toward requirements listed in #288 

1. When there is message loss, and trigger mesage replay, existing state in memory should not be lost.
2. The source actors should also accept push-style messages. For example, A device may push the message directly into the source task actors. 
3. For exactly once mesage delivery, the state in memory need to be checkpointed, the state storage api should be part of the TaskActor.
4. How to track the clock timestamp of task effectively if this task has no output mesage.

clockfly:
The purpose of this PR is to sepeate the TaskActor and Task, so that we don't need to create a actor system just to test the function of a task.

I am NOT satified of this refactory considering the requirements in https://github.com/intel-hadoop/gearpump/issues/288 , but it should be much better than previsous code. 

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/442)

### DIFF
--- a/streaming/src/main/scala/org/apache/gearpump/streaming/executor/Executor.scala
+++ b/streaming/src/main/scala/org/apache/gearpump/streaming/executor/Executor.scala
@@ -16,26 +16,24 @@
  * limitations under the License.
  */
 
-package org.apache.gearpump.streaming
+package org.apache.gearpump.streaming.executor
 
 import akka.actor.SupervisorStrategy.Restart
 import akka.actor._
-import org.apache.gearpump._
 import org.apache.gearpump.cluster.MasterToAppMaster.ReplayFromTimestampWindowTrailingEdge
-import org.apache.gearpump.cluster.scheduler.Resource
 import org.apache.gearpump.cluster.{ExecutorContext, UserConfig}
 import org.apache.gearpump.streaming.AppMasterToExecutor._
-import org.apache.gearpump.streaming.Executor.RestartExecutor
 import org.apache.gearpump.streaming.ExecutorToAppMaster.RegisterExecutor
+import org.apache.gearpump.streaming.executor.Executor.{RestartExecutor, TaskLocationReady}
 import org.apache.gearpump.streaming.task.TaskActor.RestartTask
-import org.apache.gearpump.streaming.task.{TaskId, TaskLocations}
+import org.apache.gearpump.streaming.task.{TaskId, TaskLocations, TaskWrapper}
 import org.apache.gearpump.transport.{Express, HostPort}
 import org.apache.gearpump.util.{Constants, LogUtil}
 import org.slf4j.Logger
-import scala.language.postfixOps
-import scala.concurrent.duration._
 
-case object TaskLocationReady
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
 
 class Executor(executorContext: ExecutorContext, userConf : UserConfig)  extends Actor {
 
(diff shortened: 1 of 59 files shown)


streaming/src/main/scala/org/apache/gearpump/streaming/executor/Executor.scala
```diff
@@ -64,10 +64,11 @@ class Executor(executorContext: ExecutorContext, userConf : UserConfig)  extends
...
-    case LaunchTask(taskId, taskContext, taskClass) => {
+    case LaunchTask(taskId, taskContext, taskClass, taskActorClass) => {
       LOG.info(s"Launching Task $taskId for app: ${appId}, $taskClass")
+      val task = new TaskWrapper(taskClass, taskContext, userConf)
       val taskDispatcher = context.system.settings.config.getString(Constants.GEARPUMP_TASK_DISPATCHER)
-      val task = context.actorOf(Props(taskClass, taskContext, userConf).withDispatcher(taskDispatcher), "group_" + taskId.groupId + "_task_" + taskId.index)
+      val taskActor = context.actorOf(Props(taskActorClass, taskContext, userConf, task).withDispatcher(taskDispatcher), "group_" + taskId.groupId + "_task_" + taskId.index)
     }
     case TaskLocations(locations) =>
       val result = locations.flatMap { kv =>
```
huafengw: I noticed that this class's ackage name is not consistent with directory structure.

streaming/src/main/scala/org/apache/gearpump/streaming/task/TaskActor.scala
```diff
@@ -31,7 +31,7 @@ import org.apache.gearpump.util.{LogUtil, TimeOutScheduler, Util}
 import org.apache.gearpump.{Message, TimeStamp}
 import org.slf4j.Logger
 
-abstract class TaskActor(val taskContext : TaskContext, userConf : UserConfig) extends Actor with ExpressTransport  with TimeOutScheduler{
+class TaskActor(val taskContext : TaskContextData, userConf : UserConfig, val task: TaskWrapper) extends Actor with ExpressTransport  with TimeOutScheduler{
```
huafengw: maybe rename ```val taskContext``` to taskContextData is better.

streaming/src/main/scala/org/apache/gearpump/streaming/task/TaskWrapper.scala
```diff
@@ -0,0 +1,95 @@
...
+  override def actorOf(props: Props, name: String): ActorRef = actor.context.actorOf(props, name)
+
+  override def onStart(startTime: StartTime): Unit = {
+    if (null != task) {
+      LOG.error("Task.onStart should not be called multiple times...")
+    }
+    val constructor = taskClass.getConstructor(classOf[TaskContext], classOf[UserConfig])
+    task = constructor.newInstance(this, userConf)
+    task.onStart(startTime)
+  }
```
huafengw: When the TaskActor encounter some exception and restarted, the onStart() will be recalled. But at that time the TaskWrapper object's task is not null, right?

clockfly: Hi huafeng,

You are right, eventually we will only re-create a task when the executor fails(consider JVM crash).

The purpose of this PR is to seperate TaskActor with Task. I will leave this to issue #452

huafengw:
+1

PR #446, opened 2015-02-12
fix #419 Add a timestamp for app logs

### LINKED ISSUE #419: Add a timestamp for app logs
currently, the logs looks like this:
```
logs/
  applicationData/
     username/
        app1/
        app2
```

When user restart the master, then it will restart the appId from 1, and we cannot differentiate these two run then.

There also should be a timestamp in the folder name. Like this:
  yymmddhhmmss_app1
  yymmddhhmmss_app2

It is important to put the timestamp before the appId, so that the log folder can get sorted natually

huafengw:
Now the app log folder name is like 2015-2-12-10-12_app1

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/446)

### DIFF
--- a/core/src/main/scala/org/apache/gearpump/cluster/master/Master.scala
+++ b/core/src/main/scala/org/apache/gearpump/cluster/master/Master.scala
@@ -26,7 +26,8 @@ import org.apache.gearpump.cluster.ClientToMaster._
 import org.apache.gearpump.cluster.MasterToAppMaster._
 import org.apache.gearpump.cluster.MasterToWorker._
 import org.apache.gearpump.cluster.WorkerToMaster._
-import org.apache.gearpump.cluster.master.Master.WorkerTerminated
+import org.apache.gearpump.cluster.master.InMemoryKVService._
+import org.apache.gearpump.cluster.master.Master.{MasterInfo, WorkerTerminated}
 import org.apache.gearpump.cluster.scheduler.Scheduler.ApplicationFinished
 import org.apache.gearpump.jarstore.JarStore
 import org.apache.gearpump.util.{ActorUtil, Constants, LogUtil, Util}
@@ -48,6 +49,8 @@ private[cluster] class Master extends Actor with Stash {
   private var scheduler : ActorRef = null
 
   private var workers = new immutable.HashMap[ActorRef, Int]
+
+  private val birth : Long = System.currentTimeMillis()
 
   LOG.info("master is started at " + ActorUtil.getFullPath(context.system, self.path) + "...")
 
@@ -74,8 +77,8 @@ private[cluster] class Master extends Actor with Stash {
       self forward RegisterWorker(workerId)
     case RegisterWorker(id) =>
       context.watch(sender())
-      sender ! WorkerRegistered(id)
-      scheduler forward WorkerRegistered(id)
+      sender ! WorkerRegistered(id, MasterInfo(self, birth))
+      scheduler forward WorkerRegistered(id, MasterInfo(self, birth))
       workers += (sender() -> id)
       LOG.info(s"Register Worker $id....")
     case resourceUpdate : ResourceUpdate =>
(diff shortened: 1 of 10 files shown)


clockfly:
Can we use some format like this:
app1_201502121012
It looks more concise.

core/src/main/scala/org/apache/gearpump/cluster/worker/Worker.scala
```diff
@@ -199,7 +199,10 @@ private[cluster] object Worker {
         val classPathPrefix = Util.getCurrentClassPath ++ ctx.classPath
         val classPath = jarPath.map(classPathPrefix :+ _).getOrElse(classPathPrefix)
 
-        val logArgs = List(s"-D${Constants.GEARPUMP_APPLICATION_ID}=${launch.appId}", s"-D${Constants.GEARPUMP_EXECUTOR_ID}=${launch.executorId}")
+        val logArgs = List(
+          s"-D${Constants.GEARPUMP_APPLICATION_ID}=${launch.appId}",
+          s"-D${Constants.GEARPUMP_EXECUTOR_ID}=${launch.executorId}",
+          s"-D${Constants.GEARPUMP_APPLICATION_TIMESTAMP}=${Util.getCurrentTime()}")
```
clockfly: Executor of same app on different worker will have different timestamp?

We need a global consistent view of timestamp, better this is the time when the app is submitted to master.

core/src/main/scala/org/apache/gearpump/util/Util.scala
```diff
@@ -44,6 +45,11 @@ object Util {
     classpathList
   }
 
+  def getCurrentTime(): String = {
```
clockfly: Because we bind getCurrentTime with format "yyyy-MM-dd-hh-mm", I think it is too specific to put this in Util.

huafengw:
Now the log directory is like 
logs/
  applicationData/
     root/
       2015-02-13-11-57/
          app1

manuzhang:
why not append the timestamp to app name ? Another layer of directory makes it harder to use.

clockfly:
@manuzhang 

The timestamp is not the timestamp when the app is created. It is the timestamp when the master is alive.

manuzhang:
+1

core/src/test/scala/org/apache/gearpump/cluster/scheduler/PrioritySchedulerSpec.scala
```diff
@@ -58,7 +59,7 @@ class PrioritySchedulerSpec(_system: ActorSystem) extends TestKit(_system) with
       scheduler.tell(RequestResource(appId, request1), mockAppMaster.ref)
       scheduler.tell(RequestResource(appId, request2), mockAppMaster.ref)
       scheduler.tell(ApplicationFinished(appId), mockAppMaster.ref)
-      scheduler.tell(WorkerRegistered(workerId1), mockWorker1.ref)
+      scheduler.tell(WorkerRegistered(workerId1, MasterInfo(null, 0)), mockWorker1.ref)
```
clockfly: Can we add a constant to represent MasterInfo(null, 0) instead of writing everywhere?

core/src/main/scala/org/apache/gearpump/cluster/master/Master.scala
```diff
@@ -162,4 +166,6 @@ private[cluster] class Master extends Actor with Stash {
 
 object Master{
   case class WorkerTerminated(workerId : Int)
+
+  case class MasterInfo(master: ActorRef, startTime : Long)
```
clockfly: Is it make sense to add a default value for MasterInfo?

Like MasterInfo(master: ActorRef, startTime = 0L)

clockfly: So, that we don't need to specify the startTime in test code.

PR #450, opened 2015-02-12
fix #449: avoid load external resource

### LINKED ISSUE #449: Remove external resource
External resource might have difficulty to be loaded, so we'd use local a font-awesome package.

stanleyxu2005:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/450)


clockfly:
+1

PR #475, opened 2015-02-13
fix #474, fix rest service UT fail.

### LINKED ISSUE #474: rest service UT fails
Check https://travis-ci.org/intel-hadoop/gearpump/builds/50595725

Seems the we don't handle the rest server start/stop in good way.

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/475)


services/src/main/scala/org/apache/gearpump/services/AppMasterService.scala
```diff
@@ -34,7 +34,7 @@ import scala.util.{Failure, Success, Try}
 
 trait AppMasterService extends HttpService  {
   import upickle._
-  val master:ActorRef
+  def master:ActorRef
```
manuzhang: what's wrong with the previous way ?

stanleyxu2005:
+1

services/src/main/scala/org/apache/gearpump/services/AppMasterService.scala
clockfly: @manuzhang 

There is a contention.

Worker still have not register itself to master(because it start in async way) when the error happen.

PR #482, opened 2015-02-13
Fixes #473

### LINKED ISSUE #473: Use webjars with spray support instead of bower to serve static resources
See https://github.com/bwestlin/spray-webjars-example
All dependencies listed in bower.json are added as library dependencies in project/Build.sbt where the services project is defined: 

libraryDependencies ++= Seq(
"org.webjars" % "angular" % "1.2.18",
"org.webjars" % "jquery" % "2.0.3",
...
)

then in StaticService.scala

pathPrefix("webjars") {
  get {
    getFromResourceDirectory("META-INF/resources/webjars")
  }
}

The ```<script>``` lines in index.html will need to be modified like

```<script type="text/javascript" src="/webjars/angular/1.2.18/angular.js"></script>```

This will replace bower and move frontend dependency management entirely into sbt, spray.

kkasravi:
Fixes #473 Use webjars with spray support instead of bower to serve static resources

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/482)


kkasravi:
We may want to do this the next release since there are some version mismatches with what we use now and what's available in mvnrepository webjars. Minimally Stanley should check the UI functionality.

clockfly:
+1

For vis.js, currently the webjar version is 3.8, I have created a PR at https://github.com/webjars/visjs/pull/3 to update the webjar version to 3.10.0.

PR #484, opened 2015-02-14
Fixes #483 visdag widget does not repaint correctly when a node is moved

### LINKED ISSUE #483: visdag widget does not repaint correctly when a node is moved
I do like the fixed graph positioning which make the data flow clearer.
I'm seeing the screenshot on a mac osx 10.10.2, Chrome browser 40.0.2214.
Also the screenshot shows that the edge text can overlay other edge text.
Assigning to Stanley though I may be able to resolve it.

screen shot 2015-02-13 at 10 17 26 am

kkasravi:
I've added version 3.10.0 of vis.js to scripts/widgets/visdag since 3.8.0 had painting issues as noted in the earlier screenshot. I've also changed the layout from hubsize to direction since I think in general we want to see the logical flow of upstream nodes to downstream nodes rather than 'node with most edges' to 'nodes with least edges'. We can go back to webjars once they provide a vis.js 3.10.0. New screenshot is below 

screen shot 2015-02-13 at 5 10 21 pm

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/484)


clockfly:
+1

PR #496, opened 2015-02-16
fix #493 add implicit sender

### LINKED ISSUE #493: TaskActor fail to kick it self to start produce message
The tasks in our examples, Wordcount.Split etc. It will send a message to self to start produce messages, which is ```self ! Message("start") ```, but now the Task is not an actor any more so this tell action's sender is nosender, so that the message will be ignored by TaskActor and the whole dag fail to  start.

huafengw:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/496)


clockfly:
+1

PR #503, opened 2015-02-16
fix #502, fix KafkaStorage loading data

### LINKED ISSUE #502: KafkaStorage loading data takes a long time
KafkaStorage uses `immutable.LIst` (which has bad performance) to construct loaded data but `immutable.List`, which takes a long time (not finished for a one-minute job). That will block KafkaSource and KafkaStreamProducer, and thus the whole DAG.

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/503)


huafengw:
+1

PR #507, opened 2015-02-16
fix #504, 1) AppMaster return more detailed application runtime informat...

### LINKED ISSUE #504: AppMaster rest  service should return more detailed application data
These info are needed:

1. clock
2. executor list
3. appname

clockfly:
...ion. 2) fix a worker bug when returning executors which run on worker.

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/507)

### DIFF
--- a/core/src/main/scala/org/apache/gearpump/cluster/ClusterMessage.scala
+++ b/core/src/main/scala/org/apache/gearpump/cluster/ClusterMessage.scala
@@ -19,6 +19,7 @@
 package org.apache.gearpump.cluster
 
 import akka.actor.ActorRef
+import org.apache.gearpump.TimeStamp
 import org.apache.gearpump.cluster.master.Master.{MasterInfo, MasterDescription}
 import org.apache.gearpump.cluster.scheduler.{Resource, ResourceAllocation, ResourceRequest}
 import org.apache.gearpump.cluster.worker.WorkerDescription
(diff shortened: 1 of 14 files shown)


core/src/main/scala/org/apache/gearpump/cluster/ClusterMessage.scala
```diff
@@ -72,7 +73,9 @@ object AppMasterToMaster {
 
   case class GetAppData(appId: Int, key: String)
   case class GetAppDataResult(key: String, value: Any)
-  case class AppMasterDataDetail(appId: Int, application: Application)
+  case class AppMasterDataDetail(
+      appId: Int, appName: String = null, application: Application = null,
+      actorPath: String = null, clock: TimeStamp = 0, executors: List[String] = null)
```
huafengw: Should this class in the streaming package since the "clock" property belongs to streaming applications?

services/src/main/scala/org/apache/gearpump/services/AppMasterService.scala
```diff
@@ -47,15 +50,12 @@ trait AppMasterService extends HttpService  {
...
-                   case Some(app) =>
-                     complete(write(app))
-                   case None =>
-                     complete(StatusCodes.InternalServerError, "UserConfig is null")
-                 }
-               case Failure(ex) => complete(StatusCodes.InternalServerError, s"An error occurred: ${ex.getMessage}")
-             }
+
+                 val streamApp: StreamingAppMasterDataDetail = value
+                 complete(write(streamApp))
```
huafengw: So now we only query the info of streaming application?

clockfly:
@huafengw 

I have add some source code comments, please check the latest version.

huafengw:
+1

Issue #27, opened 2014-09-09
Integrate Gear scheduler with YARN


kkasravi:
YARN resource manager and Gearpump Master need to integrate. It may be that Gearpump Master provides a virtual resource allocation which is sync'd with actual resources allocated by YARN RM. This would allow us to include attributes not yet in YARN like SLA's. 

Requirements/Features for YARN Client and Appmaster (WIP)
- [ ] Support hadoop secure mode
- [x] Use HDFS to package and distribute the binaries
- [ ] Allow logging to occur within YARN working directory
- [ ] Configuration file needs to indicate how many workers to allocate with possible locations
- [ ] Configuration needs to be able to handle Master HA creation and deployment.
- [x] Configuration file needs to specify minimum master ram, vmcore requirements
- [x] Configuration file needs to specify minimum worker ram, vmcore requirements
- [x] Configuration file should specify where in HDFS to place jars for appmaster and workers
- [ ] Client needs to use YARN cluster API to find best nodes to run Master(s)
- [ ] AppMaster needs to use YARN cluster API to find best nodes to run Workers
- [ ] Need to be able to save configuration state so RM killing and restarting AppMaster works.

clockfly:
Currently, we can implement a fast version, by justing starting master and workers in YARN. 
We can take care of yarn allocatation delegation later.

kkasravi:
Assigning to Tomek. TIanlun - Tomek is new member and has YARN background.

devaraj-kavali:
Kam, Is it expected here to start Master and workers for each GearPump Application(i.e. one yarn application for each gearpump application) or launching GearPump master and workers one time as a long running service and handles all the gearpump applications(i.e. one yarn application for all gearpump applications).

I have just started with Gearpump, I could probably help this part once I get familiarize with the code.

kkasravi:
Hi Devaraj

[...]

PR #553, opened 2015-03-02
First attempt. AM that runs /bin/date command in yarn containers

TomaszT:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/553)


clockfly:
@TomaszT 

Thanks for your contribution.

Please take a look at project https://github.com/yahoo/storm-yarn, I think it is better for us for port from it directly. 
There are many pieces in storm-yarn that we want, like supporting hadoop secure mode, package and distribute the binaries via HDFS distributed cache, logging within YARN working directory.

kkasravi:
Hi Sean 
Welcome back from vacation! 
Storm-yarn has some interesting ideas though I would suggest we use akka instead of thrift to communicate between client and master. This would be a better fit with current design. The commands that storm uses such as add supervisors, start, stop nimbus (master), etc are good to port.

kkasravi:
I had also suggested we study slider, since this also has a dedicated protocol to do things like watch child processes and restart them.

kkasravi:
+1 

I suggest we move this into yarn branch so we have something to use as a starting point. We then work on AppMaster/Client to meet requirements that have been listed in #27. Sean I have added the requirements you listed above to the list of requirements in #27 description.

clockfly:
By the way, 

we follow scala style guide at http://docs.scala-lang.org/style/
For example, 2 space for indention, no "tab" char should be used. 

And there is a commit guideline at 
https://github.com/intel-hadoop/gearpump/blob/master/doc/CommitGuideline.md
for example: there should be bug id in commit log.

More comments about the style can be found inline.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ApplicationMaster.scala
```diff
@@ -0,0 +1,193 @@
...
+import org.apache.hadoop.yarn.api.records.FinalApplicationStatus
+
+/**
+ * Yarn ApplicationMaster. 
+ */
+class ApplicationMaster(amArgs: ApplicationMasterArguments, yarnConf : YarnConfiguration) {
+
+import org.apache.gearpump.experiments.yarn.ApplicationMaster._
+
+	private val LOG: Logger = LogUtil.getLogger(getClass)
```
clockfly: We should avoid using tab

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ApplicationMaster.scala
```diff
@@ -0,0 +1,193 @@
...
+
+/**
+ * Yarn ApplicationMaster. 
+ */
+class ApplicationMaster(amArgs: ApplicationMasterArguments, yarnConf : YarnConfiguration) {
+
+import org.apache.gearpump.experiments.yarn.ApplicationMaster._
+
+	private val LOG: Logger = LogUtil.getLogger(getClass)  
+  private val ONE_SEC_INTERVAL = 1000
```
clockfly: How about a more meaningful name instead of ONE_SEC, so  that we don't need to change the var name when we change the value from 1000

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ApplicationMaster.scala
```diff
@@ -0,0 +1,193 @@
...
+	val nmClient = new NMClientAsyncImpl(containerListener)
+	nmClient.init(getYarnConf);
+	nmClient.start();
+	nmClient    
+}
+
+private def registerAm:RegisterApplicationMasterResponse = {
+	val response = amRMClient.registerApplicationMaster("", 0, "")
+
+			LOG.info(s"Max mem and vcores capabilities : " +
```
clockfly: not aligned

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ApplicationMaster.scala
```diff
@@ -0,0 +1,193 @@
...
+    LOG.info("Requesting containers")
+    requestContainers(registerAm)
+    
+    LOG.info("Stopping application master")
+    waitForContainers
+    waitForLauncherThreads
+    cleanUp 
+  }
+
+  private def initialize {
```
clockfly: All function should define a return value.

If there is no return, we should define it explicitly as

def initilaize: Unit =

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ApplicationMaster.scala
```diff
@@ -0,0 +1,193 @@
...
+    nmClientAsync = getNMClient(nmCallbackHandler)
+    rmCallbackHandler = new RMCallbackHandler(nmClientAsync, nmCallbackHandler, this)
+    amRMClient = getAMRMClient(rmCallbackHandler)
+
+  }
+
+//TODO: refactor, future instead of thread
+ private def waitForContainers {
+   while (!rmCallbackHandler.isDone && (rmCallbackHandler.completedContainersCount.get() != amArgs.containers)) {
+	   Thread.sleep(200);
```
clockfly: 200 is a constants, should be named

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ApplicationMaster.scala
```diff
@@ -0,0 +1,193 @@
...
+import org.apache.hadoop.yarn.api.records.NodeReport
+import org.apache.hadoop.yarn.api.records.ContainerExitStatus
+import org.apache.hadoop.yarn.api.records.ContainerLaunchContext
+import org.apache.hadoop.yarn.api.records.LocalResource
+import org.apache.hadoop.yarn.api.records.FinalApplicationStatus
+
+/**
+ * Yarn ApplicationMaster. 
+ */
+class ApplicationMaster(amArgs: ApplicationMasterArguments, yarnConf : YarnConfiguration) {
```
clockfly: Maybe rename this to YarnAppMaster so that we don't confuse this with Gearpupump AppMaster

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/RMCallbackHandler.scala
```diff
@@ -0,0 +1,140 @@
...
+}
+
+def onNodesUpdated(updatedNodes: java.util.List[NodeReport]) {}
+
+def onShutdownRequest()  {
+	LOG.info("Shutdown requested")        
+	done = true
+}
+
+private[RMCallbackHandler] class ContainerLauncher(container: Container, nmClientAsync:NMClientAsync, containerListener: NMCallbackHandler) extends Runnable {
```
clockfly: Maybe put ContainerLauncher under companion object better?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/RMCallbackHandler.scala
```diff
@@ -0,0 +1,140 @@
...
+  def run() {
```
clockfly: def run: Unit =

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ApplicationMaster.scala
```diff
@@ -0,0 +1,193 @@
...
+import org.apache.gearpump.experiments.yarn.ApplicationMaster._
+
+	private val LOG: Logger = LogUtil.getLogger(getClass)  
+  private val ONE_SEC_INTERVAL = 1000
+  private var amRMClient: AMRMClientAsync[ContainerRequest] = _
+  private var nmCallbackHandler: NMCallbackHandler = _
+  private var nmClientAsync: NMClientAsync = _ 
+  private var rmCallbackHandler: RMCallbackHandler = _
+  
+  def getYarnConf = yarnConf
```
clockfly: Is this a public function? 

If so, make sure we have clear return value declaration.

Issue #458, opened 2015-02-12
Evaluate using scala.js in frontend. List pro's/con's.


kkasravi:
Scala.js 1.6 is out and is no longer experimental. We should look at this and decide if it makes sense to move to it.

kkasravi:
An alternate possibility is to use Babel. https://babeljs.io/

kkasravi:
We should investigate a partial solution of just using scala.js for JSON serde if this allows us not to do JSON serde manually. This would be a big advantage.

clockfly:
We need some proof of concept to demo it actually works for ser/deser

kkasravi:
Replaced by #952 which contains the pivotal story reference.

PR #559, opened 2015-03-04
fix #558, [REST]Add missing "name" field for metric Histogram

### LINKED ISSUE #558: [REST]Add missing "name" field for metric Histogram

clockfly:
Review link: https://github.com/intel-hadoop/gearpump/commit/6860731ae060a822dead97b89a7a6ef4ef88fced

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/559)

### DIFF
--- a/core/src/main/scala/org/apache/gearpump/cluster/ClusterMessage.scala
+++ b/core/src/main/scala/org/apache/gearpump/cluster/ClusterMessage.scala
@@ -77,9 +77,21 @@ object AppMasterToMaster {
 
   //TODO:
   // clock field may not make sense for applications other than streaming
-  case class AppMasterDataDetail(
-      appId: Int, appName: String = null, application: Application = null,
-      actorPath: String = null, clock: TimeStamp = 0, executors: List[String] = null)
+  trait AppMasterDataDetail {
+    def appId: Int
+    def appName: String
+    def actorPath: String
+    def executors: List[String]
+    def toJson: String
+  }
+
+  case class GeneralAppMasterDataDetail(
+      appId: Int, appName: String = null, actorPath: String = null, executors: List[String] = null)
+    extends AppMasterDataDetail {
+    def toJson: String = {
+      upickle.write(this)
+    }
+  }
 
   case object GetAllWorkers
   case class GetWorkerData(workerId: Int)
--- a/core/src/main/scala/org/apache/gearpump/metrics/Metrics.scala
+++ b/core/src/main/scala/org/apache/gearpump/metrics/Metrics.scala
@@ -48,16 +48,34 @@ class Metrics(sampleRate: Int) extends Extension {
 }
 
 object Metrics extends ExtensionId[Metrics] with ExtensionIdProvider {
+
   val LOG: Logger = LogUtil.getLogger(getClass)
   import org.apache.gearpump.util.Constants._
 
   sealed trait MetricType
-  case class Histogram(count: Long, min: Long, max: Long, mean: Double, stddev: Double, median: Double, p75: Double, p95: Double, p98: Double, p99: Double, p999: Double) extends MetricType
+
+  case class Histogram
+      (name: String, count: Long, min: Long, max: Long, mean: Double,
+       stddev: Double, median: Double, p75: Double,
+       p95: Double, p98: Double, p99: Double, p999: Double)
+    extends MetricType
+
   case class Counter(name: String, value: Long) extends MetricType
-  case class Meter(name: String, count: Long, meanRate: Double, m1: Double, m5: Double, m15: Double, rateUnit: String) extends MetricType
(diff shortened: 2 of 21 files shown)


core/src/main/scala/org/apache/gearpump/cluster/ClusterMessage.scala
```diff
@@ -77,9 +77,21 @@ object AppMasterToMaster {
...
+    def toJson: String
+  }
+
+  case class GeneralAppMasterDataDetail(
+      appId: Int, appName: String = null, actorPath: String = null, executors: List[String] = null)
+    extends AppMasterDataDetail {
+    def toJson: String = {
+      upickle.write(this)
+    }
+  }
```
kkasravi: Why toJson?

core/src/main/scala/org/apache/gearpump/metrics/Metrics.scala
```diff
@@ -48,16 +48,34 @@ class Metrics(sampleRate: Int) extends Extension {
...
+
+  case class Timer(
+      name: String, count: Long, min: Double, max: Double,
+      mean: Double, stddev: Double, median: Double,
+      p75: Double, p95: Double, p98: Double,
+      p99: Double, p999: Double, meanRate: Double,
+      m1: Double, m5: Double, m15: Double,
+      rateUnit: String, durationUnit: String)
+    extends MetricType
+
```
kkasravi: I was not able to do Gauge[T] correctly (although we do not use it). Any ideas on how to handle [T]?

services/src/main/scala/org/apache/gearpump/services/AppMasterService.scala
```diff
@@ -51,13 +54,7 @@ trait AppMasterService extends HttpService  {
...
              onComplete((master ? AppMasterDataDetailRequest(appId)).asInstanceOf[Future[AppMasterDataDetail]]) {
                case Success(value: AppMasterDataDetail) =>
-                 if (isStreamingApplication(value)) {
-                   val streamApp: StreamingAppMasterDataDetail = value
-                   complete(write(streamApp))
-                 } else {
-                    val app: GeneralAppMasterDataDetail = value
-                   complete(write(app))
-                 }
+                 complete(value.toJson)
```
kkasravi: Oh i see why toJson is used now. Can we only import upickle in ClusterMessage? That would nicely isolate JSON conversion to just messages which would allow us to change JSON serde easily.

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/StreamingAppMasterDetail.scala
```diff
@@ -0,0 +1,87 @@
...
+        ("appId", appId),
+        ("appName", appName),
+        ("actorPath", actorPath),
+        ("clock", clock),
+        ("executors", executorsSeq.getOrElse(Js.Null)),
+        ("processors", processors),
+        ("dag", dag)
+      )
+  }
+}
```
kkasravi: This is a nice way to isolate the DAG serde and well as add new info. I guess we will also need a reader to accept DAG mods. I will look into autogen.

services/src/main/scala/org/apache/gearpump/services/AppMasterService.scala
clockfly: Good idea!

I would suggest to wrap the json library in a Util in the future .
Like 

object JsonUtil {
  def toJson(obj: T)
  def fromJson(json: String): T
}

kkasravi:
Failure in AppMasterSpec? 

+1 - few questions noted. Also we should look into using upickle on client side if we move to scala.js. It may be possible to use scala.js (and upickle) just for JSON serde if it allows us not to do JSON deserde manually on client side. Perhaps open an issue? I've added a note to #458.

services/src/main/scala/org/apache/gearpump/services/AppMasterService.scala
clockfly: I have created a new issue #560 for this.

clockfly:
Yes, I have created a issue #560, and linked your comments there.

core/src/main/scala/org/apache/gearpump/metrics/Metrics.scala
clockfly: What problems you encoutered?

clockfly:
The failure seems random. I restarted the UT.

core/src/main/scala/org/apache/gearpump/cluster/ClusterMessage.scala
clockfly: Different app type has different type of AppMasterDetail, so it is not reasonble to assure upickle can serialize every sub-class.

The toJson will requires the user to define a explict serializer.

clockfly:
All pass, I will merge

PR #572, opened 2015-03-05
fix #568, [REST] add history metrics service in backend so that the UI d...

### LINKED ISSUE #568: [REST] provide history metrics service
The REST should allow user to query history metrics data, so that user can view the history metrics timecharts in UI dashboard.

clockfly:
...ashboard can query and present the timecharts which allow user to browser by specifying a time window.

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/572)

### DIFF
--- a/streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/HistoryMetricsService.scala
+++ b/streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/HistoryMetricsService.scala
@@ -1,0 +1,287 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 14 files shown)


streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/HistoryMetricsService.scala
```diff
@@ -0,0 +1,280 @@
...
+    override def add(inputMetrics: MetricType): Unit = {
+      history.add(inputMetrics)
+      recent.add(inputMetrics)
+    }
+
+    override def read: List[HistoryMetricsItem] = {
+      history.read ++ recent.read
+    }
+  }
+}
```
kkasravi: Should we keep RETAIN_* values in a conf somewhere? Also do we want a REST/WS interface to retrieve/change these values?

streaming/src/test/scala/org/apache/gearpump/streaming/appmaster/HistoryMetricsServiceSpec.scala
```diff
@@ -0,0 +1,205 @@
...
+    //expire oldest data point, because we only keep @count records
+    store.add(Counter("count", 5))
+    store.add(Counter("count", 6))
+
+    val result = store.read
+    assert(result.size == count)
+
+    //the oldest value is expired
+    assert(result.head.value.asInstanceOf[Counter].value == 3L)
+
```
huafengw: HistoryMetricsServiceSpec failed.

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/HistoryMetricsService.scala
clockfly: Currently, we store everything in memory. It is not safe at current point to expose this config to user. 
So, I prefer to make it static if for now.

clockfly: After rethinking this, I will make them configurable, and add a comment that this value should not be set to too big, due to implementation limit.

clockfly:
@kkasravi 

comments fixed

kkasravi:
Failed test, otherwise +1 !

PR #574, opened 2015-03-06
fix #562 fix AppMaster and Executor restart infinitely

### LINKED ISSUE #562: Gearpump will restart failed executors infinitely
When executor crash or appmaster crash, then it will trigger recovery process.
The current problem is that, if the error persist(for example, user code contains exception that is thrown everytime), then we will always trigger recovery, which may generate too much log files on disks.

The solution is to set a retry limit for the recovery. If we cannot recovery the app/executor/task after xx max tries, then we should stop and report this error to user.

huafengw:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/574)

### DIFF
--- a/core/src/main/scala/org/apache/gearpump/util/RestartPolicy.scala
+++ b/core/src/main/scala/org/apache/gearpump/util/RestartPolicy.scala
@@ -1,0 +1,36 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.gearpump.util
+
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/apache/gearpump/util/RestartPolicy.scala
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * @param maxNrOfRetries the number of times is allowed to be restarted, negative value means no limit,
+ *   if the limit is exceeded the policy will not allow to restart
+ * @param withinTimeRange duration of the time window for maxNrOfRetries, Duration.Inf means no window
+ */
+class RestartPolicy (maxNrOfRetries: Int, withinTimeRange: Duration) {
+  private val status = new ChildRestartStats(null, 0, 0L)
+  private val retriesWindow = (Some(maxNrOfRetries), Some(withinTimeRange.toMillis.toInt))
+
+  def allowRestart: Boolean = {
```
clockfly: maybe change this to requestRestart?

streaming/src/main/scala/org/apache/gearpump/streaming/executor/ExecutorRestartPolicy.scala
```diff
@@ -0,0 +1,56 @@
...
+  def addTaskToExecutor(executorId: Int, taskId: TaskId): Unit = {
+    var taskSetForExecutorId = executorToTaskIds.getOrElse(executorId, Set.empty[TaskId])
+    taskSetForExecutorId += taskId
+    executorToTaskIds += executorId -> taskSetForExecutorId
+    if (!taskRestartPolocies.contains(taskId)) {
+      taskRestartPolocies += taskId -> new RestartPolicy(maxNrOfRetries, withinTimeRange)
+    }
+  }
+
+  def allowRestartExecutor(executorId: Int): Boolean = {
```
clockfly: maybe change to requestPermissionToRestartExecutor

clockfly:
+1

PR #582, opened 2015-03-06
Fixes #27 Integrate GearPump scheduler into YARN

kkasravi:
Tomek and I are iterating over this. Second pass. Conversion to actors.
No kerberos yet (Tomek is adding). Tomek may also split this file into
separate files and make other changes as suggested in upcoming followup
comments by Sean et al.

Actor message flow is below:
screen shot 2015-03-06 at 1 15 56 pm

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/582)

### DIFF
--- a/experiments/yarn/src/main/resources/reference.conf
+++ b/experiments/yarn/src/main/resources/reference.conf
@@ -6,9 +6,9 @@ gearpump {
       excludejars = "slf4j-log4j12-1.7.7.jar"
     }
     applicationmaster {
-      name = "gearpump"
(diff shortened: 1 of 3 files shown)


kkasravi:
Oh, I should note that this PR is mostly soliciting some feedback from GearPump team, we haven't added any testing yet and haven't verified behavior within a yarn cluster but we're testing within a yarn cluster. There is still some significant features to implement - like actually running the GearPump master.

PR #585, opened 2015-03-09
fix #584, KafkaOffsetManager only stores one offset per timestamp

### LINKED ISSUE #584: kafka offset manager only needs to store one offset per timestamp

manuzhang:
fix #584, KafkaOffsetManager only stores one offset per timestamp

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/585)


clockfly:
+1

PR #599, opened 2015-03-11
issue #27 - setting up commands for the master, worker(s) and services (...

TomaszT:
...REST server).

Commands we need to invoke :
$JAVA_HOME/bin/java org.apache.gearpump.cluster.main.Master -ip <hostip> -port <port>
$JAVA_HOME/bin/java org.apache.gearpump.cluster.main.Worker -master <host>:<port>
$JAVA_HOME/bin/java org.apache.gearpump.cluster.main.Services -master <host>:<port>

I've extracted class AppConfig. Parameters are pulled from reference.conf but it is possible to override them with command line, like so :
$JAVA_HOME... configuration.yarn.container.count 4 configuration.yarn.container.vcores 2
(a little bit cumbersome)
Maybe we should create some kind of short alias name for cmdLine parameters that will be often overridden to make this more convenient. Will come back to this in the next iteration. 
Please share if you have some other ideas.

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/599)


experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/AppConfing.scala
```diff
@@ -0,0 +1,50 @@
+package org.apache.gearpump.experiments.yarn
```
clockfly: Filename typo.

license header

TomaszT: Done

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/AppConfig.scala
```diff
@@ -0,0 +1,68 @@
...
+  val YARNAPPMASTER_MAIN = "configuration.yarn.applicationmaster.main"
+  val CONTAINER_COUNT = "configuration.yarn.container.count"
+  val CONTAINER_MEMORY = "configuration.yarn.container.memory"
+  val CONTAINER_VCORES = "configuration.yarn.container.vcores"
+  val EXCLUDE_JARS = "configuration.yarn.client.excludejars"
+  val HDFS_PATH = "configuration.yarn.client.hdfsPath"
+  val JARS = "configuration.yarn.client.jars"
+  val GEARPUMPMASTER_COMMAND = "configuration.gearpump.master.command"
+  val GEARPUMPMASTER_MAIN = "configuration.gearpump.master.main"
+  val GEARPUMPMASTER_IP = "configuration.gearpump.master.ip"
```
clockfly: Is this two config used? 
Otherwise we may want to remove GEARPUMPMASTER_IP  and GEARPUMPMASTER_PORT

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/AppConfig.scala
```diff
@@ -0,0 +1,68 @@
...
+  val YARNAPPMASTER_QUEUE = "configuration.yarn.applicationmaster.queue"
+  val YARNAPPMASTER_MAIN = "configuration.yarn.applicationmaster.main"
+  val CONTAINER_COUNT = "configuration.yarn.container.count"
+  val CONTAINER_MEMORY = "configuration.yarn.container.memory"
+  val CONTAINER_VCORES = "configuration.yarn.container.vcores"
+  val EXCLUDE_JARS = "configuration.yarn.client.excludejars"
+  val HDFS_PATH = "configuration.yarn.client.hdfsPath"
+  val JARS = "configuration.yarn.client.jars"
+  val GEARPUMPMASTER_COMMAND = "configuration.gearpump.master.command"
+  val GEARPUMPMASTER_MAIN = "configuration.gearpump.master.main"
```
clockfly: I didn't see this constant be used anywhere? Some code missing?

experiments/yarn/src/main/resources/reference.conf
```diff
@@ -1,4 +1,4 @@
-gearpump {
+configuration {
```
clockfly: There is a tradition to put every config under path gearpump, so that we can easy filter out what configs we provides.

Would you mind put every config in this project under 

gearpump {
  yarn {
    //other configs
  }
}

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/AppConfig.scala
```diff
@@ -0,0 +1,68 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.gearpump.experiments.yarn
+
+import org.apache.gearpump.cluster.main.{ParseResult, ArgumentsParser, CLIOption}
+import com.typesafe.config.Config
+
+object EnvVars {
```
clockfly: Is EnvVars  specifically used in class AppConfig or also used else where?

If this is also used else where, I will recommend to store the EnvVars to a stand alone with a more specific name like YarnConstants

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/AppConfig.scala
```diff
@@ -0,0 +1,68 @@
...
+object CmdLineVars {
+  val APPMASTER_IP = "ip"
+  val APPMASTER_PORT = "port"
+}
+
+class AppConfig(cliopts: ParseResult, conf: Config) {
+  
+   def getEnv(key: String): String = {
+    Option(cliopts) match {
+      case Some(_cliopts) =>
```
clockfly: One small tip, you can use cliopts as name. When resolution, it will find the var in the insider scope.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/Client.scala
```diff
@@ -48,45 +48,32 @@ Features for YARNClient
...
-  val APPMASTER_MAIN = "gearpump.yarn.applicationmaster.main"
-  val CONTAINER_COUNT = "gearpump.yarn.container.count"
-  val CONTAINER_MEMORY = "gearpump.yarn.container.memory"
-  val CONTAINER_VCORES = "gearpump.yarn.container.vcores"
-  val EXCLUDE_JARS = "gearpump.yarn.client.excludejars"
-  val HDFS_PATH = "gearpump.yarn.client.hdfsPath"
-  val JARS = "gearpump.yarn.client.jars"
-}
 
 trait ClientAPI {
```
clockfly: Is it reasonble to add some documents in the public trait?

So that the document can show up the scala doc.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/AppConfig.scala
kkasravi: Also you can use ``` `cliopts` ``` which prevents the warning ['Suspicious shadowing by a Variable Pattern'](http://stackoverflow.com/questions/7905023/in-scala-pattern-matching-what-is-suspicious-shadowing-by-a-variable-pattern). EG
```
   def getEnv(key: String): String = {
    Option(cliopts) match {
      case Some(`cliopts`) =>
        `cliopts`.exists(key) match {
          case true =>
            `cliopts`.getString(key)
          case false =>
            conf.getString(key)
        }
      case None =>
        conf.getString(key)
    }
```

TomaszT:
I'll create another PR for issue #27

PR #614, opened 2015-03-16
fix#608 increase the maximum frame size of Akka

huafengw:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/614)


stanleyxu2005:
+1

PR #616, opened 2015-03-16
fix #615

### LINKED ISSUE #615: Incorrect Dag edge width
The representation of node throughput looks wired. The lower side might have more throughput then its upper side. After debugging the width (throughput) of edges are not calculated correctly.

stanleyxu2005:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/616)


huafengw:
+1

PR #622, opened 2015-03-19
fix#618 fix no data returned when call Metrics rest

huafengw:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/622)


stanleyxu2005:
+1

PR #623, opened 2015-03-20
fix #434 add api/v1.0 prefix for all rest services

### LINKED ISSUE #434: Add api/v1.0/ prefix for all REST related calls
We should add "api" and a version path segment

huafengw:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/623)


clockfly:
+1

PR #655, opened 2015-03-24
fix #654, Use yarn to distribute whole gearpump package instead of jars under lib/. load log4j.properties into classpath so that we have logs.

### LINKED ISSUE #654: [yarn] Use yarn to distribute whole gearpump package instead of jars under lib/

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/655)

### DIFF
--- a/experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/YarnContainerUtil.scala
+++ b/experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/YarnContainerUtil.scala
@@ -21,9 +21,7 @@ package org.apache.gearpump.experiments.yarn
 import org.apache.gearpump.util.LogUtil
 import org.apache.hadoop.fs.FileSystem
 import org.apache.hadoop.fs.Path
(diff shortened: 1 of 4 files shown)


experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/YarnContainerUtil.scala
```diff
@@ -44,11 +39,10 @@ object YarnContainerUtil {
   val LOG: Logger = LogUtil.getLogger(getClass)
   
   //TODO: move to config
-  val HDFS_JARS_DIR = "/user/gearpump/jars/"
+  val HDFS_ROOT = "/user/gearpump/"
```
huafengw: Extract this config

clockfly: Let's do this in fix https://github.com/intel-hadoop/gearpump/issues/658

huafengw:
+1

PR #668, opened 2015-03-24
Fixes #653 Start services in Yarn Application Master and register it's U...

### LINKED ISSUE #653: Start services in Yarn Application Master and register it's URL as the trackingURL
- [ ] set trackingURL when calling AMRMClientAsync.registerApplicationMasterRequest
- [ ] keep track of containerId's when launching master containers
- [ ] in AmActor.receive ContainerStarted when all master containers have started, start services

kkasravi:
...RL as the trackingURL

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/668)

### DIFF
--- a/core/src/main/scala/org/apache/gearpump/cluster/main/ArgumentsParser.scala
+++ b/core/src/main/scala/org/apache/gearpump/cluster/main/ArgumentsParser.scala
@@ -27,7 +27,7 @@ class ParseResult(optionMap : Map[String, String], remainArguments : Array[Strin
 
   def getBoolean (key : String) = optionMap.get(key).get.toBoolean
 
-  def exists(key : String) = optionMap.get(key).isDefined
+  def exists(key : String) = !optionMap.getOrElse(key,"").isEmpty
 
   def remainArgs : Array[String] = this.remainArguments
 }
--- a/services/src/main/scala/org/apache/gearpump/cluster/main/Services.scala
+++ b/services/src/main/scala/org/apache/gearpump/cluster/main/Services.scala
@@ -19,6 +19,7 @@
 package org.apache.gearpump.cluster.main
 
 import akka.actor.{ActorSystem, Props}
+import com.typesafe.config.{Config, ConfigFactory}
 import org.apache.gearpump.cluster.main.Local._
 import org.apache.gearpump.cluster.master.MasterProxy
 import org.apache.gearpump.cluster.{ClusterConfig, UserConfig}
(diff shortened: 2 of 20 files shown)


kkasravi:
This is quite there - I'm getting an error when the GearPump services tries to show the UI as shown below. Still debugging this.

screen shot 2015-03-24 at 3 26 54 am

clockfly:
Hi Kam,

Have you set the the following config correctly in service UI?
```

  ##############################
  ### Required to change!!
  ### You need to set the ip address or hostname of this machine
  ###
  akka.remote.netty.tcp.hostname = "127.0.0.1"

```

You can simplily override the gear.conf by providing a java property. like this:
```
-Dakka.remote.netty.tcp.hostname=my_hostname_or_ip_of_ui_server
```

kkasravi:
It seems to be related to how the trackingURL is set - in the YARN UI - clicking on the trackingURL goes through a proxy that looks like http://ip-10-10-10-217.eu-west-1.compute.internal:8088/proxy/application_1426259233511_0045/#/cluster/master
even though I set the trackingURL to the correct services URL.

clockfly:
The web proxy may not work.

As we have REST call and long connection call, I am not sure it will still work when the YARN rewrite the REST URL in web page.

clockfly:
Another trick I use when I used before is to add a jump page. 
In the jump page, it will use window.location="new url" to redirect to actual UI server address.

You may want to give a try on this.

services/src/main/scala/org/apache/gearpump/cluster/main/Services.scala
```diff
@@ -37,18 +38,58 @@ object Services extends App with ArgumentsParser {
   }
 
   override val options: Array[(String, CLIOption[Any])] = Array(
-    "master" -> CLIOption[String]("<host1:port1,host2:port2,host3:port3>", required = true))
+    "master" -> CLIOption[String]("<host1:port1,host2:port2,host3:port3>", required = true),
```
clockfly: I don't see necessity to add new cli option.

You can use java -D property, it will be parsed by typesafe Config and override the setting in gear.conf.

core/src/main/scala/org/apache/gearpump/cluster/main/ArgumentsParser.scala
```diff
@@ -27,7 +27,17 @@ class ParseResult(optionMap : Map[String, String], remainArguments : Array[Strin
 
   def getBoolean (key : String) = optionMap.get(key).get.toBoolean
 
-  def exists(key : String) = optionMap.get(key).isDefined
+  def exists(key : String) = optionMap.get(key) match {
+    case Some(value) =>
+      value.length match {
```
clockfly: value.empty better?

services/src/main/scala/org/apache/gearpump/cluster/main/Services.scala
kkasravi: But we are not running services in a separate VM - rather we are launching services within YarnAM

core/src/main/scala/org/apache/gearpump/cluster/main/ArgumentsParser.scala
kkasravi: Fixed - thanks!

kkasravi:
This now works - the problem was we were not correctly setting restapiRoot

diff --git a/services/dashboard/views/dashboard.js b/services/dashboard/views/dashboard.js
index aff0b20..fa52249 100644
--- a/services/dashboard/views/dashboard.js
+++ b/services/dashboard/views/dashboard.js
@@ -30,7 +30,7 @@ angular
     updateChartInterval: 2000,
     updateVisDagInterval: 2000,
     restapiAutoRefreshInterval: 2000,
-    restapiRoot: '/api/v1.0',
+    restapiRoot: location.pathname + 'api/v1.0',
     webSocketSendTimeout: 500
   })

You can now select the trackingURI in the YARN RM UI and it will bring up the GearPump UI!

services/src/main/scala/org/apache/gearpump/cluster/main/Services.scala
clockfly: Same story. You can use System.setProperty() to add that config, the properties will get loaded in Service automatically.

clockfly:
Hi Kam,

Is this ready for merge?

kkasravi:
see PR https://github.com/intel-hadoop/gearpump/pull/702

PR #669, opened 2015-03-25
fix #658 make config of Gearpump on Yarn configurable

### LINKED ISSUE #658: [yarn] Make yarn related config dynamic
Currently, we are hardcoding everything in reference,conf, we should make them dynamic so that user can specify a app config via yarnClient

huafengw:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/669)

### DIFF
--- a/experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
+++ b/experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
@@ -40,6 +40,7 @@ import org.apache.gearpump.experiments.yarn.NodeManagerCallbackHandler
 import org.apache.gearpump.experiments.yarn.ResourceManagerClientActor
 import org.apache.gearpump.util.LogUtil
 import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{FileUtil, FileSystem, Path}
 import org.apache.hadoop.net.NetUtils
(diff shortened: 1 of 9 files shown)


experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
```diff
@@ -341,7 +341,13 @@ object YarnApplicationMaster extends App with ArgumentsParser {
         LOG.error("Caught exception", throwable)
         throwable.printStackTrace()
     }
+  }
 
+  private def fetchConfigOnHDFS(path: String) = {
```
clockfly: Please use HDFs distributed cache to distribute the conf.

1. Create a conf folder under HDFS://user/geapump/
2. upload the jars to that hdfs conf folder.
3. Add the conf folder to local Resource
Example:
```
+    Map(
+      "pack" -> newYarnAppResource(fs, new Path(s"$hdfsRoot/$version.tar.gz"),
+        LocalResourceType.ARCHIVE, LocalResourceVisibility.PUBLIC))
       conf" -> newYarnAppResource(fs, new Path(s"$hdfsRoot/conf"),
+        LocalResourceType.FILETYPE, LocalResourceVisibility.PUBLIC))
```
4. Then during runtime, the conf folder will be under container working directory.

Then you need to add conf into classpath of YarnAM, Master, and Worker.

clockfly:
+1

PR #674, opened 2015-03-25
fix #672, extend Task to support unmanaged messages so that every task can serve as full functional service

### LINKED ISSUE #672: [Streaming] extend Task to support unmanaged messages so that every task can serve as full functional service
In streaming module, we support fault-torellance of message delivery. The message type, and the message delivery routine must match the predefined DAG denition and runtime deployment so that we can protect against zombie tasks.

Sometimes, this is not that flexible. For example, we may want to drop a message to any task and query the in memory task status. In this case, the message type and message delivery routine is not pre-determined. 

So we may want to introduce the concept of un-managed message.

For managed message, the streaming framework will take over at-least once(or exactly once) delivery, and will be resilient to all kinds of failures like network down, jvm crash, message loss, and etc..

For un-managed message, the streaming framework have no responsibility. 

The updated task interface looks like this:
```
/**
 * Streaming Task interface
 */
trait TaskInterface {

  /**
   * @param startTime startTime that can be used to decide from when a source producer task should replay the data source, or from when a processor task should recover its checkpoint data in to in-memory state.
   */
  def onStart(startTime : StartTime) : Unit

  /**
   * @param msg message send by upstream tasks
   */
  def onNext(msg : Message) : Unit

  /**
   * This can be used to cleanup resource when the application finished.
   */
  def onStop() : Unit

  /**
   * handler for unmanaged message
   * @return
   */
  def receiveUnManagedMessage: Receive = null
}
```

For message whose type is not classOf[Message], then receiveUnManagedMessage will be used to handle the message.

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/674)

### DIFF
--- a/streaming/src/main/scala/org/apache/gearpump/streaming/task/Task.scala
+++ b/streaming/src/main/scala/org/apache/gearpump/streaming/task/Task.scala
@@ -18,6 +18,7 @@
 
 package org.apache.gearpump.streaming.task
 
+import akka.actor.Actor.Receive
 import akka.actor.{Cancellable, Props, ActorRef, ActorSystem}
 import org.apache.gearpump.Message
 import org.apache.gearpump.cluster.UserConfig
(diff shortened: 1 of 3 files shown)


clockfly:
Motivation: #672

manuzhang:
+1

PR #694, opened 2015-03-26
fix #638, Use Subscription to link two processor

### LINKED ISSUE #638: [dynamic dag]Use Subscription to link two processor
Child of #101

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/694)


clockfly:
This is a sub-task for dynamic dag.

streaming/src/main/scala/org/apache/gearpump/streaming/task/Subscriber.scala
```diff
@@ -0,0 +1,62 @@
...
+
+    edges.foldLeft(List.empty[Subscriber]) { (list, nodeEdgeNode) =>
+      val (_, partitioner, downstreamProcessorId) = nodeEdgeNode
+      val downstreamProcessor = dag.processors(downstreamProcessorId)
+      list :+ Subscriber(downstreamProcessorId, partitioner, downstreamProcessor)
+    }
+  }
+}
+
+case class Subscribe(processorId: Int, subscriber: Subscriber)
```
manuzhang: How are these two used ?

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/TaskSchedulerImpl.scala
```diff
@@ -174,5 +175,5 @@ class TaskSchedulerImpl(appId : Int, config: Config)  extends TaskScheduler {
 }
 
 object TaskSchedulerImpl {
-  case class TaskLaunchData(taskId: TaskId, taskDescription : TaskDescription, dag : DAG)
+  case class TaskLaunchData(taskId: TaskId, taskDescription : TaskDescription, subscriptions: List[Subscriber])
```
whjiang: Shall we add an application field here to provide general info about an application, e.g. user name, credential, config...?

streaming/src/main/scala/org/apache/gearpump/streaming/task/Subscriber.scala
whjiang: What's the relationship between Subscriber and Processor?

clockfly: Subscriber = processorId +  edge(partitioner) + processorDefinition.

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/TaskSchedulerImpl.scala
clockfly: We already have the data channel via executor.

streaming/src/main/scala/org/apache/gearpump/streaming/task/Subscriber.scala
clockfly: processorId is dynamically generated.
partitioner describe relation between processor.

Basically, with Subscriber, we can turn DAG[Processor, Partitioner] to List[((processorId, Processor), List[Subscriber])]

This will flatten the DAG to make it more easy to be used for single processor.

clockfly:
Anyone help to review this?

clockfly:
This is the first step for dynamic dag.

This PR decouple the task with DAG. Each Task will be given downstream processor information with subscription on a need-to-know basis, and in future the subscription can be dynamically added or removed.

services/src/main/scala/org/apache/gearpump/services/ActorUtilService.scala
```diff
@@ -0,0 +1,67 @@
...
+  implicit val system: ActorSystem
+  private val LOG = LogUtil.getLogger(getClass)
+
+  def actorUtilRoute = {
+    implicit val ec: ExecutionContext = actorRefFactory.dispatcher
+    implicit val timeout = Constants.FUTURE_TIMEOUT
+    pathPrefix("api"/s"$REST_VERSION") {
+
+      // this is a internal API
+      path("internal" / "actorutil") {
```
whjiang: better to add some flag to turn this test util off.

conf/gear.conf
```diff
@@ -29,14 +32,20 @@ gearpump {
...
-  netty-dispatcher = "akka.actor.default-dispatcher"
+  task-dispatcher {
+    mailbox-type = "akka.dispatch.SingleConsumerOnlyUnboundedMailbox"
+    throughput = 100
+    fork-join-executor {
+      parallelism-factor = 2
+      parallelism-max = 8
+      parallelism-min = 2
+    }
+  }
```
manuzhang: why do we change the settings here ?

core/src/main/scala/org/apache/gearpump/transport/netty/Client.scala
```diff
@@ -113,7 +112,7 @@ class Client(conf: NettyConfig, factory: ChannelFactory, hostPort : HostPort) ex
     }
   }
 
-  private def send(msgs: Iterator[TaskMessage]) {
+  private def send(channel: Channel, msgs: Iterator[TaskMessage]) {
```
manuzhang: Could we use a different name than the global `channel` ?

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/TaskSchedulerImpl.scala
```diff
@@ -81,11 +81,12 @@ class TaskSchedulerImpl(appId : Int, config: Config)  extends TaskScheduler {
...
     dag.processors.foreach { params =>
-      val (taskGroupId, taskDescription) = params
+      val (processorId, taskDescription) = params
       0.until(taskDescription.parallelism).map((taskIndex: Int) => {
-        val taskId = TaskId(taskGroupId, taskIndex)
+        val taskId = TaskId(processorId, taskIndex)
         val locality = taskLocator.locateTask(taskDescription)
-        val taskLaunchData = TaskLaunchData(taskId, taskDescription, dag.subGraph(taskGroupId))
+        val taskLaunchData = TaskLaunchData(taskId, taskDescription,
+          Subscriber.of(processorId, dag))
```
whjiang: shall we move Subscriber.of(processorId, dag) out of the loop? It seems it is processor specific instead of task specific.

conf/gear.conf
clockfly: The old behavior is to use pin-dispatcher, If there are multiple(suppose 1000 tasks), then the performance will be very bad(as it will try to create 1000 threads)

Now, the tasks will share the same thread pool.

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/TaskSchedulerImpl.scala
clockfly: done

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/ClockService.scala
```diff
@@ -75,10 +82,12 @@ class ClockService(dag : DAG, store: AppDataStore) extends Actor with Stash {
       processorIdWithDescription =>
         val (processorId, description) = processorIdWithDescription
         val taskClocks = new Array[TimeStamp](description.parallelism).map(_ => startClock)
-        val processorClock = new ProcessorClock(processorId, startClock, taskClocks)
-        processorClocks.add(processorClock)
+        val processorClock = new ProcessorClock(taskClocks)
         processorClockLookup.put(processorId, processorClock)
     }
+
+    this.processorIdToLevel = dag.graph.topologicalOrderIterator.zipWithIndex.toMap
```
manuzhang: Given how they are generated, are processId and level the same value ?

clockfly: For now yes(but topology sort is unstable). But in future, we will different processor can share same level Id.

streaming/src/main/scala/org/apache/gearpump/streaming/task/TaskActor.scala
```diff
@@ -124,92 +100,49 @@ class TaskActor(val taskContextData : TaskContextData, userConf : UserConfig, va
...
-      }
+  def minClockAtCurrentTask: TimeStamp = {
+    this.subscriptions.foldLeft(Long.MaxValue){ (clock, subscription) =>
+      Math.min(clock, subscription._2.minClock)
+    }
+  }
+
+  private def allowSendingMoreMessages(): Boolean = {
+    this.subscriptions.foldLeft(true) {
+      (status, subscrption) => status && subscrption._2.allowSendingMoreMessages()
```
manuzhang: could break early here

streaming/src/main/scala/org/apache/gearpump/streaming/task/TaskActor.scala
```diff
@@ -124,92 +100,49 @@ class TaskActor(val taskContextData : TaskContextData, userConf : UserConfig, va
...
-    while (flowControl.allowSendingMoreMessages() && !done) {
+
+    while (allowSendingMoreMessages() && !done) {
       val msg = queue.poll()
       if (msg != null) {
         msg match {
           case SendAck(ack, targetTask) =>
             transport(ack, targetTask)
-            LOG.debug(s"Sending ack back, target taskId: $taskId, my task: $taskId, received message: ${ack.actualReceivedNum}")
+            //LOG.info(s"Sending ack back, target taskId: $taskId, my task: $taskId, received message: ${ack.actualReceivedNum}")
```
manuzhang: don't we need the log here and below ? or mark them debug ?

streaming/src/main/scala/org/apache/gearpump/streaming/task/Subscription.scala
```diff
@@ -0,0 +1,137 @@
...
+   * @param ack
+   */
+  def receiveAck(ack: Ack): Unit = {
+
+    val index = ack.taskId.index
+
+    if (ack.sessionId == sessionId) {
+      if (ack.actualReceivedNum == ack.seq) {
+        if (ack.seq >= candidateMinClockSince(index)) {
+          minClockValue(ack.taskId.index) = candidateMinClock(index)
```
manuzhang: why not use the value `index` in some places ?

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/ClockService.scala
manuzhang: How about the case where a processor has two upstream processors and there is no relation between the two ?

clockfly: We will add a virtual link between these two processors, so that the level can be ordered.

manuzhang: so that means one's min time could be decided by another who it has no relation with ?

clockfly: yes

whjiang: Why this virtual link is needed? Is there any better way to make topo sort stable? E.g. when there are multiple nodes without input, additionally sort them with nodeId. Something like that.

streaming/src/main/scala/org/apache/gearpump/streaming/task/Subscription.scala
```diff
@@ -0,0 +1,137 @@
...
+    pendingMessageCount(partition) += 1
+
+    if (messageCount(partition) % ONE_ACKREQUEST_PER_MESSAGE_COUNT == 0) {
+      val ackRequest = AckRequest(taskId, messageCount(partition), sessionId)
+      transport.transport(ackRequest, targetTask)
+    }
+  }
+
+  def broadcast(msg: AnyRef): Unit = {
+    transport.transport(msg, allTasks: _*)
```
whjiang: why broadcast msg doesn't follow the same rule for sendMessage(), including messageCount and ack?

clockfly: The intention is used to send message without fault-torellance, like latency probe message.

I will rename the function name to probeLatency

def probeLatency(): Unit = {
  transport.transport(LatencyProbe， allTask: _*)
}

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/ClockService.scala
clockfly: The virtual link is a "Metaphor", it doesn't exist in code.

```
Is there any better way to make topo sort stable?
```
We actually don't care that.

streaming/src/main/scala/org/apache/gearpump/streaming/task/TaskActor.scala
```diff
@@ -86,36 +85,13 @@ class TaskActor(val taskContextData : TaskContextData, userConf : UserConfig, va
...
-
-      transport(msg, outputTaskIds(partition))
-      val ackRequest = flowControl.sendMessage(partition)
-      if (null != ackRequest) {
-        transport(ackRequest, outputTaskIds(partition))
-      }
-
-      start = start + 1
-    }
+    this.subscriptions.foreach(_._2.sendMessage(msg))
```
whjiang: It seems our model is: all subscriptions will receive the message. Inside each subscription, one task actor will receive.

Do we want to support use case like: one of all subscription receive the message. Or one task actor can generate multiple output streams?
For example, for IoT data flow in, if the data exceeds some threshold, generate  alert message to one stream (subscription), otherwise generate data to a normal stream (subscription).

clockfly: @whjiang  Can you create a new issue to discuss the use case you described?

As this PR is a code level refactory of old code, without changing the message delivery rules.

whjiang:
+1. As I am not familiar with the flow control and clock tracking system, it is better if other people can help to review these parts.

streaming/src/main/scala/org/apache/gearpump/streaming/task/TaskActor.scala
clockfly: We can do this in future performance optimization when we find this is really an issue.

streaming/src/main/scala/org/apache/gearpump/streaming/task/TaskActor.scala
clockfly: But I think this case can be solved with filters.

```
val stream = streamApp.newStream(input).process[Either(A, B)](processor)
val normalOut = stream.filter(_.left)
val alerts = stream.filter(_.right)
```

whjiang: Yes. This is a good bypass. I created issue #730 before your comment. 

I am not sure this bypass can work for all cases and languages like Java or not. I will post your bypass in #730.

clockfly:
@manuzhang 

Can you take a look at impl of Subscription（replacing flow control and clock tracker)

manuzhang:
+1

PR #711, opened 2015-03-28
fix #709: npe when streaming dag is not initialized

### LINKED ISSUE #709: Failure to load any Task results in failure of the entire DAG
In the kafka -> hbase use case (see #400) failure to load the CpuPersistor or MemoryPersistor causes the entire DAG setup to fail. Many messages related to Task.onStart should not be called more than once. @stanleyxu2005, @whjiang @clockfly  We should provide feedback in the UI where a particular DAG node has a big 'X' over it...

stanleyxu2005:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/711)


clockfly:
+1

PR #718, opened 2015-03-31
fix #708, allow user to pass in a customized storm config

### LINKED ISSUE #708: [storm connector] allow user to pass in a customized configuration
Besides `defaults.yaml`, storm allow users to have a customized `storm.yaml` configuration which overrides the defaults. We should allow user to do so as well.

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/718)


clockfly:
+1

PR #752, opened 2015-04-03
fix #713, provide an option to read from beginning of a topic

### LINKED ISSUE #713: KafkaSource should provide an option to read from beginning of topic
something like setOffset(0)

Right now has setStartTime however if you set this to 0 - and a prior recorded offset has been stored you will get an Offset Range Exception

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/752)


clockfly:
+1

PR #754, opened 2015-04-03
fix #729 remove argument '-master'

### LINKED ISSUE #729: how to start a worker
the way to start a worker is `bin/worker -ip <worker ip address>`, so

1. the usage info is incorrect 

   ```scala
      val options: Array[(String, CLIOption[Any])] =
         Array("ip"->CLIOption[String]("<master ip address>",required = true))
   ```

2. Since worker ip is used for netty tcp host which is already defined in `gear.conf`, why does user need to pass in the worker ip ?

huafengw:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/754)

### DIFF
--- a/conf/gear.conf
+++ b/conf/gear.conf
@@ -1,13 +1,3 @@
-### Put Akka configuration under this section
-base {
-
-  ##############################
-  ### Required to change!!
-  ### You need to set the ip address or hostname of this machine
-  ###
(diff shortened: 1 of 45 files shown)


clockfly:
+1

kkasravi:
@huafengw @clockfly what replaces -master?

PR #769, opened 2015-04-07
fix #768 Serious performance degration of ui server on windows

### LINKED ISSUE #768: Serious performance degration of ui server on windows
I noticed the Ui server take 100% CPU. After further digging, I found this is a problem when upgrade from akka 2.3.6 to akka 2.3.7.

ticket: https://github.com/akka/akka/issues/16295

reference: http://stackoverflow.com/questions/29453848/high-cpu-on-akka-io-pinned-dispatcher-on-windows

We should probably do a quick release to fix this problem.

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/769)

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -239,6 +239,11 @@ base {
     server.parsing.max-content-length = "1000M"
   }
 
+  ## work-around to fix akka io performance issue on windows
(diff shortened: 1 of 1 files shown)


manuzhang:
+1

PR #783, opened 2015-04-08
add wildcard match to return metrics

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/783)

### DIFF
--- a/core/src/main/scala/org/apache/gearpump/cluster/ClusterMessage.scala
+++ b/core/src/main/scala/org/apache/gearpump/cluster/ClusterMessage.scala
@@ -58,7 +58,7 @@ object ClientToMaster {
 
   case class QueryAppMasterConfig(appId: Int)
 
-  case class QueryHistoryMetrics(appId: Int, path: String)
+  case class QueryHistoryMetrics(appId: Int, path: String, readLatest: Boolean = false)
 }
 
 object MasterToClient {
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/apache/gearpump/cluster/ClusterMessage.scala
```diff
@@ -58,7 +58,7 @@ object ClientToMaster {
 
   case class QueryAppMasterConfig(appId: Int)
 
-  case class QueryHistoryMetrics(appId: Int, path: String)
+  case class QueryHistoryMetrics(appId: Int, path: String, readLatest: Boolean = false)
```
stanleyxu2005: If there is a data type Period or duration, I'd suggest change the method to `case class QueryMetrics(appId: Int, path: String, period: Period = null)`

core/src/main/scala/org/apache/gearpump/util/FileServer.scala
```diff
@@ -79,7 +79,7 @@ class FileServer(rootDir: File, host: String, port : Int) extends Actor with Sta
         case Success(data) =>
           sender ! HttpResponse(entity = HttpEntity(data))
         case Failure(ex) =>
-          LOG.error("get file failed, " + ex.getMessage, ex)
+          LOG.error("get file failed, " + ex.getMessage)
```
stanleyxu2005: `LOG.error("Failed to get file: " + ex.getMessage)`

clockfly:
It should work now.

clockfly:
It should work now.

clockfly:
It should work now.

stanleyxu2005:
+1

PR #792, opened 2015-04-09
fix #790, 1) return detail task data in appmaster REST. 2) bind executor...

### LINKED ISSUE #790: [rest] add rest service to return task detail information

clockfly:
... id with executor system id

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/792)

### DIFF
--- a/conf/gear.conf
+++ b/conf/gear.conf
@@ -4,8 +4,6 @@
 #########################################
 gearpump {
 
-  ## Whether we allow remote debug
-  remote-debug-executor-jvm = false
 
   ##############################
   ### Required to change!!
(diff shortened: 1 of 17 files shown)


conf/gear.conf
```diff
@@ -168,3 +168,7 @@ worker {
 ### Put UI services specific akka config here
 ui {
 }
+
+base {
+  akka.scheduler.tick-duration = 10
```
huafengw: Add some explanation here

core/src/main/scala/org/apache/gearpump/cluster/ClusterMessage.scala
```diff
@@ -94,12 +94,14 @@ object AppMasterToMaster {
...
-    def executors: List[String]
+
+    // from executor Id to executor path
+    def executors: Map[Int, String]
     def toJson: String
   }
 
   case class GeneralAppMasterDataDetail(
-      appId: Int, appName: String = null, actorPath: String = null, executors: List[String] = null)
+      appId: Int, appName: String = null, actorPath: String = null, executors: Map[Int, String] = null)
```
huafengw: Personally think we should avoid using null as the default value which may cause NPE

huafengw:
+1

Issue #684, opened 2015-03-25
[Yarn] When launch Master and Worker in container, jvm option -Xmx is not set


huafengw:

clockfly:
The worker can exceed the container memory and be killed if we don't set this property.

TomaszT:
Setting -Xmx alone will prevent JVM heap from expanding above some level, but will not make calculations of JVM real memory footprint easier.  For better performance, we should also set -Xms to the same level as -Xmx (we will save on few FullGCs at the beginning). Unfortunately that is not enough because of JVM lazy initialization mechanism, so we need to add another one : 
 * -XX:+AlwaysPreTouch (that will force JVM to start with full size heap). 

For more accurate calculations we can also force JVM to start with full sized PermSize (< jdk8) and CacheSize with :
 * XX:PermSize=64m -XX:MaxPermSize=64m and 
 * -XX:InitialCodeCacheSize=64m -XX:ReservedCodeCacheSize=64m.

After JVM start we can observe the amount of memory it occupies. It still will be bigger than sum of the above, because of : 
 * +Xss*number of threads
 * some memory for CMS 
 * ...

I did the test for -Xmx1024m on my laptop and 1250m was enough for Yarn. Because this has high priority I could put all that params into configuration : 
 * gearpump.master.command (or jvmopts)
 * gearpump.worker.command (or jvmopts)

like so : 
```
  worker {
    command = "$JAVA_HOME/bin/java -Xmx1024m -Xms1024m -Xss1m -XX:+AlwaysPreTouch -XX:PermSize=64m -XX:MaxPermSize=64m -XX:InitialCodeCacheSize=64m -XX:ReservedCodeCacheSize=64m -XX:+UseConcMarkSweepGC -XX:CMSInitiatingOccupancyFraction=80 -XX:+UseParNewGC -XX:NewRatio=3"
    main = "org.apache.gearpump.cluster.main.Worker"
    containers = "1"
    memory = "1250"
    vcores = "1"
    logname = "worker.log"
  }
```
And later, look for more elegant solution with logic that will try to calculate value for Yarn. What do you guys think?

clockfly:
I like the idea of putting related config in configuration, like the method you mentioned:
```
  worker {
    command = "$JAVA_HOME/bin/java -Xmx1024m ...."
  }
```

As for the default config, we may want to use a simple and conservative XmX memory setting, 
like this:
```
    command = "$JAVA_HOME/bin/java -Xmx1024m"
[...]

PR #808, opened 2015-04-10
Issue #684 - setting -Xmx for master and worker when running on Yarn. 

TomaszT:
Setting -Xmx256m. For now it seems like more than enough. I've also checked that with this setting JVMs won't exceed container limit and won't be killed.

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/808)


clockfly:
+1

Thanks!

PR #811, opened 2015-04-13
fix #801, add config service for master and worker.

### LINKED ISSUE #801: show unified configurations on dashboard
Configs are now read from multiple places (reference.conf under core, reference.conf under application project, gear.conf, application.conf, java properties) with precedence. We need to provide a unified view of config on dashboard such that users could check whether their configuration actually effect. We could have a read-only version now and allow users to modify configurations on dashboard in the future.

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/811)


stanleyxu2005:
+1

PR #813, opened 2015-04-14
fix #741 add a example transport use case

### LINKED ISSUE #741: add an example use case of transport monitoring

huafengw:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/813)


clockfly:
+1

Issue #824, opened 2015-04-16
[DAG] Add default edge implementation


clockfly:
See discussion at #821

So that, we can write the DAG like this:

Graph(A ~>B~>C) instead of Graph(A~partitioner~>B~partitioner~>C)

PR #827, opened 2015-04-16
issue #824, allow to use default partitioner when defining a DAG

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/827)

### DIFF
--- a/core/src/main/scala/org/apache/gearpump/util/Graph.scala
+++ b/core/src/main/scala/org/apache/gearpump/util/Graph.scala
@@ -30,7 +30,7 @@ import scala.language.implicitConversions
  * Application DAG
  */
 
-class Graph[N, E](private[Graph] val graph : DefaultDirectedGraph[N, Edge[E]]) extends Serializable{
+class Graph[N , E](private[Graph] val graph : DefaultDirectedGraph[N, Edge[E]]) extends Serializable{
   import org.apache.gearpump.util.Graph._
 
   def addVertex(vertex : N): Unit = {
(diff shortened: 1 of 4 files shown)


streaming/src/main/scala/org/apache/gearpump/streaming/StreamApplication.scala
```diff
@@ -74,5 +82,6 @@ object StreamApplication {
   }
 
   val DAG = "DAG"
-}
 
+  private val defaultPartitioner = new HashPartitioner()
```
whjiang: Can we make HashParitioner as global default partitioner? But, make every StreamApplication able to override it with its own partitioner as app-wide default partitioner?

core/src/main/scala/org/apache/gearpump/util/Graph.scala
```diff
@@ -256,10 +258,28 @@ object Graph {
...
+    }
+  }
+
+  class NodeList[N, E](nodes: List[N]) extends Path[N, E](nodes.map(Left(_))) {
+    override def ~[Edge](edge: Edge): Path[N, Edge] = {
+      new Path(nodes.map(Left(_)) :+ Right(edge))
+    }
+
+    override def ~>[Node >: N](node: Node): Path[Node, E] = {
+      new NodeList(nodes :+ node)
```
whjiang: What happens if path is: a ~> b ~RRPartitioner ~> c?

What is the partitioner for a~> b? i assume it shall be HashPartitioner. But, the code seems it will be all RRPartitioner.

PR #828, opened 2015-04-16
Fixes #53 [rest] rest interface to submit a dag by JSON representation

### LINKED ISSUE #53: REST interface to submit a dag by JSON representation
from kam
```
kkasravi commented 4 hours ago
Design document is here which discusses how a gearpump grammar can be developed with examples. It is not yet complete but will be the basis for the design when complete.
```

kkasravi:
Still need to add UT and make sure I can submit all examples via REST API however this command will submit a DAG via REST API for complexdag. Also I would like to have SubmitApplicationRequest defined on the client using scala.js. 
```
curl -XPOST -H 'Content-Type:application/json' -d @applicationrequest.json http://localhost:8090/api/v1.0/application -v -s
```
where applicationrequest.json contains
```
{  "appName": "complexdag",
  "appJar": "./output/target/pack/examples/gearpump-examples-assembly-0.3.5-SNAPSHOT.jar",
  "processors": [[0,{"taskClass":"org.apache.gearpump.streaming.examples.complexdag.Source_0","parallelism":1}],[5,{"taskClass":"org.apache.gearpump.streaming.examples.complexdag.Sink_2","parallelism":1}],[10,{"taskClass":"org.apache.gearpump.streaming.examples.complexdag.Node_3","parallelism":1}],[1,{"taskClass":"org.apache.gearpump.streaming.examples.complexdag.Source_1","parallelism":1}],[6,{"taskClass":"org.apache.gearpump.streaming.examples.complexdag.Sink_1","parallelism":1}],[9,{"taskClass":"org.apache.gearpump.streaming.examples.complexdag.Node_4","parallelism":1}],[2,{"taskClass":"org.apache.gearpump.streaming.examples.complexdag.Node_1","parallelism":1}],[7,{"taskClass":"org.apache.gearpump.streaming.examples.complexdag.Node_0","parallelism":1}],[3,{"taskClass":"org.apache.gearpump.streaming.examples.complexdag.Sink_0","parallelism":1}],[11,{"taskClass":"org.apache.gearpump.streaming.examples.complexdag.Sink_3","parallelism":1}],[8,{"taskClass":"org.apache.gearpump.streaming.examples.complexdag.Sink_4","parallelism":1}],[4,{"taskClass":"org.apache.gearpump.streaming.examples.complexdag.Node_2","parallelism":1}]],
  "dag": {"vertices":[0,2,9,5,8,1,4,11,6,10,3,7],"edges":[[0,"org.apache.gearpump.partitioner.HashPartitioner",3],[9,"org.apache.gearpump.partitioner.HashPartitioner",11],[2,"org.apache.gearpump.partitioner.HashPartitioner",11],[0,"org.apache.gearpump.partitioner.HashPartitioner",6],[2,"org.apache.gearpump.partitioner.HashPartitioner",10],[4,"org.apache.gearpump.partitioner.HashPartitioner",10],[7,"org.apache.gearpump.partitioner.HashPartitioner",11],[2,"org.apache.gearpump.partitioner.HashPartitioner",9],[10,"org.apache.gearpump.partitioner.HashPartitioner",11],[0,"org.apache.gearpump.partitioner.HashPartitioner",10],[0,"org.apache.gearpump.partitioner.HashPartitioner",5],[1,"org.apache.gearpump.partitioner.HashPartitioner",8],[0,"org.apache.gearpump.partitioner.HashPartitioner",4],[0,"org.apache.gearpump.partitioner.HashPartitioner",2],[1,"org.apache.gearpump.partitioner.HashPartitioner",7]]}
}
```

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/828)


core/src/main/scala/org/apache/gearpump/cluster/client/ClientContext.scala
```diff
@@ -20,8 +20,8 @@ package org.apache.gearpump.cluster.client
```
clockfly: ClientContext is quite low level, and has significant impact to almost all classes.

Can you try to build things on top of ClientContext, instead of changing it?

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/SubmitApplicationRequest.scala
```diff
@@ -0,0 +1,91 @@
...
+            val (p1, part, p2) = triple
+            dag.addEdge(p1, part, p2)
+          })
+      }
+
+    })
+    SubmitApplicationRequest(appName = appName, appJar = appJar, processors = processors, dag = dag)
+  }
+
+  implicit def convert(httpRequest: HttpRequest): SubmitApplicationRequest = {
```
clockfly: Is it necessary to depends on class HttpRequest? 
How about depending on a json string?

clockfly: Is it necessary to depends on class HttpRequest? 
How about depending on a json string?

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/SubmitApplicationRequest.scala
```diff
@@ -0,0 +1,91 @@
...
+import scala.util.{Failure, Success, Try}
+
+case class SubmitApplicationRequest (
+    appName: String = null,
+    appJar: String = null,
+    processors: Map[ProcessorId, ProcessorDescription] = null,
+    dag: Graph[Int, String] = null)
+
+object SubmitApplicationRequest {
+  private val LOG: Logger = LogUtil.getLogger(getClass)
```
clockfly: This method is too complicated. Can you try to use automatic conversion?

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/SubmitApplicationRequest.scala
kkasravi: implicit def convert is required for spray. We then use upickle to do the conversion rather than spray's default way of doing things by having all services mixin Json4sSupport. There could be a cleaner way, I'll do some more investigation.

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/SubmitApplicationRequest.scala
kkasravi: This is very similar to object StreamingAppMasterDataDetail which does a 'custom' upickle.Writer. In this case we do a custom upickle.Reader. The complication is how we serialize Graph which is custom. I if we just do a custom read, write on Graph then both StreamingAppMasterDataDetail and SubmitApplicationRequest could use the default read, write. I'll open a separate task on this. We also need to have the same scala code appear on the client using scala.js

kkasravi:
Error is 
```
[info] KafkaStreamProducerSpec:
[info] - KafkaStreamProducer should pull messages from kafka *** FAILED ***
[info]   kafka.admin.AdminOperationException: replication factor: 1 larger than available brokers: 0
[info]   at kafka.admin.AdminUtils$.assignReplicasToBrokers(AdminUtils.scala:70)
[info]   at kafka.admin.AdminUtils$.createTopic(AdminUtils.scala:171)
[info]   at kafka.utils.TestUtils$.createTopic(TestUtils.scala:179)
[info]   at org.apache.gearpump.streaming.kafka.util.KafkaServerHarness$class.createTopicUntilLeaderIsElected(KafkaServerHarness.scala:50)
[info]   at org.apache.gearpump.streaming.examples.kafka.KafkaStreamProducerSpec.createTopicUntilLeaderIsElected(KafkaStreamProducerSpec.scala:42)
[info]   at org.apache.gearpump.streaming.examples.kafka.KafkaStreamProducerSpec$$anonfun$1.apply$mcV$sp(KafkaStreamProducerSpec.scala:69)
[info]   at org.apache.gearpump.streaming.examples.kafka.KafkaStreamProducerSpec$$anonfun$1.apply(KafkaStreamProducerSpec.scala:64)
[info]   at org.apache.gearpump.streaming.examples.kafka.KafkaStreamProducerSpec$$anonfun$1.apply(KafkaStreamProducerSpec.scala:64)
```
I'll retry ...

examples/streaming/kafka/src/test/scala/org/apache/gearpump/streaming/examples/kafka/KafkaStreamProducerSpec.scala
```diff
@@ -87,7 +90,7 @@ class KafkaStreamProducerSpec extends PropSpec with Matchers with BeforeAndAfter
       producer.onNext(Message("next"))
 
       var expectedMessage = messages
-      verify(context).output(argMatch[Message](msg => {
+      verify(context, new After(10000L, VerificationModeFactory.times(1))).output(argMatch[Message](msg => {
```
clockfly: @kkasravi 

I know there are ocassially timeout in UT, but making the timeout to """10000L""" will possibly hang the UT?

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/SubmitApplicationRequest.scala
clockfly: OK

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/SubmitApplicationRequest.scala
clockfly: ok

examples/streaming/kafka/src/test/scala/org/apache/gearpump/streaming/examples/kafka/KafkaStreamProducerSpec.scala
kkasravi: Yes, it may have made it worse. I'll lower that value.

kkasravi:
I've reverted ClientContext to not extend ClientSubmitter and overloaded ClientContext.apply

clockfly:
+1

Please add some test case to ensure this actual work. For example, the json representive has been converted to right DAG.

Another followup(maybe in another PR) is allow user to dynamic submit a jar, so  that the class definition in json can be found from this jar.

kkasravi:
Ok, I'll work on this later this evening and tomorrow.

kkasravi:
Rebased with master. Added UT but it is not passing at the moment due to incorrect way of posting the HttpEntity. Will correct in the AM.

kkasravi:
A few problems that should be noted:

- This test requires the build to provide a complexdag jar which means every travis build now runs 
```
sbt clean assembly pack coverage test
```
- I'm not sure if the VERSION of gearpump is available anywhere within the code. If not I'll need to read it from version.sbt. We may want to make this available in reference.conf

streaming/src/main/scala/org/apache/gearpump/streaming/StreamApplication.scala
```diff
@@ -77,11 +77,25 @@ class StreamApplication(override val name : String,  inputUserConfig: UserConfig
 object StreamApplication {
 
   def apply (name : String, dag: Graph[_ <: Processor[Task], _ <: Partitioner], userConfig: UserConfig): StreamApplication = {
-    val graph = dag.mapVertex(Processor.ProcessorToProcessorDescription(_))
+    import Processor._
+    val graph = dag.mapVertex(ProcessorToProcessorDescription)
     new StreamApplication(name, userConfig, graph)
   }
 
+  def apply (name : String, jarName: String, processors: Map[ProcessorId, ProcessorDescription], dag: Graph[Int, String]): StreamApplication = {
```
clockfly: Instead of creating a new apply, can we use the following instead?
```
def apply (name : String, dag: Graph[_ <: Processor[Task], _ <: Partitioner], userConfig: UserConfig) 
```

clockfly:
@kkasravi 

I don't see why there is a necessity to depends on a jar file for the UT. As far as I know, what you want to verify in UT is that the json user submitted matches the DAG representation we expected.

The UT should be easy to write by mocking Master, so that we can check the submitted application definition in mocked master.

services/src/test/scala/org/apache/gearpump/services/SubmitApplicationRequestServiceSpec.scala
```diff
@@ -0,0 +1,101 @@
...
+
+import scala.concurrent.duration._
+
+class SubmitApplicationRequestServiceSpec extends FlatSpec with PropertyChecks with ScalatestRouteTest with SubmitApplicationRequestService with AppMasterService with Matchers {
+  private val LOG: Logger = LogUtil.getLogger(getClass)
+  def actorRefFactory = system
+  def master = TestCluster.master
+
+  val dag = Graph.empty[Int,String]
+  List(0,2,9,5,8,1,4,11,6,10,3,7).foreach(dag.addVertex)
```
clockfly: This graph seems two complicated for a UT, can we use a simple graph?

core/src/main/scala/org/apache/gearpump/util/Graph.scala
```diff
@@ -286,4 +288,88 @@ object Graph {
...
+            val vertices = value.asInstanceOf[Js.Arr]
+            vertices.value.map(vertex => {
+              vertex.value.asInstanceOf[Int]
+            }).toSeq.foreach(vertex => graph.addVertex(vertex))
+          case "edges" =>
+            val edges = value.asInstanceOf[Js.Arr]
+            edges.value.map(aedge => {
+              var path = upickle.readJs[(Int,String,Int)](aedge)
+              var node1 = path._1
+              var edge = Class.forName(path._2).asInstanceOf[Partitioner]
```
clockfly: This will fail in services daemon.

As the Services daemon will not be able to find the customized partitioner in class path.

core/src/main/scala/org/apache/gearpump/cluster/client/ClientContext.scala
```diff
@@ -37,18 +37,18 @@ import scala.concurrent.duration.Duration
...
 //TODO: add interface to query master here
-class ClientContext(config: Config) {
+class ClientContext(config: Config, sys:Option[ActorSystem], mster: Option[ActorRef]) {
   private val LOG: Logger = LogUtil.getLogger(getClass)
   private implicit val timeout = Timeout(5, TimeUnit.SECONDS)
 
   private val masters = config.getStringList(Constants.GEARPUMP_CLUSTER_MASTERS).toList.flatMap(Util.parseHostList)
 
-  implicit val system = ActorSystem(s"client${Util.randInt}" , config)
+  implicit val system = sys.getOrElse(ActorSystem(s"client${Util.randInt}" , config))
```
clockfly: Instead of doing tricky Option here, how about adding a explicit parameter system: ActorSystem in constructor. 

We can create a default Actorsystem in companion apply if not provided.

core/src/main/scala/org/apache/gearpump/util/Graph.scala
```diff
@@ -286,4 +288,88 @@ object Graph {
       this ~ edge ~> node
     }
   }
+
+  implicit val writer1: Writer[Graph[Int, Partitioner]] = upickle.Writer[Graph[Int, Partitioner]] {
```
clockfly: Strictly speaking, Partitioner should not be referenced in this core project.

services/src/test/scala/org/apache/gearpump/services/SubmitApplicationRequestServiceSpec.scala
kkasravi: Ok, yes i can mock the object in this way. For a functional suite we should perhaps submit different DAG's and verify the response DAG's are in agreement. For this UT I will just verify the JSON has been deserialized as expected into something that client.submitApplication would expect.

core/src/main/scala/org/apache/gearpump/util/Graph.scala
kkasravi: this probably won't be used since we are just checking that the inbound DAG matches expected so i will remove it

core/src/main/scala/org/apache/gearpump/cluster/client/ClientContext.scala
kkasravi: ok

core/src/main/scala/org/apache/gearpump/util/Graph.scala
kkasravi: Ok, i will remove these writers since they won't be used with the new UT.

streaming/src/main/scala/org/apache/gearpump/streaming/StreamApplication.scala
kkasravi: I don't think this is possible since we do not have these Processor classes available in the services JVM. We would need to load the appJar via a class loader and then instantiate each class using the same class loader.

kkasravi:
Note - I had to mock master locally due to the following bug when trying to combine TestKit and RouteTest 
https://groups.google.com/forum/#!topic/spray-user/qtLisDRzCF0

I've kept the dag example the same since it is the complexdag example and was easier to leave it this way.

core/src/main/scala/org/apache/gearpump/cluster/client/ClientContext.scala
kkasravi: Note - since creation of the ActorSystem takes a config parameter, it's not possible to create it in the companion object

.travis.yml
```diff
@@ -3,7 +3,7 @@ language:
 - scala
 before_install:
 - sudo pip install codecov
-script: sbt clean coverage test
+script: sbt clean assembly pack coverage test
```
clockfly: Do we still need to change this build script after using mock?

clockfly:
@kkasravi 
```
I've kept the dag example the same since it is the complexdag example and was easier to leave it this way.
```
OK

.travis.yml
kkasravi: oops we need to revert, updating now.

core/src/main/scala/org/apache/gearpump/cluster/client/ClientContext.scala
clockfly: @kkasravi 

You can pass only ActorSystem, and retrieve config by system.settings.config

streaming/src/main/scala/org/apache/gearpump/streaming/StreamApplication.scala
clockfly: In that case, how about 

```
def apply (name : String, dag: Graph[String, String], userConfig: UserConfig): StreamApplication = {
}
```

The first parameter of Graph is the processor, then second is partitioner.

clockfly:
I can fix the rest things in another PR

+1

PR #834, opened 2015-04-17
Fixes #719 add Kafka Source and HBase Sink for dsl

### LINKED ISSUE #719: add Kafka Source and HBase Sink for dsl

kkasravi:
Modeled after CanBuildFrom where an implicit graph is made available to each Stream[T]'s methods.  The implicit graph should be injectable (noted in TODO in StreamApp object)

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/834)

### DIFF
--- a/experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/Stream.scala
+++ b/experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/Stream.scala
@@ -18,34 +18,38 @@
 
 package org.apache.gearpump.streaming.dsl
 
+import org.apache.gearpump.cluster.UserConfig
+import org.apache.gearpump.streaming.dsl.op.OpType.Closure
 import org.apache.gearpump.streaming.dsl.op._
-import org.apache.gearpump.streaming.task.Task
+import org.apache.gearpump.streaming.task.{Task, TaskContext}
 import org.apache.gearpump.util.Graph
 import org.slf4j.LoggerFactory
 
-class Stream[T](dag: Graph[Op, OpEdge], private val thisNode: Op, private val edge: Option[OpEdge] = None) {
+import scala.reflect.ClassTag
+
+class Stream[T:ClassTag](private val graph: Graph[Op,OpEdge], private val thisNode:Op, private val edge: Option[OpEdge] = None) {
 
   /**
    * convert a value[T] to a list of value[R]
-   * @param fun
+   * @param fun function
    * @param description the descripton message for this opeartion
-   * @tparam R
+   * @tparam R return type
    * @return
    */
-  def flatMap[R](fun: T => TraversableOnce[R], description: String = null): Stream[R] = {
+  def flatMap[R: ClassTag](fun: T => TraversableOnce[R], description: String = null): Stream[R] = {
     val flatMapOp = FlatMapOp(fun, Option(description).getOrElse("flatmap"))
-    dag.addVertex(flatMapOp )
-    dag.addEdge(thisNode, edge.getOrElse(Direct), flatMapOp)
-    new Stream(dag, flatMapOp)
+    graph.addVertex(flatMapOp )
+    graph.addEdge(thisNode, edge.getOrElse(Direct), flatMapOp)
+    new Stream[R](graph, flatMapOp)
   }
 
   /**
    * convert value[T] to value[R]
-   * @param fun
-   * @tparam R
+   * @param fun function
+   * @tparam R return type
    * @return
    */
-  def map[R](fun: T => R, description: String = null): Stream[R] = {
+  def map[R: ClassTag](fun: T => R, description: String = null): Stream[R] = {
     this.flatMap ({ data =>
       Option(fun(data))
     }, Option(description).getOrElse("map"))
@@ -70,9 +74,9 @@ class Stream[T](dag: Graph[Op, OpEdge], private val thisNode: Op, private val ed
    */
   def reduce(fun: (T, T) => T, description: String = null): Stream[T] = {
     val reduceOp = ReduceOp(fun, Option(description).getOrElse("reduce"))
-    dag.addVertex(reduceOp)
-    dag.addEdge(thisNode, edge.getOrElse(Direct), reduceOp)
-    new Stream(dag, reduceOp)
+    graph.addVertex(reduceOp)
+    graph.addEdge(thisNode, edge.getOrElse(Direct), reduceOp)
+    new Stream(graph, reduceOp)
   }
 
   /**
@@ -89,10 +93,10 @@ class Stream[T](dag: Graph[Op, OpEdge], private val thisNode: Op, private val ed
    */
   def merge(other: Stream[T], description: String = null): Stream[T] = {
     val mergeOp = MergeOp(thisNode, other.thisNode, Option(description).getOrElse("merge"))
-    dag.addVertex(mergeOp)
-    dag.addEdge(thisNode, edge.getOrElse(Direct), mergeOp)
-    dag.addEdge(other.thisNode, other.edge.getOrElse(Shuffle), mergeOp)
-    new Stream(dag, mergeOp)
+    graph.addVertex(mergeOp)
+    graph.addEdge(thisNode, edge.getOrElse(Direct), mergeOp)
+    graph.addEdge(other.thisNode, other.edge.getOrElse(Shuffle), mergeOp)
+    new Stream[T](graph, mergeOp)
   }
 
   /**
@@ -114,11 +118,10 @@ class Stream[T](dag: Graph[Op, OpEdge], private val thisNode: Op, private val ed
    */
   def groupBy[Group](fun: T => Group, parallism: Int = 1, description: String = null): Stream[T] = {
     val groupOp = GroupByOp(fun, parallism, Option(description).getOrElse("groupBy"))
-    dag.addVertex(groupOp)
-    dag.addEdge(thisNode, edge.getOrElse(Shuffle), groupOp)
-    new Stream(dag, groupOp)
+    graph.addVertex(groupOp)
+    graph.addEdge(thisNode, edge.getOrElse(Shuffle), groupOp)
+    new Stream[T](graph, groupOp)
   }
-
 
   /**
    * connect with a low level Processor(TaskDescription)
@@ -127,24 +130,22 @@ class Stream[T](dag: Graph[Op, OpEdge], private val thisNode: Op, private val ed
    * @tparam R
    * @return
    */
-  def process[R](processor: Class[_ <: Task], parallism: Int, description: String = null): Stream[R] = {
+  def process[R: ClassTag](processor: Class[_ <: Task], parallism: Int, description: String = null): Stream[R] = {
     val processorOp = ProcessorOp(processor, parallism, Option(description).getOrElse("process"))
-    dag.addVertex(processorOp)
-    dag.addEdge(thisNode, edge.getOrElse(Shuffle), processorOp)
-    new Stream(dag, processorOp, Some(Shuffle))
+    graph.addVertex(processorOp)
+    graph.addEdge(thisNode, edge.getOrElse(Shuffle), processorOp)
+    new Stream[R](graph, processorOp, Some(Shuffle))
   }
+
 }
 
 class KVStream[K, V](stream: Stream[Tuple2[K, V]]){
-
   /**
    * Apply to Stream[Tuple2[K,V]]
    * Group by the key of a KV tuple
    * For (key, value) will groupby key
-   * @tparam T
    * @return
    */
-  import Stream._
   def groupByKey(parallism: Int = 1): Stream[Tuple2[K, V]] = {
     stream.groupBy(Stream.getTupleKey[K, V], parallism, "groupByKey")
   }
@@ -165,11 +166,27 @@ class KVStream[K, V](stream: Stream[Tuple2[K, V]]){
 }
 
 object Stream {
+
+  def apply[T: ClassTag](graph: Graph[Op, OpEdge], node: Op, edge: Option[OpEdge]) = new Stream[T](graph, node, edge)
+
   def getTupleKey[K, V](tuple: Tuple2[K, V]): K = tuple._1
 
   def sumByValue[K, V](numeric: Numeric[V]): (Tuple2[K, V], Tuple2[K, V]) => Tuple2[K, V]
   = (tuple1, tuple2) => Tuple2(tuple1._1, numeric.plus(tuple1._2, tuple2._2))
 
   implicit def streamToKVStream[K, V](stream: Stream[Tuple2[K, V]]): KVStream[K, V] = new KVStream(stream)
+
+  implicit class Sink[T: ClassTag](stream: Stream[T]) extends java.io.Serializable {
+
+    def sink(closure: Closure[T], parallelism: Int = 1, description: String = null): Stream[T] = {
+      val sinkOp = SinkOp(closure, parallelism, Option(description).getOrElse("sink"))
+      stream.graph.addVertex(sinkOp)
+      stream.graph.addEdge(stream.thisNode, Shuffle, sinkOp)
+      new Stream[T](stream.graph, sinkOp)
+    }
+    def writeToHBase(closure: Closure[T], parallelism: Int = 1, description: String = null): Stream[T] = {
+      sink(closure, parallelism, description)
+    }
+  }
 
 }
--- a/experiments/dsl/src/test/scala/org/apache/gearpump/streaming/dsl/StreamAppSpec.scala
+++ b/experiments/dsl/src/test/scala/org/apache/gearpump/streaming/dsl/StreamAppSpec.scala
@@ -21,6 +21,7 @@ package org.apache.gearpump.streaming.dsl
 import akka.actor.ActorSystem
 import org.apache.gearpump.cluster.TestUtil
 import org.apache.gearpump.cluster.client.ClientContext
+import org.apache.gearpump.streaming.dsl.op.{OpEdge, Op}
 import org.apache.gearpump.streaming.dsl.plan.OpTranslator._
 import org.mockito.Mockito.when
 import org.scalatest._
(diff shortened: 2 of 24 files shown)


experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/Stream.scala
```diff
@@ -127,25 +133,38 @@ class Stream[T](dag: Graph[Op, OpEdge], private val thisNode: Op, private val ed
...
+    graph.addEdge(thisNode, edge.getOrElse(Shuffle), processorOp)
+    new Stream[R](processorOp, Some(Shuffle))
+  }
+
+  /**
+   * connect with a low level Processor(TaskDescription)
+   * @param processors
+   * @return
+   */
+  def branch(processors: ProcessorOp[_ <: Task]*)(implicit graph: Graph[Op,OpEdge]): Stream[T] = {
```
clockfly: There is already process, why adding this?

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/Stream.scala
```diff
@@ -23,29 +23,35 @@ import org.apache.gearpump.streaming.task.Task
 import org.apache.gearpump.util.Graph
 import org.slf4j.LoggerFactory
 
-class Stream[T](dag: Graph[Op, OpEdge], private val thisNode: Op, private val edge: Option[OpEdge] = None) {
+import scala.reflect.ClassTag
+
+class Stream[T:ClassTag](private val thisNode:Op, private val edge: Option[OpEdge] = None) {
+
+  def and: Stream[T] = {
```
clockfly: ? what is this?

clockfly:
-1

1. branch and and doesn't make sense.
2. It may not be a good idea to add an implicit graph everywhere

What I expected was something like:
streamApp.from(kafkaSource).flatmap.reduce.writeTo(hbaseSink)

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/Stream.scala
kkasravi: See PipeLineDSL 
```

    app.from(ProcessorOp(classOf[KafkaProducer], 1, "KafkaProducer")).
      branch(ProcessorOp(classOf[CpuProcessor], processors, "CpuProcessor"), ProcessorOp(classOf[CpuPersistor], persistors, "CpuPersistor")).and.
      branch(ProcessorOp(classOf[MemoryProcessor], processors, "MemoryProcessor"), ProcessorOp(classOf[MemoryPersistor], persistors, "MemoryPersistor"))
```

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/Stream.scala
kkasravi: Stream.process does not allow branching - only linear downstream chaining. This allows multiple downstream nodes to be attached at a single upstream node.

kkasravi:
The use case
```
streamApp.from(kafkaSource).flatmap.reduce.writeTo(hbaseSink)
```
doesn't map to what's done in PipeLine currently where the source node is branched into 2 downstream paths - one for 'cpu' the other for 'memory'. This results in 2 sinks (or writeTo's).

kkasravi:
By the way I made an attempt to change Graph[Op,OpEdge] to Graph[V[_]<:Op,E<:OpEdge] that would allow Graph to be a typeclass so that different types of Graphs could be injected into Stream[T] that have defined new types of Op, OpEdges. This brings the design much closer to scala collections (or even scala-graph). This ends up being a pretty involved change since we pull in higher kinded types.

kkasravi:
Investigating the build failure.

kkasravi:
We could do something like this
```
streamApp.from(kafkaSource).flatmap(cpuprocessor).reduce(cpupersistor).to(hbaseSink).and.flatmap(cpuprocessor).reduce(cpupersistor).to(hbaseSink)

**Comment by kkasravi** [comment]
Ok, i'm removing ProcessorOp and lifting the various onNext operations out into the scope of PipeLineDSL. Right now it will only deliver batchSize messages. If this direction is a '+1', I'll continue to correctly define a source, sink Op and remove 'and', 'batch' etc.

**Comment by clockfly** [comment]
For the branch,

val root = streamApp.from(kafkaSource)
root.flatmap(cpuprocessor).reduce(cpupersistor).to(hbaseSink)
root.flatmap(cpuprocessor).reduce(cpupersistor).to(hbaseSink)

No need to use "and" or "branch"

File: experiments/dsl/src/test/scala/org/apache/gearpump/streaming/dsl/StreamAppSpec.scala [hunk 3]
```diff
@@ -39,19 +40,19 @@ class StreamAppSpec  extends FlatSpec with Matchers with BeforeAndAfterAll {
 
 
   it should "be able to generate multiple new streams" in {
-    val app = new StreamApp("dsl", context)
-    app.fromCollection(List("A").iterator, 1)
-    app.fromCollection(List("B").iterator, 1)
+    val app = StreamApp("dsl", context)
+    app(List("A"), 1)
```
clockfly: It is not good to construct a app by a source, as one app can have multiple source.

For example:

val app = StreamApp("appName")
app.fromSource(Kafka).map.reduce.writeTo(sink)
app.from(anotherSource).writeTo(sink2)

kkasravi:
Ok, right now i have this:
```
  def filter(dimension: String)(msg: Message): Traversable[Message] = {
    val jsonData = msg.msg.asInstanceOf[String]
    val metrics = read[Array[Datum]](jsonData)
    val data = metrics.flatMap(datum => {
      datum.dimension match {
        case dimension =>
          Some(datum)
        case _ =>
          None
      }
    }).toSeq.toArray
    Array(Message(write[Array[Datum]](data), msg.timestamp)).toTraversable
  }

  def application(context: ClientContext, config: ParseResult): Unit = {
    import Messages._
    import PipelineConfig._
    val pipeLinePath = config.getString("conf")
    val pipelineConfig = PipelineConfig(ConfigFactory.parseFile(new java.io.File(pipeLinePath)))
    val processors = pipelineConfig.getInt(PROCESSORS)
    val persistors = pipelineConfig.getInt(PERSISTORS)
    val kafkaConfig = KafkaConfig(pipelineConfig)
    val repo = new HBaseRepo {
      def getHBase(table: String, conf: Configuration): HBaseSinkInterface = HBaseSink(table, conf)
    }
    val appConfig = UserConfig.empty.withValue(KafkaConfig.NAME, kafkaConfig).withValue(PIPELINE, pipelineConfig).withValue(HBASESINK, repo)
    val source: TimeReplayableSource = new KafkaSource("KafkaSource", TaskId(0,0), 1, kafkaConfig, kafkaConfig.getMessageDecoder)
    val app = StreamApp("PipeLine", context, appConfig)
    val producer = app(source, 1, kafkaConfig.getConsumerEmitBatchSize, "kafka-producer")
    producer.flatMap(filter(CPU)).process(classOf[CpuPersistor], persistors, "cpu")
    producer.flatMap(filter(MEM)).process(classOf[MemoryPersistor], persistors, "memory")
```
Using reduce function (ReduceOp) right now doesn't do what i would like it to do, i need to create something like reduceWhile (similar to list.takeWhile) where taskContext.output is only called when (T) => Boolean returns true which in this case would be a timestamp delta.

experiments/dsl/src/test/scala/org/apache/gearpump/streaming/dsl/StreamAppSpec.scala
kkasravi: ok a new empty graph is constructed for each app instance. The example above should now result in 2 sources in the DAG. Below is a similar example:
```
val app = StreamApp("App", context, appConfig)
implicit val graph: Graph[Op,OpEdge] = app.graph
val producer1 = app(source, 1, kafkaConfig.getConsumerEmitBatchSize, "kafka-producer")
val data = "This is a good start, bingo!! bingo!!"
val producer2 = app(data.lines.toList, 1)
```
Still working on the sink/writeTo

clockfly:
Why cannot you just use the dsl filter()?

clockfly:
@kkasravi 
I think if we should use strong typed message, instead of doing manual serialization/deserializaton.

kkasravi:
Yes I think i can use filter + reduce. I should have this change checked in before your Monday.

This is a good point for Graph. I'll create a case class that holds vertices and edges.

clockfly:
Can we avoiding using implicit graph?

For example, if user want to define two StreamApp in same program, then they will conflict with each other.

kkasravi:
I'll revert to where the Graph is defined within Stream during Stream creation. Having graph as an implicit parameter for each Stream method (similar to scala collections CanBuildFrom) doesn't quite make sense since the graph is building a DAG across Stream methods rather than helping to build the next Stream[T].

clockfly:
@kkasravi 

It may be challenging to support any ReplayableSource in a generic way.
How about adding a OP specifially for kafka.

Something like:
```
Stream.fromKafka[T](kafkaConfig， other config) {
}
```

When we get more confidence about supporting generic Replayable source, then we can revisit this.

kkasravi:
This latest update exposes some possible problems within the DSL - where serialization errors occur if any of the method references other methods or vals outside that method's scope. Probably a given - but suggests we handle the errors better and possibly look at spores.

kkasravi:
I will probably need to merge master into this branch. Waiting on the build. PIpeLineDSLSpec is something to look at that is emulating kafka, hbase.

whjiang:
Maybe we can follow Spark streaming API for Kafka:
```
import kafka.serializer.StringDecoder
import org.apache.spark.SparkConf
import org.apache.spark.streaming.{Seconds, StreamingContext}
import org.apache.spark.streaming.kafka.KafkaUtils
 
val ssc = new StreamingContext(new SparkConf, Seconds(60))
 
// hostname:port for Kafka brokers, not Zookeeper
val kafkaParams = Map("metadata.broker.list" -> "localhost:9092,anotherhost:9092")
 
val topics = Set("sometopic", "anothertopic")
 
val stream = KafkaUtils.createDirectStream[String, String, StringDecoder, StringDecoder](
  ssc, kafkaParams, topics)
```

clockfly:
@kkasravi 

Maybe you should throw way custom serialization and try strong typed stream this time, it will work this time:)

kkasravi:
Yes, it would make the code smaller and much more elegant. I will do so. Thanks.

kkasravi:
I've converted PipeLineDSL and PipeLineDSLSpec to not do any custom serialization.
I'm getting a kryo serialization exception. I think I'm not entering the Array[Datum] class in the right conf file. I've added to conf/pipeline.conf

```
gearpump {
  serializers {
    "org.apache.gearpump.experiments.pipeline.Messages$Datum" = ""
    "org.apache.gearpump.experiments.pipeline.Messages.Datum" = ""
  }
}
```


```
[ERROR] [04/22/2015 13:38:00.562] [OneForOneStrategy] Encountered unregistered class ID: 257443811
com.esotericsoftware.kryo.KryoException: Encountered unregistered class ID: 257443811
        at com.esotericsoftware.kryo.util.DefaultClassResolver.readClass(DefaultClassResolver.java:119)
        at com.esotericsoftware.kryo.Kryo.readClass(Kryo.java:641)
        at com.esotericsoftware.kryo.Kryo.readClassAndObject(Kryo.java:752)
        at org.apache.gearpump.streaming.MessageSerializer.read(MessageSerializer.scala:34)
        at org.apache.gearpump.streaming.MessageSerializer.read(MessageSerializer.scala:26)
        at com.esotericsoftware.kryo.Kryo.readClassAndObject(Kryo.java:764)
        at org.apache.gearpump.serializer.FastKryoSerializer.deserialize(FastKryoSerializer.scala:66)
        at org.apache.gearpump.transport.netty.Server$$anonfun$msgHandler$1$$anonfun$applyOrElse$3$$anonfun$apply$1.apply(Server.scala:70)
        at org.apache.gearpump.transport.netty.Server$$anonfun$msgHandler$1$$anonfun$applyOrElse$3$$anonfun$apply$1.apply(Server.scala:67)
        at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
        at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
        at org.apache.gearpump.transport.netty.Server$$anonfun$msgHandler$1$$anonfun$applyOrElse$3.apply(Server.scala:67)
        at org.apache.gearpump.transport.netty.Server$$anonfun$msgHandler$1$$anonfun$applyOrElse$3.apply(Server.scala:61)
        at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
        at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
        at scala.collection.immutable.Map$Map1.foreach(Map.scala:116)
        at scala.collection.TraversableLike$class.map(TraversableLike.scala:245)
        at scala.collection.AbstractTraversable.map(Traversable.scala:104)
        at org.apache.gearpump.transport.netty.Server$$anonfun$msgHandler$1.applyOrElse(Server.scala:61)
        at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:170)
        at akka.actor.Actor$class.aroundReceive(Actor.scala:465)
        at org.apache.gearpump.transport.netty.Server.aroundReceive(Server.scala:34)
        at akka.actor.ActorCell.receiveMessage(ActorCell.scala:516)
        at akka.actor.ActorCell.invoke(ActorCell.scala:487)
        at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:254)
        at akka.dispatch.Mailbox.run(Mailbox.scala:221)
        at akka.dispatch.Mailbox.exec(Mailbox.scala:231)
        at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
        at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
        at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
        at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
```

clockfly:
@kkasravi 

Please search the LOG for 257443811, there should be some log saying implicit register ... class 
Then, you need to add that class into the serializer list.

kkasravi:
Ok, thanks. The PipeLineDSL example is looking pretty compelling. I think there is still some amount of refinement for a standard Source and Sink but it is getting there.

clockfly:
@kkasravi 

For the kryo exception, I can reproduce it. There is no information about which class cannot be serialized. 

I have created a issue https://github.com/intel-hadoop/gearpump/issues/871 
And will submit a PR soon.

clockfly:
@kkasravi 

https://github.com/intel-hadoop/gearpump/pull/872 should solve the problem.
When there is error log, you then need to add the class name into gear.conf serializer list.

kkasravi:
Thanks @clockfly. I'm now testing with Kafka and HBase on a hadoop cluster. I've verified that it works for stubbed/mock interfaces. I'll update when the kafka/hbase pathway is working. If you're anxious to merge this to master, i can get this working on new issue.

clockfly:
@kkasravi 

Generally, I think the patch is good.

Is it possible to add a shortcut for kafka and hbase. For example, add a operator in Stream called fromKafka, and add a sink in Stream called writeToHbase()?

Then the PipleLineDSL code will be much simplified. 

Currently, I see in HbaseSink, there is also many map-like transformation job, is it possible to break that to a map operator and a pure-simple hbase sink?

kkasravi:
@clockfly Yes, I'll add these operators. See #53 for recent updates made today and some related questions.

kkasravi:
Hi @clockfly the PipeLineDSL and PipeLineDSLSpec are much shorter 
```
    val app = StreamApp("PipeLineDSL", context, appConfig)
    val producer = app.readFromKafka(new ReplayableSource(kafkaConfig), 1, "time-replayable-producer")

    producer.flatMap(metrics => {
      Some(metrics.flatMap(datum => {
        datum.dimension match {
          case CPU =>
            Some(datum)
          case _ =>
            None
        }
      }))
    }).reduce((msg1, msg2) => {
      val now = System.currentTimeMillis
      def average(average: TAverage)(metrics: Array[Datum]): Array[Datum] = {
        metrics.flatMap(datum => {
          average.average(datum, now)
        })
      }
      average(TAverage(50))(msg2)
    }).writeToHBase(toHBaseSink)

    producer.flatMap(metrics => {
      Some(metrics.flatMap(datum => {
        datum.dimension match {
          case MEM =>
            Some(datum)
          case _ =>
            None
        }
      }))
    }).reduce((msg1, msg2) => {
      val now = System.currentTimeMillis
      def average(average: TAverage)(metrics: Array[Datum]): Array[Datum] = {
        metrics.flatMap(datum => {
          average.average(datum, now)
        })
      }
      average(TAverage(100))(msg2)
    }).writeToHBase(toHBaseSink)
```

experiments/kafka-hbase-pipeline/src/main/scala/org/apache/gearpump/experiments/pipeline/PipeLineDSL.scala
```diff
@@ -0,0 +1,157 @@
...
+    val persistors = pipeLineConfig.getInt(PERSISTORS)
+    val kafkaConfig = KafkaConfig(pipeLineConfig)
+    val repo = new HBaseRepo {
+      def getHBase(table: String, conf: Configuration): HBaseSinkInterface = HBaseSink(table, conf)
+    }
+    val appConfig = UserConfig.empty.withValue(KafkaConfig.NAME, kafkaConfig).withValue(PIPELINE, pipeLineConfig).withValue(HBASESINK, repo)
+    System.setProperty(Constants.GEARPUMP_CUSTOM_CONFIG_FILE, pipeLinePath)
+
+    val app = StreamApp("PipeLineDSL", context, appConfig)
+    val producer = app.readFromKafka(new ReplayableSource(kafkaConfig), 1, "time-replayable-producer")
```
clockfly: Can this be simplified by only relying on a kakfaConfig?

For example:
app.readfromKafka(kafkaConfig, name)

kkasravi:
app.fromKafka now takes a kafkaConfig and a message converter to deserialize the message to the target type. 
```
    val producer = app.readFromKafka(kafkaConfig, msg => {
      val jsonData = msg.msg.asInstanceOf[String]
      val envelope = read[Envelope](jsonData)
      val body = read[Body](envelope.body)
      body.metrics
    }, 1, "time-replayable-producer")
```
I think the converter (Message => T) is unavoidable since the user may or may not have control over the kafka topic schema

clockfly:
+1

clockfly:
committed.

PR #850, opened 2015-04-21
fix #849, set default hostname to 127.0.0.1 in UT

### LINKED ISSUE #849: [UT] set default hostname to 127.0.0.1 in UT

clockfly:


huafengw:
+1

PR #857, opened 2015-04-22
fix #856 fix service launch failed

### LINKED ISSUE #856: Service launch failed

huafengw:


clockfly:
@huafengw 

There is also a same bug in Local

clockfly:
+1

PR #858, opened 2015-04-22
issue #853, fix resource leak(thread not closed, process not killed in t...

clockfly:
...ime) in UT. Also increase the PermGen size to avoid Permgen OOM.

### DIFF
--- a/services/src/test/scala/org/apache/gearpump/services/TestCluster.scala
+++ b/services/src/test/scala/org/apache/gearpump/services/TestCluster.scala
@@ -18,28 +18,19 @@
 
 package org.apache.gearpump.services
 
-import org.apache.gearpump.cluster.AppMasterToMaster.MasterData
+import akka.actor.ActorRef
 import org.apache.gearpump.cluster.TestUtil
 import org.apache.gearpump.cluster.TestUtil.MiniCluster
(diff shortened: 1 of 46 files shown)


services/src/test/scala/org/apache/gearpump/services/TestCluster.scala
```diff
@@ -0,0 +1,36 @@
...
+object TestCluster {
+  private val LOG = LogUtil.getLogger(getClass)
+
+  lazy val cluster: MiniCluster = TestUtil.startMiniCluster
+  lazy val master: ActorRef = {
+    LOG.info("Accessing master of TestCluster...")
+    StreamingTestUtil.startAppMaster(cluster, 0)
+    cluster.mockMaster
+  }
+}
```
huafengw: Seems the MiniCluster is not shut down when it's used.

clockfly: Yes, it is a singleton.

I am not sure whether we should close the singleton.

clockfly: There may exist a bug in akka cluster, 
https://github.com/akka/akka/issues/17253

It will cause the akka cluster actor system failed to start. The singleton is a work-around to avoid frequent start/stop of akka cluster.

huafengw:
+1

Issue #865, opened 2015-04-23
Change the default timeout setting in akka test expectMsg


clockfly:
akka.test.single-expect-default

Check http://doc.akka.io/docs/akka/snapshot/general/configuration.html

PR #866, opened 2015-04-23
issue #865, Change the default timeout setting in akka test expectMsg

clockfly:


core/src/main/resources/reference.conf
```diff
@@ -269,7 +269,12 @@ base {
   akka.scheduler.tick-duration = 1
 
   akka {
-    
+
+    test {
+      ##  duration to wait in expectMsg in akka test
+      single-expect-default = 10s
+    }
+
```
manuzhang: why is this not put in `test.conf` ?

clockfly: @manuzhang 

We need to make this config global visible. 
When user create a default ActorSystem, then test.conf will not be loaded.

manuzhang:
+1

PR #878, opened 2015-04-26
fix #877: source node and sink node were not calculated correctly.

### LINKED ISSUE #877: Incorrect Sink Processor Receive Throughput values on chart
The values on sink processor receive throughput of SOL application is incorrect.

stanleyxu2005:


clockfly:
+1

PR #880, opened 2015-04-27
fix #879: split metrics into different views and changed tooltip control

### LINKED ISSUE #879: Split metrics in different tabs
The benefit is to improve the performance

stanleyxu2005:


clockfly:
+1

PR #882, opened 2015-04-28
fix #881: diverse issues of skew charts and made tooltip nice

### LINKED ISSUE #881: Skew chart user experienece improvment
1. When there are too many data on skew chart, better have a way to zoom data. 
2. Always show skew chart, even there is only one task
3. Fix the unit typo in charts. 
4. More human readable values in chart's tooltip. 
5. Area of tasks' metrics should be stacked in chart

stanleyxu2005:


clockfly:
+1

PR #884, opened 2015-04-28
fix #876: clock is updated every second

### LINKED ISSUE #876: Replay application clock more smoothly
Dashboard will query application clock every 2 seconds (by default). So the clock is not updated smoothly from user experience perspective.

stanleyxu2005:


clockfly:
+1

PR #886, opened 2015-04-28
fix #885: wrong application clock in some case

### LINKED ISSUE #885: Wrong application clock
The application clock would not update expectedly. The problem is lead by the commit (6534f95664fafd6601fc9bb8c13219542f686338).

stanleyxu2005:


clockfly:
+1

PR #889, opened 2015-04-29
fix #801: added download links for configurations

### LINKED ISSUE #801: show unified configurations on dashboard

stanleyxu2005:


clockfly:
+1

PR #909, opened 2015-05-08
fix#679 add a InterMediateService to forward rest request

huafengw:
Seems spray do not support forward a rest request


huafengw:
Now the rest api looks like 
``http://localhost:8090/api/v1.0/internal/intermediate/delete/api/v1.0/appmaster/0```
and the intermediate rest only support DELETE and PUT

clockfly:
1. Maybe we should not duplicate the version twice. 
2. This URL should only be available for POST, as GET can be cached by intermediate proxy.

huafengw:
Updated. Now the rest api looks like
``http://localhost:8090/api/v1.0/internal/intermediate/delete/appmaster/0``
and this url only support POST

clockfly:
+1

Issue #710, opened 2015-03-28
Refactor YarnAM to better handle state including container failure states


TomaszT:
Currently in case of container failure, YarnAM will refuse to start new one. Also YarnAM could be refactored and use FSM, so the code would be cleaner and easier to read.

clockfly:
Solved by PR https://github.com/intel-hadoop/gearpump/pull/1024

PR #911, opened 2015-05-11
FIx #661, specs and refactoring. 

### LINKED ISSUE #661: Add unittests for Yarn

TomaszT:
Hi,
these are the specs for all but 3 classes from org.apache.gearpump.experiments.yarn package. Classes that are currently missing specs are : 
  * org.apache.gearpump.experiments.yarn.client.Client -> Kam is working on that one
  * org.apache.gearpump.experiments.yarn.master.ServiceLauncherActor -> this will be added with #710, because it will be heavily refactored
  * org.apache.gearpump.experiments.yarn.master.ResourceManagerCallbackHandler -> this one is work in progress and also will be influenced by #710.

Refactoring part is related to adding some factories. There was a few places in AmActor where objects were created and there was no way to do DI.


experiments/yarn/src/test/resources/gear.conf
```diff
@@ -0,0 +1,165 @@
+#########################################
```
clockfly: @TomaszT 

"gear.conf" should not be included here.

If you want to create some default configuration file, you can use file name "reference.conf",
and only contains settings which is interesting to gearpump-yarn.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
```diff
@@ -189,8 +188,7 @@ class Client(configuration:AppConfig, yarnConf: YarnConfiguration, yarnClient: Y
     val appContext = yarnClient.createApplication.getApplicationSubmissionContext
     appContext.setApplicationName(getEnv(YARNAPPMASTER_NAME))
 
-    val containerContext: ContainerLaunchContext = YarnContainerUtil.getContainerContext(yarnConf, getCommand)
-    containerContext.setLocalResources(YarnContainerUtil.getAMLocalResourcesMap(yarnConf, getConfiguration))
+    val containerContext = DefaultContainerLaunchContextFactory(yarnConf, getConfiguration).newInstance(getCommand)
     appContext.setAMContainerSpec(containerContext)
     appContext.setResource(getAMCapability)
     appContext.setQueue(getEnv(YARNAPPMASTER_QUEUE))
```
clockfly:
```
  def deploy() = {
    LOG.info("Starting AM")
    //uploadAMResourcesToHDFS()
    uploadConfigToHDFS()
    yarnClient.init(yarnConf)
    yarnClient.start()
    val appContext = yarnClient.createApplication.getApplicationSubmissionContext
    appContext.setApplicationName(getEnv(YARNAPPMASTER_NAME))

    val containerContext = DefaultContainerLaunchContextFactory(yarnConf, getConfiguration).newInstance(getCommand)
    appContext.setAMContainerSpec(containerContext)
    appContext.setResource(getAMCapability)
    appContext.setQueue(getEnv(YARNAPPMASTER_QUEUE))
    
    yarnClient.submitApplication(appContext)
    monitorAM(appContext)
  }

  deploy()
```

It maybe better to call deploy() when using Client, insteading of constructing Client.

clockfly:
@TomaszT 

I find it unclear in several places:

1). There are about 20 event definition in Actions.scala, but I cannot understand what there are about, as these event are part of the AmActor interface, can you document all of them?
```scala
object Actions {
  sealed trait Reason
  case class Failed(throwable: Throwable) extends Reason
  case object ShutdownRequest extends Reason
  case object AllRequestedContainersCompleted extends Reason
  
  sealed trait ContainerType
  case object Master extends ContainerType
  case object Woker extends ContainerType
  case object Service extends ContainerType

  sealed trait YarnApplicationMasterState
  case object RequestingMasters extends YarnApplicationMasterState
  case object RequestingWorkers extends YarnApplicationMasterState
  case object RequestingService extends YarnApplicationMasterState
  
  sealed trait YarnApplicationMasterData
  case object Uninitialized extends YarnApplicationMasterData
  
  case class LaunchContainers(containers: List[Container]) extends YarnApplicationMasterData 
  case class LaunchWorkerContainers(containers: List[Container])
  case class LaunchServiceContainer(containers: List[Container])
  case class ContainerRequestMessage(memory: Int, vCores: Int)
  case class RMHandlerDone(reason: Reason, rMHandlerContainerStats: RMHandlerContainerStats)
  case class RMHandlerContainerStats(allocated: Int, completed: Int, failed: Int)
  case class RegisterAMMessage(appHostName: String, appHostPort: Int, appTrackingUrl: String)
  case class AMStatusMessage(appStatus: FinalApplicationStatus, appMessage: String, appTrackingUrl: String)
  case class ContainerInfo(container:Container, containerType: ContainerType, launchCommand: String => String)
  case class ContainerStarted(containerId: ContainerId)
}
```

2). About RmHandlerDone, I see  RMHandlerDone have wrapped another case class "Reason", how about making RMHandlerDone explit. 
For example, it may looks better to change from
```scala
  sealed trait Reason
  case class Failed(throwable: Throwable) extends Reason
  case object ShutdownRequest extends Reason
  case object AllRequestedContainersCompleted extends Reason
  
 case class RMHandlerDone(reason: Reason, rMHandlerContainerStats: RMHandlerContainerStats)
```

to something like
```scala
sealed trait ResourceManagerResponse
case class RMAllContainerAllocated extends ResourceManagerResponse
case class RMAllocationFailed(ex) extends ResourceManagerResponse
```

And ShutdownRequest  is confusing to me, I don't know what it is.


3). In AmActor receive handler, there is a case branch 
```scala
    case amResponse: RegisterApplicationMasterResponse =>
      LOG.info("Received RegisterApplicationMasterResponse")
      requestMasterContainers(amResponse)
```
Note here that RegisterApplicationMasterResponse  is a class defined in YARN project. This is not a good practice to rely on a YARN class "RegisterApplicationMasterResponse" in case match, as 
   a) akka require all message to be immutable, and we have no control over RegisterApplicationMasterResponse   and don't know whether it is immutable or not.
   b) It create a dependency on YARN class, and the abstraction level is not consistent with other case match branches.

I will suggest to create a new class to represent that message which is defined in scope of Gearpump.

4). I saw there are a lot of public val, var definition in various class, can you change the visibility that we only expose the inner class implementation when necessary.

5). In good practice, case match messages in receive handler should be defined in companion object of the Actor.  For example, RegisterAMMessage is only used by Actor ResourceManagerClientActor, so I think message RegisterAMMessage  should be defined in object ResourceManagerClientActor

The story is same for Actor AmActor

6). I don't find "ResourceManagerCallbackHandlerActor" is used anywhere? Maybe we should remove it if it is not used?

7). I see there are ServicesLauncherActor, mind renaming it to UILauncher? And for other Actor class with name suffixed by "Actor", the "Actor" suffix may not be necessary.


Thanks 

Sean

kkasravi:
@clockfly @TomaszT For RegisterApplicationMasterResponse we can wrap this in something like case class RegisterAppMasterResponse(response: RegisterApplicationMasterResponse) but we'll still need to send this over the wire.

TomaszT:
@clockfly just pushed in reply to you previous comment. Configuration files are not used anymore, problem with deploy() was resolved by Kam (he also commited to this branch).
As to your review, unfortunately it was based on outdated code (I think). For example there is no Action object anymore. I'll go through this review and try to find things that still needs to be changed.
Thanks

TomaszT:
@kkasravi @clockfly RegisterApplicationMasterResponse in the current code is wrapped in case class but it is still mutable. If we really want, we should create immutable copy of this class.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/AMRMClientAsyncFactory.scala
```diff
@@ -0,0 +1,17 @@
+package org.apache.gearpump.experiments.yarn
```
clockfly: license header

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/NodeManagerCallbackHandler.scala
```diff
@@ -39,4 +40,15 @@ class NodeManagerCallbackHandler(am: ActorRef) extends NMClientAsync.CallbackHan
   def onStopContainerError(containerId: ContainerId, throwable: Throwable) {
     LOG.error(s"Container exception : $containerId", throwable)
   }
-}
\ No newline at end of file
+}
+
+trait NodeManagerCallbackHandlerFactory {
```
clockfly: Instead of declaring a factory, in scala, there is a cleaner way to do dependency injection, Like function itself.

For example. instead of 

```scala
class AmActor(...
              nodeManagerCallbackHandlerFactory: NodeManagerCallbackHandlerFactory) extends Actor {

```
You can pass function like

```scala
class AmActor(...
              nodeManagerCallbackHandlerFactory: () => CallbackHandler) extends Actor {
```

Then you can avoid declaring boilplate factory class.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClientActor.scala
```diff
@@ -1,44 +1,47 @@
...
   val LOG = LogUtil.getLogger(getClass)
   var client: AMRMClientAsync[ContainerRequest] = _
   
   override def receive: Receive = {
     case rmCallbackHandler: ResourceManagerCallbackHandler =>
       LOG.info("Received RMCallbackHandler")
-      client = start(rmCallbackHandler)
+      Try({
+        client = start(rmCallbackHandler)
+        sender ! AMRMClientAsyncStartup(Try(true))
```
clockfly: It is unclear to me what the "sender" here  represent?

Why it reply AMRMClientAsyncStartup when receiving a ResourceManagerCallbackHandler ?

Do you want to express logic like:
```scala
case Connect(resourceManagerCallbackHandler) =>
   val resoruceManagerClient = connect(resourceManagerCallbackHandler)
    sender ! Connected(resoruceManagerClient)
```

Is this what you intend to do?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
```diff
@@ -18,69 +18,113 @@
...
+
+object AmActorProtocol {
+  import AmStates._
+
+  sealed trait YarnApplicationMasterData
+  case class AMRMClientAsyncStartup(status:Try[Boolean])
+  case class LaunchContainers(containers: List[Container]) extends YarnApplicationMasterData
+  case class LaunchWorkerContainers(containers: List[Container])
+  case class LaunchServiceContainer(containers: List[Container])
+  case class ContainerRequestMessage(memory: Int, vCores: Int)
```
clockfly: Do we need Message as the suffix here? 
It is used for other case that it is not a message?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
```diff
@@ -97,19 +141,15 @@ class AmActor(appConfig: AppConfig, yarnConf: YarnConfiguration) extends Actor {
       
     case containerRequest: ContainerRequestMessage =>
       LOG.info("AM: Received ContainerRequestMessage")
-      amRMClient ! containerRequest
+      rmClientActor ! containerRequest
```
clockfly: Can we add a comment here that 
What is the response of the ContainerRequest? How the response is passed back? And Who has the responsiblity to handle that response?

From  the API, I  think the Response should be handled here.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClientActor.scala
```diff
@@ -1,44 +1,47 @@
...
       LOG.info("Received ContainerRequestMessage")
       client.addContainerRequest(createContainerRequest(containerRequest))
     case amAttr: RegisterAMMessage =>
       LOG.info(s"Received RegisterAMMessage! ${amAttr.appHostName}:${amAttr.appHostPort}${amAttr.appTrackingUrl}")
       val response = client.registerApplicationMaster(amAttr.appHostName, amAttr.appHostPort, amAttr.appTrackingUrl)
-      LOG.info("got response : " + response)
-      yarnAM ! response
-    case amStatus: AMStatusMessage =>
+      LOG.info("sending response : " + response)
+      sender ! RegisterAppMasterResponse(response=response)
```
clockfly: should have a space before and after "=". Besides as the name "response" is clear enough, maybe not necessary to do "response = response"?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClientActor.scala
```diff
@@ -1,44 +1,47 @@
...
+    case amStatus:
+      AMStatusMessage =>
```
clockfly: This two should be in same line.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClientActor.scala
```diff
@@ -1,44 +1,47 @@
...
       LOG.info("Received AMStatusMessage")
       client.unregisterApplicationMaster(amStatus.appStatus, amStatus.appMessage, amStatus.appTrackingUrl)
```
clockfly: It is confusing to me why it do unregisterApplicationMaster when it received message like AmStatusMessage?

I don't find comments about AMStatusMessage, is it a normal status of Application Master? or exit status?

experiments/yarn/src/test/scala/org/apache/gearpump/experiments/yarn/ResourceManagerClientActorSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+import org.apache.hadoop.yarn.client.api.async.AMRMClientAsync
+import org.apache.hadoop.yarn.conf.YarnConfiguration
+import org.mockito.ArgumentCaptor
+import org.scalatest.Matchers._
+import org.scalatest.{Assertions, BeforeAndAfter, FlatSpecLike}
+import org.specs2.mock.Mockito
+
+import scala.util.{Failure, Try}
+
+class ResourceManagerClientActorSpec extends TestKit(ActorSystem("testsystem"))
```
clockfly: From my understanding, the purpose of ResourceManagerClientActor is to serve as a bridge to YARN. 
If this is the case, then all message from YARN should comes to this class first, and then forwarded to related handlers.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClientActor.scala
```diff
@@ -1,44 +1,47 @@
...
   }
 
-  private[this] def createContainerRequest(attrs: ContainerRequestMessage): ContainerRequest = {
+  def createContainerRequest(attrs: ContainerRequestMessage): ContainerRequest = {
```
clockfly: Maybe private this?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClientActor.scala
```diff
@@ -47,16 +50,11 @@ class ResourceManagerClientActor(yarnConf: YarnConfiguration, yarnAM: ActorRef)
     new ContainerRequest(capability, null, null, priority)
   }
 
-  private[this] def start(rmCallbackHandler: ResourceManagerCallbackHandler): AMRMClientAsync[ContainerRequest] = {
+  private def start(rmCallbackHandler: ResourceManagerCallbackHandler): AMRMClientAsync[ContainerRequest] = {
     LOG.info("starting AMRMClientAsync")
-    import YarnApplicationMaster._
-    val amrmClient: AMRMClientAsync[ContainerRequest] = AMRMClientAsync.createAMRMClientAsync(TIME_INTERVAL, rmCallbackHandler)
+    val amrmClient: AMRMClientAsync[ContainerRequest] = clientFactory.newIntance(YarnApplicationMaster.TIME_INTERVAL, rmCallbackHandler)
```
clockfly: Do we need a type declaration here?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClientActor.scala
```diff
@@ -1,44 +1,47 @@
...
 import org.apache.hadoop.yarn.util.Records
-import akka.actor.Actor
-import akka.actor.ActorRef
-import akka.actor.actorRef2Scala
-import org.apache.gearpump.experiments.yarn.master.YarnApplicationMaster
 
-class ResourceManagerClientActor(yarnConf: YarnConfiguration, yarnAM: ActorRef) extends Actor {
+import scala.util.{Failure, Success, Try}
+
+class ResourceManagerClientActor(yarnConf: YarnConfiguration, clientFactory : AMRMClientAsyncFactory[ContainerRequest]) extends Actor {
```
clockfly: The factory can be simplified by using a functon 
()=>Client

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/NodeManagerCallbackHandler.scala
```diff
@@ -2,7 +2,7 @@ package org.apache.gearpump.experiments.yarn
 
 import java.nio.ByteBuffer
```
clockfly: We miss a license header here

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/AMRMClientAsyncFactory.scala
```diff
@@ -0,0 +1,17 @@
...
+
+import org.apache.hadoop.yarn.client.api.AMRMClient.ContainerRequest
+import org.apache.hadoop.yarn.client.api.async.AMRMClientAsync
+
+trait AMRMClientAsyncFactory[T <: ContainerRequest] {
```
clockfly: This can be simplified by using function

clockfly:
There miss license header in ServicesLauncherActor, ResourceManagerCallbackHandlerActor

clockfly:
And also ResourceManagerCallbackHandler

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
```diff
@@ -18,69 +18,113 @@
...
+object AmActorProtocol {
+  import AmStates._
+
+  sealed trait YarnApplicationMasterData
+  case class AMRMClientAsyncStartup(status:Try[Boolean])
+  case class LaunchContainers(containers: List[Container]) extends YarnApplicationMasterData
+  case class LaunchWorkerContainers(containers: List[Container])
+  case class LaunchServiceContainer(containers: List[Container])
+  case class ContainerRequestMessage(memory: Int, vCores: Int)
+  case class RMHandlerDone(reason: Reason, rMHandlerContainerStats: RMHandlerContainerStats)
```
clockfly: Can you document what is RMHandlerDone?
What it means?

It is a enum with several possible states?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
```diff
@@ -258,36 +255,37 @@ class AmActor(appConfig: AppConfig, yarnConf: YarnConfiguration) extends Actor {
...
-    amRMClient ! PoisonPill
+    rmClientActor ! PoisonPill
     success
     }
+
 }
- 
-class RMCallbackHandlerActor(appConfig: AppConfig, yarnAM: ActorRef) extends Actor {
+
+class RMCallbackHandlerActor(appConfig: AppConfig) extends Actor {
```
clockfly: The function in this class is very simple, only sending a message with no event handler defined.

Is it truely necessary to define it as an actor?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
```diff
@@ -312,7 +310,7 @@ object YarnApplicationMaster extends App with ArgumentsParser {
    * For yet unknown reason this is needed for my local pseudo distributed cluster.   
    * 
    */
-  def getForcedDefaultYarnConf:Configuration = {
+  def getForcedDefaultYarnConf:YarnConfiguration = {
```
clockfly:
```scala
 object YarnApplicationMaster extends App with ArgumentsParser {
```
Can you document what is YarnApplicationMaster? 
Should we move YarnApplicationMaster to a seperate file instead declaring in file AmActor?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
```diff
@@ -18,69 +18,113 @@
...
+  case class RMClientActorProps(props: Props)
+
+  def getRMCallbackHandlerActorProps(appConfig: AppConfig) : RMCallbackHandlerActorProps =  RMCallbackHandlerActorProps(Props(classOf[RMCallbackHandlerActor], appConfig))
+  def getRMClientActorProps(yarnConfiguration: YarnConfiguration) : RMClientActorProps =  RMClientActorProps(Props(classOf[ResourceManagerClientActor], yarnConfiguration))
+}
+
+object AmStates {
+  sealed trait Reason
+  case class Failed(throwable: Throwable) extends Reason
+  case object ShutdownRequest extends Reason
```
clockfly: Please document what AmStates is? What AmStates represents?

Especially I cannot understand what ShutdownRequest represents?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
```diff
@@ -18,69 +18,113 @@
...
+  case object AllRequestedContainersCompleted extends Reason
+}
+
+object AmActorProtocol {
+  import AmStates._
+
+  sealed trait YarnApplicationMasterData
```
clockfly: Looks like YarnApplicationMasterData only have one child, which is LaunchContainers.

Do we still need this sealed trait anymore?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
```diff
@@ -18,69 +18,113 @@
...
+  case class AMRMClientAsyncStartup(status:Try[Boolean])
```
clockfly: What is AMRMClientAsyncStartup?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
```diff
@@ -18,69 +18,113 @@
...
- * Yarn ApplicationMaster.
- */
-class AmActor(appConfig: AppConfig, yarnConf: YarnConfiguration) extends Actor {
   val LOG: Logger = LogUtil.getLogger(getClass)
-  val nodeManagerCallbackHandler = createNodeManagerCallbackHandler
-  val nodeManagerClient: NMClientAsync = createNMClient(nodeManagerCallbackHandler)
-  val rmCallbackHandler = context.actorOf(Props(classOf[RMCallbackHandlerActor], appConfig, self), "rmCallbackHandler")
-  val amRMClient = context.actorOf(Props(classOf[ResourceManagerClientActor], yarnConf, self), "amRMClient")
-  val containersStatus = collection.mutable.Map[Long, ContainerInfo]()
+  val nodeManagerCallbackHandler = nodeManagerCallbackHandlerFactory.newInstance(self)
```
clockfly: Please make val private if there are not supposed to be public.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
```diff
@@ -18,69 +18,113 @@
...
+  val nodeManagerClient: NMClientAsync = nmClientAsyncFactory.newInstance(nodeManagerCallbackHandler, yarnConf)
```
clockfly: Not necessary for the type declaration.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/ContainerCommand.scala
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.gearpump.experiments.yarn.AppConfig
+import org.apache.gearpump.experiments.yarn.Constants._
+import org.apache.gearpump.transport.HostPort
+import org.apache.gearpump.util.Constants
+import org.apache.hadoop.yarn.api.ApplicationConstants
+
+
+abstract class ContainerCommand {
+  val appConfig: AppConfig
+  val version = appConfig.getEnv("version")
```
clockfly: What is the version here represents? Gearpump's version? YARN's version?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
```diff
@@ -18,69 +18,113 @@
...
   var masterContainersStarted = 0
   var workerContainersStarted = 0
   var workerContainersRequested = 0
-
   val version = appConfig.getEnv("version")
-  
+
+  var servicesActor: Option[ActorRef] = None
+
   override def receive: Receive = {
```
clockfly: Better to break this receive handler to several handler, so that it is more clear what the case branch is actually doing.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
```diff
@@ -18,69 +18,113 @@
...
 
+object AmActor {
+  case class RMCallbackHandlerActorProps(props: Props)
+  case class RMClientActorProps(props: Props)
+
+  def getRMCallbackHandlerActorProps(appConfig: AppConfig) : RMCallbackHandlerActorProps =  RMCallbackHandlerActorProps(Props(classOf[RMCallbackHandlerActor], appConfig))
+  def getRMClientActorProps(yarnConfiguration: YarnConfiguration) : RMClientActorProps =  RMClientActorProps(Props(classOf[ResourceManagerClientActor], yarnConfiguration))
+}
+
+object AmStates {
```
clockfly: Is AmStates used else where besides AmActor? Should we make this public?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
```diff
@@ -18,69 +18,113 @@
...
+  sealed trait Reason
+  case class Failed(throwable: Throwable) extends Reason
+  case object ShutdownRequest extends Reason
+  case object AllRequestedContainersCompleted extends Reason
+}
+
+object AmActorProtocol {
```
clockfly: As message here is part of the interface, may  be we should group them and document them better?

clockfly:
@TomaszT 

I think the code is not well organized. There are many unnecesary class, the code is written more complicated than it is supposed, and the message flow seems not clear, and there are many naming issues.

1. From the code, you have ResourceManagerClientActor and AmActor. If my understanding is right, ResourceManagerClientActor  is defined to bridge message between AppLicationMaster and YARN.  If that is the case, then all events should be bridged by ResourceManagerClientActor, with no exception.

2. About the message naming. In AmStates and AmActorProtocol, there are some message that is not easy to tell what it is about. 

3. AmActor should only accept message that it can understand, I do think no YARN class should be passed to AmActor. AmActorProtocol should translate YARN mesage to message that AmActor can understand. 

4. The code seems has an assumption about the message sending order of different Actor. Ideally each Actor should be able to work independently, without assumption of other Actors.

5. There are some style issue, like  missing license header and declaring unnessary factory. 

6. In some Actor definition, there is no event handler defined, we are using the side effect of Actor to do something asyncly. If we don't have event handler, how about using future instead of Actor?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/NodeManagerCallbackHandler.scala
kkasravi: Added to [PR](https://github.com/TomaszT/gearpump/pull/6) on Tomek's branch
Not all factories have been eliminated but most
```
class AmActor(appConfig: AppConfig, yarnConf: YarnConfiguration,
              rmCallbackHandlerActorProps: RMCallbackHandlerActorProps,
              rmClientActorProps: RMClientActorProps,
              createNMClient: (NodeManagerCallbackHandler, YarnConfiguration) => NMClientAsync,
              createNodeManagerCallbackHandler: (ActorRef) => NodeManagerCallbackHandler,
              containerLaunchContextFactory: ContainerLaunchContextFactory) extends Actor {
```

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClientActor.scala
kkasravi: I think this part is also a bit confusing - an activity or sequence diagram would help.
ResourceManagerCallbackHandler is sent by RMCallbackHandlerActor and ResourceManagerCallbackHandlerActor. Both Actors seem very similar - maybe one is for testing only?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
kkasravi: This is due to ContainerRequest being a YARN type - but perhaps the latter can be disambiguated by full package references.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
kkasravi: This is early drawing of message exchange (it may differ now).
yarnam mermaid
Different actor creation is to eliminate Thread manipulation (if you look at many non trivial AppMasters for example [spark](https://github.com/freedafeng/adspark/blob/bebd1924a9b434fcb4fa719f515fe910fd663ba2/yarn/stable/src/main/scala/org/apache/spark/deploy/yarn/ApplicationMaster.scala) you'll see many Thread specific operations. The goal was to have actors exchange messages to do this work rather than do it within different Thread contexts.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClientActor.scala
kkasravi: Fixed in PR on Tomek's branch

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClientActor.scala
kkasravi: Fixed in [PR](https://github.com/TomaszT/gearpump/pull/6) on Tomek's branch

experiments/yarn/src/test/scala/org/apache/gearpump/experiments/yarn/ResourceManagerClientActorSpec.scala
kkasravi: Yes I think this is the basic approach - most messages are either from amactor (asking ResourceManagerClientActor to do something with AMRMClient or returning info to amActor that AMRMClient has done

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClientActor.scala
kkasravi: referenced in *Spec so can not be private

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClientActor.scala
kkasravi: Fixed in [PR](https://github.com/TomaszT/gearpump/pull/6) on Tomek's branch

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClientActor.scala
kkasravi: Fixed in [PR](https://github.com/TomaszT/gearpump/pull/6) on Tomek's branch

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/NodeManagerCallbackHandler.scala
kkasravi: Fixed in [PR](https://github.com/TomaszT/gearpump/pull/6) on Tomek's branch

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/AMRMClientAsyncFactory.scala
kkasravi: Fixed in [PR](https://github.com/TomaszT/gearpump/pull/6) on Tomek's branch

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
kkasravi: This was a message to indicate that the ResourceManagerClient had successfully started the AMRMClient. I will try and create a new diagram with Tomek to better show the message exchange

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/ContainerCommand.scala
kkasravi: This is GearPump's version as provided in the cliOpts not HOCON - should probably be either moved to a conf file or represented there. BTW although we have version.sbt we do not have a way of moving this into a HOCON representation by script or programmatically in scala or otherwise and probably should.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
kkasravi: States and Data (Reason) will be moved info FSM eg AmActor extends Actor with FSM[AmState,AmData] or something like that. Tomek will provide details on this which are being worked on in #710.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
clockfly: RMCallbackHandlerActor looks unnecessary.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClientActor.scala
kkasravi: RMCallbackHandlerActor has been removed in [PR 923](https://github.com/intel-hadoop/gearpump/pull/923)

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
kkasravi: Removed in [PR 923](https://github.com/intel-hadoop/gearpump/pull/923)

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/AmActor.scala
kkasravi: This has been addressed in [PR 923](https://github.com/intel-hadoop/gearpump/pull/923)

kkasravi:
Closing this one since it has been replaced by PR #923

PR #914, opened 2015-05-12
fix #6, add transaction api

### LINKED ISSUE #6: Support exactly once message delivery guarantee

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/914)

### DIFF
--- a/experiments/state/src/main/scala/org/apache/gearpump/streaming/state/api/State.scala
+++ b/experiments/state/src/main/scala/org/apache/gearpump/streaming/state/api/State.scala
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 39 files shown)


experiments/state/src/main/scala/org/apache/gearpump/streaming/state/api/State.scala
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * State API provides a container for users to keep their in memory state
+ * persistently.
+ */
+trait State[T] {
+  /**
+   * set a persistent store that will checkpoint state periodically
+   * @param checkpointStore
+   */
+  def setCheckpointStore(checkpointStore: CheckpointStore): Unit
```
whjiang: who is the caller of this function?

Or more generally, when reading the APIdoc of this class, I don't know who is the caller and in which circumentance will these APIs been called. Better to documnet it more clear.

whjiang: What's the realationship between State and StateOp? I had thought State is the real state, but it seems State is more like a StateManager. If so, better to rename State to some name like StateManager to avoid confusing. As I just read some fragment of code, I am highly possible wrong.

clockfly:
Please study this paper about the API design
http://static.googleusercontent.com/media/research.google.com/en/us/pubs/archive/41378.pdf

manuzhang:
thanks for the reference, I will study it and improve the State API

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/api/State.scala
manuzhang: I've mixed framework API and user API. I'd better hide the former.

manuzhang:
I will rework the interface. Besides that, the exact-once implementation is not so different from the [current docs](http://www.gearpump.io/site/0.3/internal/). 

Now the window computation is decoupled from checkpointing. A new message will update states of all windows it belongs to immediately. No need to merge states on window expiration. 

Suppose a user wants to count messages in the last 5 seconds at every second (a sliding window). A message at second 2 will update states of windows `[0, 5), [1, 6), [2, 7)`. On checkpoint, we checkpoint all the in memory states with the checkpoint timestamp. State of a window is removed when the window expires.

manuzhang:
opened #951 for this

PR #916, opened 2015-05-13
fix #602: dashboard will freeze when server is unreachable

### LINKED ISSUE #602: Service unreachable notice
When server is disconnected, the dashboard ui should be freeze or blocked.

stanleyxu2005:
also uses angularstrap to replace diverse tiny angular directive projects (e.g. angular-tooltips)


clockfly:
+1

PR #917, opened 2015-05-14
fix #907: calculate application clock update frequence for a 30 secon…

### LINKED ISSUE #907: clock could rewind on dashboard
The dashboard requests min clock update at certain intervals. During the interval, the clock advances at client clock's pace such that it looks going forward smoothly. This will lead to clock rewind when the application clock advances at a different pace. In extreme cases, what if the application clock pauses for some time ?

stanleyxu2005:
…d time frame


clockfly:
+1

PR #923, opened 2015-05-16
Fixes #661 Add unittests for Yarn

### LINKED ISSUE #661: Add unittests for Yarn

kkasravi:
@clockfly @TomaszT Please use this PR rather than [PR 911](https://github.com/intel-hadoop/gearpump/pull/911) since the latest changes have been extensive. UT tests have been preserved and suggestions noted below in PR 911 have been incorporated.

- [X] From the code, you have ResourceManagerClientActor and AmActor. If my understanding is right, ResourceManagerClientActor is defined to bridge message between AppLicationMaster and YARN. If that is the case, then all events should be bridged by ResourceManagerClientActor, with no exception.

- [X] About the message naming. In AmStates and AmActorProtocol, there are some message that is not easy to tell what it is about.

- [X] AmActor should only accept message that it can understand, I do think no YARN class should be passed to AmActor. ResourceManagerClientActor should translate YARN mesage to message that AmActor can understand.

- [X] The code seems has an assumption about the message sending order of different Actor. Ideally each Actor should be able to work independently, without assumption of other Actors.

- [x] There are some style issue, like missing license header and declaring unnessary factory.

- [X] In some Actor definition, there is no event handler defined, we are using the side effect of Actor to do something asyncly. If we don't have event handler, how about using future instead of Actor?


kkasravi:
Design message flow:

1. YarnApplicationMaster on startup must instantiate both a AMRMClientAsync and a NMClientAsync and provide each with callback handlers. These two YARN clients handle ResourceManager and NodeManager messages respectively. The AMRMClientAsync is started in a separate actor described in 2. below. The NMClientAsync is instantiated within the YarnApplicationMaster. After creating the NMClientAsync within it's variable initialization the YarnApplicationMaster's preStart will instantiate a subclass of AMRMClientAsync.CallbackHandler called ResourceManagerCallbackHandler and put a reference to itself in this subclass. At this point one of the two clients the YarnApplicationMaster needs has been created (NMClientAsync) a provided a callback handler.

2.  YarnApplicationMaster then creates a ResourceManagerClient actor and sends it the ResourceManagerCallbackHandler it initialized in 1.

3. When the ResourceManagerClient receives the ResourceManagerCallbackHandler it instantiates a AMRMClientAsync and gives it the ResourceManagerCallbackHandler. The ResourceManagerClient then sends a message back to the YarnApplicationMaster of AMRMClientAsyncStartup(status)

4. When the YarnApplicationMaster receives the AMRMClientAsyncStartup(status=success) it now knows that both YARN clients have been created and intialized. The YarnApplicationMaster then sends the ResourceManagerClient a RegisterAMMessage(hostname, port, trackingURL). 

5. Upon receiving the RegisterAMMessage, the ResourceManagerClient will then make a synchronous YARN API call AMRMClientAsync.registerApplicationMaster. It then sends the response back within a RegisterAppMasterResponse to the YarnApplicationMaster.

6. Upon receiving the RegisterAppMasterResponse the YarnApplicationMaster creates a request for one or more master containers by sending a ContainerRequestMessage to the ResourceManagerClient.

7. Upon receiving the ContainerRequestMessage, the ResourceManagerClient calls AMRMClientAsync.addContainerRequest with the ContainerRequestMessage information.

8. The YARN RM will next notify the ResourceManagerClient that a container has been allocated by calling the ResourceManagerCallbackHandler.onContainersAllocated. Within this method a LaunchContainers message is sent back to the YarnApplicationMaster.

9. Upon receiving a LaunchContainers message the YarnApplicationMaster checks its internal state and may either ask for more masters or more workers. It then calls NMClientAsync.startContainerAsync providing the launch command.

10. The NMClientAsync callback handler NodeManagerCallbackHandler.onContainersStarted will then be called by YARN as soon as each container starts up. This callback handler then sends a ContainerStarted to the YarnApplicationMaster. 

11. Upon receiving a ContainerStarted the YarnApplicationMaster checks its internal state and transitions to request workers if all masters have been allocated and launched. 

TBD A Diagram describing the above flow.

clockfly:
@kkasravi 

Thank you! 

There are several comments that are still valid:
https://github.com/intel-hadoop/gearpump/pull/911#discussion_r30295523
https://github.com/intel-hadoop/gearpump/pull/911#discussion_r30295254
https://github.com/intel-hadoop/gearpump/pull/911#discussion_r30295954
https://github.com/intel-hadoop/gearpump/pull/911#discussion_r30296237
https://github.com/intel-hadoop/gearpump/pull/911#discussion_r30296437

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
```diff
@@ -0,0 +1,338 @@
...
+  var masterAddr: Option[HostPort] = None
+  var masterContainersStarted = 0
+  var workerContainersStarted = 0
+  var workerContainersRequested = 0
+  val version = appConfig.getEnv("version")
+
+  var servicesActor: Option[ActorRef] = None
+
+  override def preStart(): Unit = {
+    val rmCallbackHandler = createRMCallbackHandler(appConfig, self)
```
clockfly: Why creating RMCallbackHandler here? instead of creating in RmClientActor?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClient.scala
```diff
@@ -1,44 +1,46 @@
...
   val LOG = LogUtil.getLogger(getClass)
   var client: AMRMClientAsync[ContainerRequest] = _
   
   override def receive: Receive = {
     case rmCallbackHandler: ResourceManagerCallbackHandler =>
       LOG.info("Received RMCallbackHandler")
-      client = start(rmCallbackHandler)
+      Try({
+        client = start(rmCallbackHandler)
+        sender ! AMRMClientAsyncStartup(Try(true))
```
clockfly: I suggest wrap rmCallbackHandler in object named Connect(rmCallbackHandler), and wrap cient in Connected(client)

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
```diff
@@ -51,23 +52,26 @@ trait ClientAPI {
...
 
   val LOG: Logger = LogUtil.getLogger(getClass)
   def getConfiguration = configuration
   def getEnv = getConfiguration.getEnv _
   def getYarnConf = yarnConf
-  def getFs = FileSystem.get(getYarnConf)  
+  def getFs = fileSystem
   def jarPath = new Path(getFs.getHomeDirectory, getEnv(HDFS_ROOT) + "/jars/" )
 
   val version = configuration.getEnv("version")
```
clockfly: We should remove "version" or making it more explicit?

Why AppConfig will provide method like getEnv(), it is like openning a bottle but we don't know what it contains.
Maybe AppConfig should be more strong-typed, with clear method name to indicate what method it contains.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
```diff
@@ -162,44 +168,120 @@ class Client(configuration:AppConfig, yarnConf: YarnConfiguration, yarnClient: Y
...
-    while (appState != YarnApplicationState.FINISHED &&
-      appState != YarnApplicationState.KILLED &&
-      appState != YarnApplicationState.FAILED) {
-      Thread.sleep(1000)
-      appReport = yarnClient.getApplicationReport(appId)
-      appState = appReport.getYarnApplicationState
+    val monitor = getEnv("monitor").toBoolean
+    var watch = true
+
+    while(watch) {
```
clockfly: The while loop here is suspicious.

From my understanding, the appState  will not change.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
```diff
@@ -0,0 +1,338 @@
...
+import org.apache.hadoop.yarn.api.records.{Container, ContainerId, FinalApplicationStatus}
+import org.apache.hadoop.yarn.client.api.AMRMClient.ContainerRequest
+import org.apache.hadoop.yarn.client.api.async.impl.NMClientAsyncImpl
+import org.apache.hadoop.yarn.client.api.async.{AMRMClientAsync, NMClientAsync}
+import org.apache.hadoop.yarn.conf.YarnConfiguration
+import org.slf4j.Logger
+
+import scala.util.{Failure, Success, Try}
+
+object AmStates {
```
clockfly: What is the meaning of AmStates?

Application master state? 
how to understand ShutdownRequest ?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
```diff
@@ -0,0 +1,338 @@
...
+
+object AmActorProtocol {
+  import AmStates._
+
+  case class AMRMClientAsyncStartup(status: Try[Boolean])
+  case class LaunchContainers(containers: List[Container])
+  case class LaunchWorkerContainers(containers: List[Container])
+  case class LaunchServiceContainer(containers: List[Container])
+  case class ContainerRequestMessage(memory: Int, vCores: Int)
+  case class RMHandlerDone(reason: Reason, rMHandlerContainerStats: RMHandlerContainerStats)
```
clockfly: It is difficult to understand what RMHandlerDone actually means?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
```diff
@@ -0,0 +1,338 @@
...
+                            createNodeManagerCallbackHandler: (ActorRef) => NodeManagerCallbackHandler,
+                            containerLaunchContextFactory: (String) => org.apache.hadoop.yarn.api.records.ContainerLaunchContext) extends Actor {
+
+  import AmActorProtocol._
+  import AmStates._
+
+  val LOG: Logger = LogUtil.getLogger(getClass)
+  val nodeManagerCallbackHandler = createNodeManagerCallbackHandler(self)
+  val nodeManagerClient: NMClientAsync = createNMClient(nodeManagerCallbackHandler, yarnConf)
+  val rmClientActor = context.actorOf(propsRMClient, "rmClient")
```
clockfly: Why creating nodeManagerClient inside YarnAppMaster? It create a dependency which may not be necessary.
Can we delay the creation till YarnAppMaster is constructed?

clockfly: Why creating rmClientActor  inside YarnApplicationMaster? Can you pass the rmClientActor  into YarnApplicaitonMaster?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
```diff
@@ -0,0 +1,338 @@
...
+
+    case AMRMClientAsyncStartup(status) =>
+      status match {
+        case Success(true) =>
+          LOG.info("received AMRMClientAsyncStartup")
+          val port = appConfig.getEnv(YARNAPPMASTER_PORT).toInt
+          val target = host + ":" + port
+          val addr = NetUtils.createSocketAddr(target)
+          rmClientActor ! RegisterAMMessage(addr.getHostName, port, trackingURL)
+        case Success(false) =>
```
clockfly: It is really confusing to see something like Success(false)

Maybe we should create different types to represent different state in the case match.

Like 
ResourceManagerConnected
ResourceManagerConnectionFailed(ex: Throwable)

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
```diff
@@ -0,0 +1,338 @@
...
+  private def setMasterAddrIfNeeded(containers: List[Container]) {
+    masterAddr match {
+      case None =>
+        masterAddr = Some(HostPort(containers.head.getNodeId.getHost, appConfig.getEnv(GEARPUMPMASTER_PORT).toInt))
+      case _ =>
+    }
+
+  }
+
+  private def needMoreMasterContainersState: Boolean = {
```
clockfly: what needMoreMasterContainersState mean?
It means need more master contaner?
Or need more master container state?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
```diff
@@ -0,0 +1,338 @@
...
+
+  }
+
+  private def cleanUp(done: RMHandlerDone): Boolean = {
+    LOG.info("Application completed. Stopping running containers")
+    nodeManagerClient.stop()
+    var success = true
+
+    val stats = done.rMHandlerContainerStats
+    done.reason match {
```
clockfly: Again, I dislike the "done.reason" case match here.  Can we use class hiearachy instead of member fields?

For example: 
sealed trait ResourceManagerHandlerStatus
class ResourceManagerHandlerFailed extends ResourceManagerHandlerStatus

By the way, I cannot understand what RmHandler actually mean.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
```diff
@@ -0,0 +1,338 @@
...
+      case failed: Failed =>
+        val message = s"Failed. total=${appConfig.getEnv(WORKER_CONTAINERS).toInt}, completed=${stats.completed}, allocated=${stats.allocated}, failed=${stats.failed}"
+        rmClientActor ! AMStatusMessage(FinalApplicationStatus.FAILED, message, null)
+        success = false
+      case ShutdownRequest =>
```
clockfly: It is unclear to see match like:

done.reason match {
  case xx: 
     if (yy) {
     }
    else {
    }
}

Maybe we should flatten all these with clear type and names.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
```diff
@@ -0,0 +1,338 @@
...
+  val TIME_INTERVAL = 1000
+
+  override val options: Array[(String, CLIOption[Any])] = Array(
+    APPMASTER_IP -> CLIOption[String]("<Gearpump master ip>", required = false),
+    APPMASTER_PORT -> CLIOption[String]("<Gearpump master port>", required = false),
+    "version" -> CLIOption[String]("<Gearpump version>", required = true)
+  )
+
+  /**
+   * For yet unknown reason this is needed for my local pseudo distributed cluster.
```
clockfly: Is this fixed? Maybe remove this if fixed?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
```diff
@@ -0,0 +1,338 @@
...
+    APPMASTER_IP -> CLIOption[String]("<Gearpump master ip>", required = false),
+    APPMASTER_PORT -> CLIOption[String]("<Gearpump master port>", required = false),
+    "version" -> CLIOption[String]("<Gearpump version>", required = true)
+  )
+
+  /**
+   * For yet unknown reason this is needed for my local pseudo distributed cluster.   
+   *
+   */
+  def getForcedDefaultYarnConf: YarnConfiguration = {
```
clockfly: What you mean by "forced"?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/ResourceManagerCallbackHandler.scala
```diff
@@ -1,21 +1,39 @@
...
-import org.apache.hadoop.yarn.api.records.Container
-import org.apache.gearpump.experiments.yarn.Actions._
-import scala.collection.JavaConversions._
+
 import scala.collection.JavaConverters._
-import org.apache.hadoop.yarn.api.records.ContainerExitStatus
-import org.apache.gearpump.experiments.yarn.Constants._
-import org.apache.gearpump.experiments.yarn.AppConfig
 
 class ResourceManagerCallbackHandler(appConfig: AppConfig, am: ActorRef) extends AMRMClientAsync.CallbackHandler {
```
clockfly: I think ResourceManagerCallbackHandler should only pass info to ResourceManagerClient

instead of passing to am?

clockfly: Besides, in function onContainersAllocated, there is call
``` am ! LaunchContainers(allocatedContainers.asScala.toList)```

I think it should only send events like ContainerAllocated instead of sending command like LaunchContainers
As it is the listener's responsbility to decide how to handle event like ContainerAllocated

File: experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala (hunk 1 above)
Comment (kkasravi) [inline experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala]: RmCallbackHandler is used to coordinate with the ResourceManagerClient. Upon reception the ResourceManagerClient knows that 
1. It can try and create the AMRMClientAsync
2. Use both the actorRef and appConfig in the ResourceManagerClient as data injected by the YarnApplicationMaster that it uses within the callback methods and to communicate with the 'sender'.
3. Signal to the YarnApplicationMaster that it was successful/unsuccessful in creating AMRMClientAsync client.

File: experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala (hunk 3 above)
Comment (kkasravi) [inline experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala]: I think we should make "version" more explicit so that we have a 'version' field that will always match version.sbt. This should be part of the build process and should probably be handled separately.

Comment (kkasravi) [inline experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala]: For getEnv - once "version" is handled in the same way as YARNAPPMASTER_COMMAND then we could make it more strongly typed. Though getEnv is used in both Client and YarnApplicationMaster and is intended to be somewhat generic. Let me see if that can be made more like UserConfig.getValue[T].

File: experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala (hunk 4 above)
Comment (kkasravi) [inline experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala]: appState will change within the loop see for example
```
        case YarnApplicationState.ACCEPTED =>
          LOG.info(s"Application $appId is $appState")
          val (ar, as) = delay(appId)
          appReport = ar
          appState = as
```

File: experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala (hunk 8 above)
Comment (kkasravi) [inline experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala]: Changed to use RMConnected, RMConnectionFailed(throwable)

File: experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala (hunk 9 above)
Comment (kkasravi) [inline experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala]: For HA Master, need more master containers allocated. This will go away when #710 is fixed. (Akka FSM)

File: experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala (hunk 13 above)
Comment (kkasravi) [inline experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala]: Changed to loadYarnConfUsingHadoopConfEnv

File: experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala (hunk 12 above)
Comment (kkasravi) [inline experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala]: Doesn't appear to be. Renamed to loadYarnConfUsingHadoopConfEnv

File: experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala (hunk 1 above)
Comment (kkasravi) [inline experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala]: YarnApplicationMaster now no longer requires a createNMCallbackHandler method. DI is still used to create YarnApplicationMaster.{nodeManagerClient, resourceManagerClient}

File: experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/ResourceManagerCallbackHandler.scala (hunk 14 above)
Comment (kkasravi) [inline experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/ResourceManagerCallbackHandler.scala]: @clockfly @TomaszT I think this is a valid point. In summary:

1. All methods in ResourceManagerCallbackHandler should only message ResourceManagerClient which then can message YarnApplicationMaster. 
2. ResourceManagerClient should only message YarnApplicationMaster about state changes that occurred in its callback handler. YarnApplicationMaster should change its own state in response to these messages. I think this will simplify design and moving to FSM. 

I'll make these changes as soon as i can.

File: experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala (hunk 1 above)
Comment (clockfly) [inline experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala]: I am not convinced that why putting RMCallbackHandler here.

I think you can put RMCallbackHandler in ResourceManagerClient. All message in RMCallbackHandler will be first sent to ResourceManagerClient, and then ResourceManagerClient forward back to YarnAppMaster if needed. 

It is strange to me that RMCallbackHandler accept an actorRef  of YarnAppMaster, it should accept actorRef of ResourceManagerClient instead.

File: experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClient.scala [hunk 15]
```diff
@@ -1,44 +1,45 @@
...
   val LOG = LogUtil.getLogger(getClass)
   var client: AMRMClientAsync[ContainerRequest] = _
   
   override def receive: Receive = {
     case rmCallbackHandler: ResourceManagerCallbackHandler =>
       LOG.info("Received RMCallbackHandler")
-      client = start(rmCallbackHandler)
+      Try({
+        client = start(rmCallbackHandler)
+        rmCallbackHandler.am ! RMConnected
```
clockfly: It is strange to get YarnAppMaster actorRef by this way.
How about we passing the ActorRef of yarnAppMaster in constuctor directly?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
clockfly: Why is AppConfig  needed in the first place?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
```diff
@@ -162,44 +168,118 @@ class Client(configuration:AppConfig, yarnConf: YarnConfiguration, yarnClient: Y
     })
   }
 
-  def monitorAM(appContext: ApplicationSubmissionContext): Unit = {
-    val appId = appContext.getApplicationId
+  def delay(appId: ApplicationId): (ApplicationReport, YarnApplicationState) = {
```
clockfly: Please define return value for public method.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
```diff
@@ -162,44 +168,118 @@ class Client(configuration:AppConfig, yarnConf: YarnConfiguration, yarnClient: Y
...
     var appReport = yarnClient.getApplicationReport(appId)
     var appState = appReport.getYarnApplicationState
-    while (appState != YarnApplicationState.FINISHED &&
-      appState != YarnApplicationState.KILLED &&
-      appState != YarnApplicationState.FAILED) {
-      Thread.sleep(1000)
-      appReport = yarnClient.getApplicationReport(appId)
-      appState = appReport.getYarnApplicationState
+    val monitor = getEnv("monitor").toBoolean
+    var watch = true
```
clockfly: What is the different between monitor and watch?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
```diff
@@ -162,44 +168,118 @@ class Client(configuration:AppConfig, yarnConf: YarnConfiguration, yarnClient: Y
...
+        case YarnApplicationState.SUBMITTED =>
+          val (ar, as) = delay(appId)
+          appReport = ar
+          appState = as
+        case YarnApplicationState.ACCEPTED =>
+          val (ar, as) = delay(appId)
+          appReport = ar
+          appState = as
+        case YarnApplicationState.RUNNING =>
+          monitor match {
```
clockfly: Why we need to check monitor again?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
```diff
@@ -162,44 +168,118 @@ class Client(configuration:AppConfig, yarnConf: YarnConfiguration, yarnClient: Y
...
-  def monitorAM(appContext: ApplicationSubmissionContext): Unit = {
-    val appId = appContext.getApplicationId
+  def delay(appId: ApplicationId): (ApplicationReport, YarnApplicationState) = {
+    Thread.sleep(1000)
+    val appReport = yarnClient.getApplicationReport(appId)
+    val appState = appReport.getYarnApplicationState
+    (appReport, appState)
+  }
+
+  def monitor(appId: ApplicationId): Unit = {
```
clockfly: I suggest removing monitor completely. It doesn't seems necesary. 
As yarn has its own dashboard and UI command to track application status.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
```diff
@@ -0,0 +1,324 @@
...
+          success = false
+        }
+      case allRequestedContainersCompleted: RMAllRequestedContainersCompleted =>
+        val message = s"Diagnostics. total=${appConfig.getEnv(WORKER_CONTAINERS).toInt}, completed=${stats.completed}, allocated=${stats.allocated}, failed=${stats.failed}"
+        resourceManagerClient ! AMStatusMessage(FinalApplicationStatus.SUCCEEDED, message, null)
+        success = true
+    }
+
+    resourceManagerClient ! PoisonPill
+    success
```
clockfly: What the return value "success" here mean?
If "success" == false, it means cleanup failed?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
```diff
@@ -0,0 +1,324 @@
...
+
+  }
+
+  private def cleanUp(state: RMTerminalState): Boolean = {
+    LOG.info("Application completed. Stopping running containers")
+    nodeManagerClient.stop()
+    var success = true
+
+    val stats = state.containerStats
+    state match {
```
clockfly: What we do with containerStats is only logging?

Maybe put the logging code at where the error happen is better?

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
kkasravi: For delay return value is (ApplicationReport, YarnApplicationState)

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
kkasravi: if -monitor is provided on the command line then the client will poll the appmaster state until a terminal state is reached (FINISHED, KILLED, FAILED). Otherwise the client will exit when a non-terminal state of RUNNING is reached.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
kkasravi: See above note about whether we watch the app until a terminate state is reached or just until the app is RUNNING.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
kkasravi: You may want to script additional actions to be completed after the application exits - so it would have value in these cases where particular clean could be done after appmaster exists. This could be done using curl and parsing the application's state from a GET but would be much more involved.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
kkasravi: removed - return value was never checked (or used)

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
kkasravi: Done.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
kkasravi: AppConfig looks for CLI options that override an environment or conf value. EG -monitor

kkasravi:
@clockfly @TomaszT 
ResourceManagerCallbackHandler now only messages ResourceManagerClient. No commands are launched from ResourceManagerCallbackHandler, rather a state change is sent to ResourceManagerClient which may then be sent to the ApplicationMaster and may result in a command being issued. This is more in line with Sean's suggestions.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
kkasravi: removed

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
kkasravi: changed to have explicit values embedded within the name

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
kkasravi: Changed to use RMError, RMShutdownRequest, RMAllRequestedContainersCompleted

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
kkasravi: done

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/ResouceManagerClient.scala
kkasravi: applicationMaster is now pulled from context.parent

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
clockfly: The script is a bash script, right?
If it involves another process, I don't think its the client's responsiblity to query whether the app is finished or not.

The script should be easy to query the app's status themselves?

Besides, are there such kind of scripts in Gearpump project for now? I think we should not include logic that may happen in future.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
clockfly: Yeah, for public method, the return value should be clearly defined  in method signatue.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/ResourceManagerCallbackHandler.scala
```diff
@@ -60,19 +76,19 @@ class ResourceManagerCallbackHandler(appConfig: AppConfig, am: ActorRef) extends
 
     //todo:wont work with non uniform containers
     (0 until requestCount).foreach(request => {
-      am ! ContainerRequestMessage(containerCount, appConfig.getEnv(GEARPUMPMASTER_VCORES).toInt)
+      resourceManagerClient ! AdditionalContainersRequest(containerCount)
```
clockfly: I am not sure whether it is right to track the allocation status in this callback.
```scala
-  val allocatedContainersCount = new AtomicInteger(0)		+  private val completedContainersCount = new AtomicInteger(0)
-  val requestedContainersCount = new AtomicInteger(0)		+  private val failedContainersCount = new AtomicInteger(0)
+  private val allocatedContainersCount = new AtomicInteger(0)
+  private val requestedContainersCount = new AtomicInteger(0)
```

Should we decide how many additionaly container to allocate here by AdditionalContainersRequest or just forward the mesage ContianerAllocated to related parties.

With some minor modification, the code maybe greatly simplified. 
1. All message from ResourceManager extends a sealed trait like  ResourceManagerResponse.

2. Callback send the response to ResourceManagerClient, like RMError extends ResourceManagerResponse

3. In ResourceManagerClient, there is a receiver to handle ResourceManagerResponse, what it does is to forward that to YarnAppMaster.

4. The status like how many container are allocated, should be tracked by YarnAppMaster.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/client/Client.scala
kkasravi: Removed -monitor option

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/ResourceManagerCallbackHandler.scala
kkasravi: @clockfly Yes I like that logic since we would be separating Callback messages from YarnAppMaster messages. However I suggest we do this in #710 since it will also incorporate FSM. Note that not all states are being tracked right now - for example in NodeManagerCallback, many of the callback methods are stubbed and do not propagate state forward to anyone. I think getting state right in #710 will take some work but should incorporate the above suggestion.

clockfly: It is not a big change.
Just move the state from Callback to YarnApplicatonMaster. And the Callback only do message forwading.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
```diff
@@ -0,0 +1,315 @@
...
+
+  private def cleanUp(state: RMTerminalState): Unit = {
+    LOG.info("Application completed. Stopping running containers")
+    nodeManagerClient.stop()
+
+    state match {
+      case RMError(throwable, stats) =>
+        LOG.info("Failed", throwable.getMessage)
+        val message = s"Failed. total=${appConfig.getEnv(WORKER_CONTAINERS).toInt}, completed=${stats.completed}, allocated=${stats.allocated}, failed=${stats.failed}"
+        resourceManagerClient ! AMStatusMessage(FinalApplicationStatus.FAILED, message, null)
```
clockfly: This is not what I expected.

As the mesage originate from ResoourcemanagerCallback? Why don't we just log at ResoourcemanagerCallback, instead of passing the message back and forth?

In this case, RMTerminalState will be simplified, as it no longer need a field containerStats

kkasravi: Agreed, this logic is convoluted. AMStatusMessage may be able to be eliminated entirely.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/ResourceManagerCallbackHandler.scala
kkasravi: Yes, ok. This is consistent with state only being handled within the YarnApplicationMaster. I'll make this change ...

clockfly:
Besides, please check the code coverage for YARN 
https://codecov.io/github/intel-hadoop/gearpump/src/main/scala/org/apache/gearpump/experiments/yarn?ref=master

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/ResourceManagerCallbackHandler.scala
kkasravi: ResourceManagerCallbackHandler.getProgress is inherited from AMRMClientAsync.CallbackHandler and prevents moving container counts into YarnApplicationMaster.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/YarnApplicationMaster.scala
kkasravi: Removed AMStatusMessage. Logging of terminal state and unregistering application master is now done in ResourceManagerClient.

experiments/yarn/src/main/scala/org/apache/gearpump/experiments/yarn/master/ResourceManagerCallbackHandler.scala
clockfly: First, the getProgress means whether the applicaiton is finished, instead of whether the container is allocated.

As Gearpump is long running, so  the progress should always be smaller than < 100%.
I think we can adopt a very simple strategy here, we always  set the progress to 50%, as the YarnAppMaster is already started, and it is long running.

clockfly:
```
ResourceManagerCallbackHandler.getProgress is inherited from AMRMClientAsync.CallbackHandler and prevents moving container counts into YarnApplicationMaster.
```

First, the getProgress means whether the applicaiton is finished, instead of whether the container is allocated.

As Gearpump is long running, so the progress should always be smaller than < 100%.
I think we can adopt a very simple strategy here, we always set the progress to 50%, as the YarnAppMaster is already started, and it is long running.

TomaszT:
How to check what will be the coverage after merging issue_661 into master?

clockfly:
@Tomasz

yo can check the codecov link

kkasravi:
This is being closed and moved to #710 since state is now being handled.

PR #927, opened 2015-05-20
fix #926: previous restcall should be dropped if the page is destroyed

### LINKED ISSUE #926: When server is restarted, dashboard might show unreachable notice
If user is viewing an application on dashboard and the server is restarted, the application does not exist any more. Now user switch to application list page (without reloading the whole page), dashboard will show an unreachable notice periodically.

stanleyxu2005:


huafengw:
+1

PR #938, opened 2015-05-22
issue #934, allow user to define uid for edge partitioner.

clockfly:


streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/TaskLocator.scala
```diff
@@ -20,70 +20,55 @@ package org.apache.gearpump.streaming.appmaster
...
+          TaskId(0,0), TaskId(0,1)
+        ]
       }
-      worker2 {
-        "task1" = 2
+      app2: {
+        workerId: [
+          TaskId(1,0), TaskId(1,1)
+        ]
       }
```
huafengw: Will TaskId be too low level? Suppose user have a very complex dag, how should they know the mapping from group id to the exact node in the DAG?

clockfly: 1. Mostly, the data is stored by AppMaster automatically, so end user don't need to understand much details, so that it can keep its locality when there is AppMaster recovery happen.

2. If user want to change it theirselves, this can be easily done, as user can do a easy query to get the id and the parallism of each processor.

huafengw:
+1

PR #939, opened 2015-05-22
fix #928 continue to improve the costly function allowSendingMoreMess…

### LINKED ISSUE #928: costly allowSendingMoreMessages() hurt performance
Every time TaskActor handles a message, it will call function allowSendingMoreMessages(), which is observed to be a costly function.

huafengw:
Use List instead of Map


clockfly:
+1

PR #942, opened 2015-05-25
issue #941, allow user to define custom partitioner by REST

clockfly:


streaming/src/main/scala/org/apache/gearpump/streaming/StreamApplication.scala
```diff
@@ -76,7 +76,9 @@ class StreamApplication(override val name : String,  inputUserConfig: UserConfig
 
 object StreamApplication {
 
-  def apply (name : String, dag: Graph[_ <: Processor[Task], _ <: Partitioner], userConfig: UserConfig): StreamApplication = {
+  val defaultPartitioner = new HashPartitioner()
```
huafengw: maybe rename to hashPartitioner

streaming/src/main/scala/org/apache/gearpump/streaming/task/Subscription.scala
```diff
@@ -42,7 +45,7 @@ class Subscription(
 
   val LOG: Logger = LogUtil.getLogger(getClass, app = appId, executor = executorId, task = taskId)
 
-  import subscriber.{partitioner, processorId, processor}
+  import subscriber.{partitioner => partitionerDescription, processorId, processor}
```
huafengw: Should do the refactor in case class Subscriber.

huafengw:
+1

PR #943, opened 2015-05-25
fix #890: add cloudera manager integration support for gearpump

### LINKED ISSUE #890: add cloudera manager integration support
support distribute gearpump package and start/stop gearpump services via Cloudera Manager.

whjiang:
This is the first part: gearpump change. The parcel build script will be moved out of gearpump as a standalone project in second part.


clockfly:
+1

PR #951, opened 2015-05-26
fix #6, add transaction api

### LINKED ISSUE #6: Support exactly once message delivery guarantee

manuzhang:


clockfly:
Good work, it will require some time to review it.

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/user/example/processor/WindowAverageProcessor.scala
```diff
@@ -0,0 +1,57 @@
...
+
+object WindowAverageProcessor {
+  val LOG: Logger = LogUtil.getLogger(classOf[WindowAverageProcessor])
+}
+
+class WindowAverageProcessor(taskContext : TaskContext, conf: UserConfig)
+  extends PersistentTask[AveragedValue](taskContext, conf) {
+
+  private val windowSize = conf.getLong(PersistentStateConfig.STATE_WINDOW_SIZE)
+    .getOrElse(throw new RuntimeException("state window size not configured"))
```
clockfly: We should never throw RuntimeException in actor.

Exception should also be avoided.

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/user/example/processor/WindowAverageProcessor.scala
```diff
@@ -0,0 +1,57 @@
...
+import org.slf4j.Logger
+
+object WindowAverageProcessor {
+  val LOG: Logger = LogUtil.getLogger(classOf[WindowAverageProcessor])
+}
+
+class WindowAverageProcessor(taskContext : TaskContext, conf: UserConfig)
+  extends PersistentTask[AveragedValue](taskContext, conf) {
+
+  private val windowSize = conf.getLong(PersistentStateConfig.STATE_WINDOW_SIZE)
```
clockfly: Maybe wrap windowSize and windowStep  in another objectt is better?

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/user/example/processor/WindowAverageProcessor.scala
clockfly: You can change the default constructor to make the dependencies explicit.

like this:
```
class WindowAverageProcessor(windowConfig: WindowConfig, taskContext : TaskContext, conf: UserConfig) {
  def this (taskContext: TaskContext, conf: UserConfig) = {
      this(conf.getLong(PersistentStateConfig.STATE_WINDOW_SIZE), context, conf)
  }
+  extends PersistentTask[AveragedValue](taskContext, conf) {

File: experiments/state/src/main/scala/org/apache/gearpump/streaming/state/user/example/processor/WindowAverageProcessor.scala [hunk 3]
```diff
@@ -0,0 +1,57 @@
...
+  override def onStart(startTime: StartTime): Unit = {
+    state.init(startTime.startTime)
+  }
+
+  override def onNext(message: Message): Unit = {
+    val value = AveragedValue(message.msg.asInstanceOf[String].toLong)
+    state.update(message.timestamp, value)
+  }
+
+  private[example] def getState: Option[AveragedValue] = {
```
clockfly: If getState is only used for UT, I suggest removing it.

val state is no much difference with  def getState

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/user/example/WindowAverage.scala
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.gearpump.cluster.main.{ArgumentsParser, CLIOption, ParseResult}
+import org.apache.gearpump.partitioner.HashPartitioner
+import org.apache.gearpump.streaming.kafka.lib.KafkaConfig
+import org.apache.gearpump.streaming.state.system.impl.PersistentStateConfig
+import org.apache.gearpump.streaming.state.user.example.processor.{GenNumProcessor, WindowAverageProcessor}
+import org.apache.gearpump.streaming.{Processor, StreamApplication}
+import org.apache.gearpump.util.Graph
+import org.apache.gearpump.util.Graph.Node
+
+object WindowAverage extends App with ArgumentsParser {
```
clockfly: How about making the name more explicit, like WindowAverageApp, so that people will not confuse it with internal impl class related with window logic.

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/user/api/PersistentTask.scala
```diff
@@ -0,0 +1,56 @@
...
+      .getOrElse(throw new RuntimeException("PersistentStateConfig not found"))
+  private val checkpointInterval = stateConfig.getCheckpointInterval
+  private val checkpointStore = stateConfig.getCheckpointStoreFactory.getCheckpointStore(conf, taskContext)
+  private val checkpointManager = new CheckpointManager(checkpointInterval, checkpointStore)
+
+  /**
+   *  this is for user to get PersistentState
+   *  user should define or import an implicit algebird monoid
+   */
+  protected def persistentState(implicit monoid: Monoid[T]): MonoidState[T] =
```
clockfly: I think the return type should be PersistentState, instead of MonidState.

As long as user implement PersistentState's interface, user should be allowed to define its own PersistentState task.

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/user/api/PersistentTask.scala
```diff
@@ -0,0 +1,56 @@
...
+    new DefaultState(monoid, checkpointManager, taskContext)
+
+  /**
+   * this is for user to get PersistentState with window functions
+   * user should define or import an implicit algebird group
+   */
+  protected def persistentState(windowSize: Long, windowStep: Long)(implicit group: Group[T]): WindowState[T] =
```
clockfly: I don't know whether it is a good idea to expose 
```persistentState(windowSize: Long, windowStep: Long)```
? Why not using 
```
def persistentState(implicit monoid: Monoid[T])
```
?

File: experiments/state/src/main/scala/org/apache/gearpump/streaming/state/user/api/PersistentTask.scala [hunk 7]
```diff
@@ -0,0 +1,56 @@
...
+  private val checkpointInterval = stateConfig.getCheckpointInterval
+  private val checkpointStore = stateConfig.getCheckpointStoreFactory.getCheckpointStore(conf, taskContext)
+  private val checkpointManager = new CheckpointManager(checkpointInterval, checkpointStore)
+
+  /**
+   *  this is for user to get PersistentState
+   *  user should define or import an implicit algebird monoid
+   */
+  protected def persistentState(implicit monoid: Monoid[T]): MonoidState[T] =
+    new DefaultState(monoid, checkpointManager, taskContext)
```
clockfly: I don't like the idea to define some default state here.
I think it is totally upto the developer about what state to use, the framework should make no assumption on this.

User can define its own persistentState or use framework provided persistentState.

And, user should also be able to combine multiple persstentState together to be a single persistentState.

Consider a use case, where use want to have three persistent states, 
1. a monid state of long
2. a monid state of CMS
3. a window state

How current design are going to support this? 

An alternative solution will be allow user to extend PersistentState, like this:
```
class MyState extends PersistentState {

private val longMonidState
private val cmsMonidState
private val windowState 

 ... 
other persistentState interface
}

```

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/system/impl/WindowState.scala
```diff
@@ -0,0 +1,145 @@
...
+  /**
+   * get Interval(startTime, endTime) for timestamp t
+   *    startTime = Math.max(lowerBound1, lowerBound2, checkpointTime)
+   *    endTime = Math.min(upperBound1, upperBound2, checkpointTime) where
+   *    lowerBound1 = step * Nmax1 <= t
+   *    lowerBound2 = step * Nmax2 + size <= t
+   *    upperBound1 = step * Nmin1 > t
+   *    upperBound2 = step * Nmin2 + size > t
+   */
+  private[impl] def getInterval(timestamp: TimeStamp, checkpointTime: TimeStamp): Interval = {
```
clockfly: Please make this function clearer.

what kind interval? purpose?

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/system/impl/WindowManager.scala
```diff
@@ -0,0 +1,46 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.gearpump.streaming.state.system.impl
+
+import org.apache.gearpump.TimeStamp
+
+class WindowManager(val windowSize: Long, val windowStep: Long) {
```
clockfly: PLease explain what this class is used for?

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/system/impl/DefaultState.scala
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.gearpump.streaming.state.system.api.MonoidState
+import org.apache.gearpump.streaming.task.TaskContext
+import org.apache.gearpump.util.LogUtil
+import org.slf4j.Logger
+
+object DefaultState {
+  val LOG: Logger = LogUtil.getLogger(classOf[DefaultState[_]])
+}
+
+class DefaultState[T](monoid: Monoid[T],
```
clockfly: Please make a better name, so that it is more easy to understand the difference with WindowState.

manuzhang:
thanks @clockfly. I've updated the pr addressing your comments and also modifying PersistentState and PersistentTask APIs.

I've been thinking really hard on this design and actually going back and forth several times for the past month in previous versions of this pr. 

Finally, this is what I feel most comfortable.

`PersistentTask` provides users with transaction support by the framework. If users have their own ways to transaction they are free to write their own transactional tasks. To use `PersistentTask`, users should pass in a `PersistentState`, which could either be the framework's built-in `UnboundedState` and `WindowState` or their own implementations.
`PersistentTask` hides `onStart` and `onNext` (where the framework recovers and checkpoints state in the background) from users but provides extra hooks for users to invoke `PersistentState` API. Currently, we have `processMessage` which is called in `onNext` and we could also add one for `onStart`. 

As to support combining states, I prefer your second solution. The framework only knows one `Persistent[T]` although that `T` could combine multiple user states. Only binary data will be transferred between user states and the framework. I think this keeps our design simple.

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/system/impl/UnboundedState.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  def apply[T](implicit monoid: Monoid[T]): UnboundedState[T] = {
+    new UnboundedState(monoid)
+  }
+}
+
+/**
+ * this is a MonoidState with no boundary
+ */
+class UnboundedState[T](monoid: Monoid[T]) extends MonoidState[T](monoid) {
```
clockfly: I prefer the name MonidState than UnboundedState

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/system/impl/WindowManager.scala
```diff
@@ -0,0 +1,54 @@
...
+ */
+package org.apache.gearpump.streaming.state.system.impl
+
+import org.apache.gearpump.TimeStamp
+
+/**
+ * this is used in window applications
+ * it keeps the current window and slide ahead when the window expires
+ */
+class WindowManager(val windowSize: Long, val windowStep: Long) {
```
clockfly: Maybe name Window is better instead of WindowManager.

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/system/impl/WindowManager.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  def slideTo(timestamp: TimeStamp): Unit = {
+    startTime = timestamp / windowStep * windowStep
+  }
+
+  def shouldSlide: Boolean = {
+    clock >= (startTime + windowSize)
+  }
+
+  def getCurrentWindow: (TimeStamp, TimeStamp) = {
```
clockfly: Maybe windowRange is better than getCurrentWindow

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/system/impl/CheckpointManager.scala
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.gearpump.streaming.state.system.impl
+
+import org.apache.gearpump.TimeStamp
+import org.apache.gearpump.streaming.state.system.api.CheckpointStore
+
+
+class CheckpointManager(checkpointInterval: Long,
```
clockfly: Maybe name CheckpointWorker or CheckpointHandler is better?

As this don't manage checkpoints.

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/system/impl/UnboundedState.scala
clockfly: Addition.

I don't like the word boundary, as I don't know what boundary means here, time boundary? I know you want to differentiate from WindowState, maybe NonWindowState better?

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/system/impl/WindowState.scala
```diff
@@ -0,0 +1,163 @@
...
+ * this is a list of states, each of which is bounded by a time window
+ * state of each window doesn't affect each other
+ *
+ * WindowState requires a Algebird Group to be passed in
+ * Group augments Monoid with a minus function which makes it
+ * possible to undo the update by messages that have left the window
+ */
+class WindowState[T](group: Group[T],
+                     taskContext: TaskContext,
+                     windowManager: WindowManager)
```
clockfly: It will make sense if rename windowManager to window.

Then the window concept perfectly match WindowState.

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/user/example/processor/GenNumProcessor.scala
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.gearpump.streaming.state.user.example.processor
+
+import org.apache.gearpump.Message
+import org.apache.gearpump.cluster.UserConfig
+import org.apache.gearpump.streaming.task.{StartTime, Task, TaskContext}
+
+class GenNumProcessor(taskContext : TaskContext, conf: UserConfig) extends Task(taskContext, conf) {
```
clockfly: Maybe name NumberGeneratorProcessor better?

clockfly:
Good work!

Only added a few small suggestions.

+1

manuzhang:
thanks @clockfly. updated against your comments.

Issue #974, opened 2015-05-28
[UT] lot of strange MasterProxy log in UT output


clockfly:

clockfly:
like this:
MasterProxy try to connect to closed actorsystem

[INFO] [05/28/2015 03:44:21.387] [MasterProxy@client220883686] sending identity to ActorSelection[Anchor(akka.tcp://master@127.0.0.1:54195/), Path(/user/master)]

[INFO] [05/28/2015 03:44:17.388] [MasterProxy@client220883686] sending identity to ActorSelection[Anchor(akka.tcp://master@127.0.0.1:54195/), Path(/user/master)]

PR #976, opened 2015-05-28
issue #974, Remove unnecessary MasterProxy and ClientContext creation in unit test

clockfly:


kkasravi:
+1. 
Question: Should MasterHarness and TestCluster be merged as a separate issue? - it seems like they do similar things. TestCluster is pretty much only used in services.

PR #982, opened 2015-06-02
fix #981, add prompt for ClassNotFoundException

### LINKED ISSUE #981: "gear app" doesn't prompt user when main class is not found
if I submit an app with incorrect main class name, only that class name is printed with no prompt what's going wrong.

manuzhang:


huafengw:
+1

PR #989, opened 2015-06-03
fix #902, delete kafka topic on close

### LINKED ISSUE #902: KafkaSource should provide an option to either delete a topic or to auto delete a topic on close
See #892

manuzhang:


external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/lib/KafkaUtil.scala
```diff
@@ -87,6 +87,17 @@ object KafkaUtil {
...
   }
 
+  def deleteTopic(zkClient: ZkClient, topic: String): Unit = {
+    try {
+      AdminUtils.deleteTopic(zkClient, topic)
+    } catch {
+      case e: Exception =>
+        LOG.error(e.getMessage)
+    } finally {
+      zkClient.close()
```
stanleyxu2005: Shouldn't `zkClient` be closed in `KafkaStorage.scala`?

manuzhang: Good point but I don't want `KafkaStorage` to worry about the zookeeper connection, whose creation and close should be contained in this method. Now I passed in a connect function so there will be no `ZkClient` objects in `KafkaStorage`. I also applied this to other methods that interact with zookeeper. The reason I don't just create a `ZkClient` here is to make the method testable.

clockfly:
+1

clockfly:
+1

PR #1009, opened 2015-06-09
fix #1008: nodes were not positioned expectedly in network graph; 

### LINKED ISSUE #1008: Nodes in network graph are not positioned expectedly
Nodes in network graph are not positioned expectedly. This is an issue introduced after visjs's option has been migrated to 4.x format (#826). And also the graph will not resize after the container is resized.

stanleyxu2005:
Also changed: 
1. Graph would not resize after container is resized
2. Updated vendor libraries


clockfly:
+1

Issue #952, opened 2015-05-26
scala.js design and implementation


kkasravi:
[Story Part 1](https://www.pivotaltracker.com/story/show/95438206)
[Story Part 2](https://www.pivotaltracker.com/story/show/95818676)
[Story Part 3](https://www.pivotaltracker.com/story/show/96348194)
[Story Part 4](https://www.pivotaltracker.com/story/show/96956046)
[Story Part 5](https://www.pivotaltracker.com/story/show/98293248)
[Story Part 6](https://www.pivotaltracker.com/story/show/98877146)
Provide design and implementation of integrating scala.js into gearpump services

Sub tasks:

- [ ] #1058 Add unit tests for org.apache.gearpump.dashboard.controllers.StreamingDag scalajs
- [ ] #1055 Get appdag.html working with data feature scalajs
- [ ] #1033 replace ngRoute and ng-breadcrumbs with scalajs UIRouter extension feature scalajs
- [ ] #1032 Add test framework to scalajs and some UT feature scalajs
- [ ] #1031 Fix compilation error on UT and resolve error for scalajs-angular_sjs0.6_2.11 feature scalajs UT
- [ ] #1023 Port appstatus.js feature scalajs
- [ ] #1016 Port tabset.js feature scalajs
- [ ] #1003 Port appdag.js, appmaster.js, appprocessor.js feature scalajs
- [ ] #1001 Get Master, Worker pages fully ported feature scalajs
- [ ] #998 	Standardize upickle serde on client feature scalajs
- [ ] #984 	Add additional dashboard definitions feature scalajs
- [ ] #979  Move common REST data to shared directory feature scalajs
- [ ] #975 Define modules using scalajs-angular feature scalajs
- [ ] #972 Define new project build structure feature scalajs
- [ ] #1172 Get appprocessor.html working with data 
- [ ] #1169 Remove unused directives, filters and add more UT 
- [ ] #1103 Get appmetrics.html working with data

kkasravi:
@mention #458

clockfly:
Sub tasks:

- [ ] #1058 Add unit tests for org.apache.gearpump.dashboard.controllers.StreamingDag scalajs
- [ ] #1055 Get appdag.html working with data feature scalajs
- [ ] #1033	replace ngRoute and ng-breadcrumbs with scalajs UIRouter extension feature scalajs
- [ ] #1032 Add test framework to scalajs and some UT feature scalajs
[...]

PR #1028, opened 2015-06-15
Issue #952

kkasravi:


kkasravi:
Will need to resubmit PR based on #952 and master

PR #1038, opened 2015-06-19
fix #1012, add DataSource and DataSink API

### LINKED ISSUE #1012: Gearpump Connector Framework
# [Story](https://www.pivotaltracker.com/story/show/96641634)
# Goal
The goal of connector is to make an easy-to-extend and easy-to-use Connector (DataSource and DataSink) system.

# How do Spark Streaming and Flink model this?
For Spark Streaming, [Input DStreamings](https://spark.apache.org/docs/latest/streaming-programming-guide.html#input-dstreams-and-receivers) and [Receiver](https://spark.apache.org/docs/latest/streaming-custom-receivers.html) for general Receiver API.

For Flink, [SourceFunction](http://ci.apache.org/projects/flink/flink-docs-release-0.8/streaming_guide.html#sources).

Both Spark Streaming and Flink model the connectors in a highly identical way. 

## DataSource
Basically, they both modeled in this way:

1. A DataSource (Spark named it as ```InputDStream``` or  ```Receiver``` and Flinked as ```SourceFunction```) generates a stream of type T (T is the decoded message type). So, it means decoding of message is the duty of DataSource. 
   1. DataSource has Start, Stop, Restart events to handle to ensure it is correct even when fault tolerant. 
2. Categorize common-used DataSources into two categories:
   1. Basic sources (not depending on another service), e.g. file system watcher, socket,  akka actor, collection, sequence. These sources have been builtin support in StreamingContext for creation. E.g. 
   ```  streamingContext.fileStream[KeyClass, ValueClass, InputFormatClass](dataDirectory)```
   2. Advanced sources (depending on other service), e.g. Kafka, Flume, .... these sources are provided via some util classes. E.g. 
   ```TwitterUtils.createStream(ssc, None)```

So, it means that for DataSource, the framework is DataSource agnostic. User specify DataSource via either predefined data sources (e.g. File system watche) or provided his own implementation of DataSource by implementing ```Receiver``` or ```SourceFunction```. The configuration of DataSource is not part of the application. E.g. for Kafka, the creation of data source in Spark Streaming is:

```
 import org.apache.spark.streaming.kafka._

 val kafkaStream = KafkaUtils.createStream(streamingContext, 
     [ZK quorum], [consumer group id], [per-topic number of Kafka partitions to consume])
```

## DataSink
For DataSink, Flink named as ```DataStreamSink``` or ```SinkFunction``` while Spark Streaming just named it as "Output Operations". But, basically, they are the same. The model is:

1. Provide a general mechanism for Sink. For Spark, it is ```foreachRDD``` function. For Flink, it is ```SinkFunction```. User implemented this function can do whatever output they want.
2. Provide builtin support for simple output, e.g. ```print```, ```saveAsTextFiles```, ```saveAsObjectFiles```, ```saveAsHadoopFiles```, ```writeAsCsv```. 

It is obvious that DataSink follows the same convention as DataSource.

## Package organization
As connectors (especially the connectors depending on external services, e.g. Kafka) are part of application instead of streaming framework, Spark put corresponding connectors under different artifact. E.g. kafka connector (both DataSource and DataSink) is in a standalone artifact ```spark-streaming-kafka_2.10```. 

Typically, each such artifact contains one DataSource and one DataSink. There are exceptions to this. E.g. Twitter connector only has DataSource. 

# Gearpump Design
We will follow the best practices from Spark Streaming and Flink if possible. 

### DataSource
We will provide a ```DataSource``` as the base class for data input stream. The signature looks like

``` abstract class DataSource[T](taskContext, userConf) extends Task(taskContext, userConf)```
```T``` is the message type.

And we will provide built-in implementation for common usages like, file system watch and Kafka.... 

The desired Kafka source shall be used in following way:

```
 import org.apache.gearpump.streaming.kafka._

 val KafkaStream = KafkaSource([ZK quorum], [consumer group id], [per-topic number of Kafka partitions to consume])
```

```kafkaSource``` is implemented in this way:

```
   object KafkaSource extends Processor[KafkaSourceTask] {
   		def apply([ZK quorum], [consumer group id], [per-topic number of Kafka partitions to consume]) {
   		    //put these arguments into taskConf
   		    ...
   		}
   }
```

### DataSink

We will provide a ```DataSink``` as the base class for data sink. The signature looks like

``` abstract class DataSink(taskContext, userConf) extends Task(taskContext, userConf)```

And we will provide built-in implementation for common usages like, file write and HBase.... 

The desired HBase sink shall be used in following way:

```
 import org.apache.gearpump.streaming.hbase._

 val hbaseSink = HBaseSink(hbaseConf)
```

```HBaseSink``` is implemented in this way:

```
   object HBaseSink extends Processor[HBaseSinkTask] {
   		def apply(hbaseConf) {
   		    //put hbaseConf into taskConf
   		    ...
   		}
   }
```

## High level API
TBD

### PR DESCRIPTION
**Pull request by manuzhang**
The API is a bit different from the design in #1012.

The major difference is that `DataSource` and `DataSink` are added as streaming interfaces rather than extending `Task`.
`KafkaSource` extends `TimeReplayableSource` which is a `DataSource` and `KafkaSink` extends `DataSink` respectively.

They are used in `KafkaSourceTask` and `KafkaSinkTask` which are provided as standard kafka connectors at task-level. 

`KafkaSourceProcessor` and `KafkaSinkProcessor` are utility objects to ease user's life of DAG setup and configuration.

Also, I've removed the Kafka RollingTopWords examples which lack maintenance and I'd like to implement with state api in the future.

### CODE REVIEW

File: streaming/src/main/scala/org/apache/gearpump/streaming/source/DataSource.scala [hunk 1]
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   *  read 'num' messages from source
+   *  Note: this is best effort. The returned
+   *  message count may be less than 'num'
+   *
+   *  @param num number of messages to read
+   */
+  def read(num: Int): List[Message]
+
+  def close(): Unit
```
whjiang: who will call this close() and read() and under which situation? Need better document?

streaming/src/main/scala/org/apache/gearpump/streaming/transaction/api/TimeReplayableSource.scala
```diff
@@ -40,24 +41,16 @@ import org.apache.gearpump.{Message, TimeStamp}
...
- *       TimeReplayableSource.pull(num).foreach { msg =>
+ *       TimeReplayableSource.read(num).foreach { msg =>
  *         TimeStampFilter.filter(msg, this.startTime).map(output)
  *       }
  *     }
  *   }
  *
  */
-trait TimeReplayableSource extends java.io.Serializable {
+trait TimeReplayableSource extends DataSource {
```
whjiang: Better document, hard to understand these 2 functions. 

If time starts from 0, and if I setStartTime(10) and then call startFromBeginning(), what data will I read, from 0 or from 10?

whjiang:
Hi Tianlun,

I will prefer following refinements:
1. It is better if you can provide some general SourceTask and SinkTask which accepts a DataSource and a DataSink as arguments.Or a ReplayableSourceTask accepts ReplayableSourceTask. What I prefer is to make user pipe creation as simple as possible. E.g. you mentioned that user can insert a timestamp filter to perform such filtering. My question is: can we standardize it in ReplayableSourceTask? Or make the common used pattern standardized? 
2. Please remove KafkaConfig and use similar parameters as Spark Streaming does. My understanding is that KafkaConfig is our invented config format which is not friendly to users.

manuzhang:
Hi Weihua,

I've updated the PR to address your comments. 

1. Users only need to extend `DataSource` and `DataSink`, which would be run by the general `DataSourceTask` and `DataSinkTask`. The source/sink is created when users set up a DAG and passed to the general `DataSourceProcessor` and `DataSinkProcessor` (please check out the `KafkaWordCount` example). 

2. Why is the `KafkaConfig` unfriendly to users ? Kafka has a long list of configs that are very tedious to be passed in as parameters. We also have our own configs required for `KafkaSource`. Since we've already implemented the `KafkaSource` for users, they only need to care about `kafka.conf`. SparkStreaming also requires users to pass in `kafkaParams` through a map (see `KafkaUtils.createDirectStream` when it uses low level kafka api.

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/StreamApp.scala
```diff
@@ -100,15 +100,13 @@ class TimeReplayableProducer[T:ClassTag](timeReplayableSource: TimeReplayableSou
 }
 
 class KafkaProducer[T:ClassTag](kafkaConfig: KafkaConfig, converter: Message => T) extends Traverse[T] {
-  val batchSize = kafkaConfig.getConsumerEmitBatchSize
-  val msgDecoder: MessageDecoder = kafkaConfig.getMessageDecoder
-  lazy val source = Some(new KafkaSource(kafkaConfig.getClientId, TaskId(0, 0), 1, kafkaConfig, msgDecoder)).map(kafkaSource => {
-    kafkaSource.startFromBeginning
+  lazy val source = Some(new KafkaSource(kafkaConfig)).map(kafkaSource => {
+    // WARN: the kafka source doesn't work without calling open
     kafkaSource
```
whjiang: I don't see the open() call. Is it a mistake?

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/lib/KafkaDecoder.scala
```diff
@@ -18,17 +18,11 @@
 
 package org.apache.gearpump.streaming.kafka.lib
 
-import com.twitter.bijection.Injection
 import org.apache.gearpump.Message
 import org.apache.gearpump.streaming.transaction.api.MessageDecoder
 
-import scala.util.{Failure, Success}
-
 class KafkaDecoder extends MessageDecoder {
```
whjiang: Better name as DefaultMessageDecoder as it is not specific to Kafka.

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/sink/KafkaSink.scala
```diff
@@ -0,0 +1,49 @@
...
+package org.apache.gearpump.streaming.kafka.sink
+
+import org.apache.gearpump.Message
+import org.apache.gearpump.streaming.kafka.lib.{KafkaConfig, KafkaUtil}
+import org.apache.gearpump.streaming.sink.DataSink
+import org.apache.gearpump.streaming.task.TaskContext
+import org.apache.kafka.clients.producer.{KafkaProducer, ProducerRecord}
+import org.apache.kafka.common.serialization.ByteArraySerializer
+
+class KafkaSink(getProducer: () => KafkaProducer[Array[Byte], Array[Byte]], topic: String) extends DataSink {
```
whjiang: Need better doc here. A little bit hard to understand the logic here.

streaming/src/main/scala/org/apache/gearpump/streaming/source/DataSource.scala
```diff
@@ -0,0 +1,66 @@
...
+   * @param startTime is the start time of system
+   */
+  def open(context: TaskContext, startTime: TimeStamp): Unit
+
+  /**
+   * read a number of messages from data source.
+   * invoked in each onNext() method of [[org.apache.gearpump.streaming.task.Task]]
+   * @return a list of messages wrapped in [[Message]]
+   */
+  def read(): List[Message]
```
whjiang: what is the maximum batch size?

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/StreamApp.scala
manuzhang: `open` requires `TaskContext` to be passed in which is not possible here.

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/StreamApp.scala
```diff
@@ -100,15 +100,13 @@ class TimeReplayableProducer[T:ClassTag](timeReplayableSource: TimeReplayableSou
 }
 
 class KafkaProducer[T:ClassTag](kafkaConfig: KafkaConfig, converter: Message => T) extends Traverse[T] {
```
whjiang: better move the KafkaProducer to kafka package instead of DSL package. Otherwise DSL package compilation will need kafka package.

streaming/src/main/scala/org/apache/gearpump/streaming/source/DataSource.scala
manuzhang: which maximum batch size ? `DataSource` has no notion of `batch size`. That is specific to `KafkaSource` which is the number of messages to read and send downstreams in one `onNext`

whjiang: read() return a list of messages. How to control the max size of the list?

streaming/src/main/scala/org/apache/gearpump/streaming/source/DataSource.scala
```diff
@@ -0,0 +1,66 @@
...
+ *      List(Message("message-1"), Message("message-2"), Message("message-3")
+ *    }
+ *
+ *    def close(): Unit = {}
+ *  }
+ * }}}
+ *
+ * subclass is required to be serializable
+ */
+trait DataSource extends java.io.Serializable {
```
whjiang: I will prefer DataSource[T] instead of DataSource. T is the user message type.

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/StreamApp.scala
manuzhang: The changes in this PR to DSL or pipeline modules are just to make them compilable with the new API. I'll open follow-up issues to address them

streaming/src/main/scala/org/apache/gearpump/streaming/source/DataSource.scala
manuzhang: that is left to api developers.

whjiang:
Hi Tianlun,

I am almost comfortable with the newest code except KafkaConfig. 

KafkaConfig is our invented config format for kafka. It mixed kafka config and our specific requirements like Decoder, timestamp etc. And also it has limitation under case like I need to read from two different kafka systems with different configurations, etc. 

My suggestion is:
1. Remove KafkaConfig from our API. Use the parameter list similar to Spark Streaming kafka connector.
2. You can keep KafkaConfig as a util class to help load/save such long parameter lists for user if you like. But, just a util class for user convenience. User is totally possible to bypass it.

streaming/src/main/scala/org/apache/gearpump/streaming/source/DataSource.scala
manuzhang: currently, we do not expose the type of message in the low-level api so I don't see the benefit of `DataSource[T]` here.

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/StreamApp.scala
manuzhang: DSL currently doesn't work well the existing low-level api so I add a note here and we should address this as follow-ups

streaming/src/main/scala/org/apache/gearpump/streaming/source/DataSource.scala
whjiang: I will still prefer to expose T as this may benefit the high level API and future low level API. Remember we have a proposal to make Message type invisible to user.

manuzhang: yes. I remember that so I'd like to defer the change till that comes in. Also, the high level API has its own `SourceTask`, which could (should ?) be merged with this source API in the future. The problem is we have so many flying pieces so I prefer to stick to the old way until we have a consistent to way to express (low-level api, high-level api, message type) throughout the system.

whjiang: Ok with your idea.

manuzhang:
Hi Weihua,

As I said, the Spark Streaming kafka connector that allows a parameter list to be passed in uses high level kafka consumer API rather than the low level kafka consumer API used by our `KafkaSource`. Besides, `KafkaSource` requires both kafka consumer and producer configs to be passed in since `KafkaSource` also stores the mapping of offset to timestamp to kafka. Plus, some configs will be passed in a long code path. (`KafkaSource` -> `FetchThread` -> `KafkaConsumer`) .

> And also it has limitation under case like I need to read from two different kafka systems with different configurations, etc.

we don't keep users from loading different kafka configurations as long as they are from different conf files like `kafka1.conf` and `kafka2.conf`.

I agree that forcing user to use `KafkaConfig` is not nice but I don't see how it could be easier for users the other way around with a few code changes.

manuzhang:
@kkasravi the API change would affect Pipeline and PipelineDSL. I only make them compile in this PR and they won't work if the PR is merged in. I'd like to defer the fix to follow-ups since this PR is already quite big.  Is that OK for you ?

clockfly:
@manuzhang 

I am not OK that any new commit break existing code. Please fix Pipeline with new updates.

manuzhang:
@kkasravi @clockfly 

The PipelineDSL is working as it was now.  The UT failure is unrelated.

streaming/src/main/scala/org/apache/gearpump/streaming/source/DataSourceProcessor.scala
```diff
@@ -0,0 +1,33 @@
...
+
+object DataSourceProcessor {
+  def apply(dataSource: DataSource,
+            parallelism: Int,
+            description: String = "",
+            taskConf: UserConfig = UserConfig.empty)(implicit system: ActorSystem): Processor[DataSourceTask] = {
+    Processor[DataSourceTask](parallelism, description = description,
+      taskConf.withValue[DataSource](DataSourceTask.DATA_SOURCE, dataSource))
+  }
+}
```
kkasravi: Maybe this should be something like
```
object DataSourceProcessor {
  def apply[T<:DataSource,U<:DataSourceTask](...)

streaming/src/main/scala/org/apache/gearpump/streaming/source/DataSource.scala
kkasravi: T would actually be Message[T] however this is a very difficult change having tried it several time and affects many files. See https://github.com/intel-hadoop/gearpump/issues/57 for some context.

kkasravi: @manuzhang  SourceTask, SinkTask, SinkConsumer, TraverseType and Traverse types have a number of similarities and I do think these types as well as TraversableSource and TraversableSink should be merged into this API's. Right now using the high level API it's possible to never expose Message which I think is a good objective since Message is an internal transport wrapper. Also SourceTask is compatible with scala collections which is another nice objective.

streaming/src/main/scala/org/apache/gearpump/streaming/source/DataSourceProcessor.scala
manuzhang: Hi Kam.

1. We will serialize `DataSource` object into `UserConfig` and deserialize it back in a running  `DataSourceTask`  here.
2. We have a general `DataSourceTask` that accepts any `DataSource` so users don't need to write their own.

streaming/src/main/scala/org/apache/gearpump/streaming/source/DataSource.scala
manuzhang: Yes. I totally agree. The only problem is that many are experimental and we can't resolve all the issues in one shot.

manuzhang:
Guys, thank you all for the good comments.
As this is a big PR and there are still concerns with the KafkaConfig, I'd like to split it into several PRs. I opened #1049 to add the source and sink API only. Closed this for now.

PR #1065, opened 2015-06-29
fix #1064: edge data was not constructed correctly

### LINKED ISSUE #1064: AppDag page breaks when edge does not have data stream
`vis.js` will raise an issue

stanleyxu2005:


clockfly:
+1

PR #1068, opened 2015-06-29
Fix #1067. Changing visibility of yarn resources

### LINKED ISSUE #1067: Visibility of YARN resources should be set to APPLICATION instead of PUBLIC for better isolation between apps and saving cluster space.
[Story](https://www.pivotaltracker.com/story/show/98065308)
From the Hortonworks [blog](http://hortonworks.com/blog/management-of-application-dependencies-in-yarn/) : 
"All the LocalResources (remote URLs) that are marked PUBLIC are accessible for containers of any user. Typically PUBLIC resources are those that can be accessed by anyone on the remote file-system and, following the same ACLs, are copied into public LocalCache. If in future, a container belonging to this or any other application (of this or any user) requests the same LocalResource, it is served from the LocalCache and thus not copied/downloaded again if it isn’t evicted from the LocalCache by then. All files in public cache will be owned by “yarn-user” (user which NodeManager runs as) with world-readable permissions, so that they can be shared by containers from all users whose containers are running on that node."

Moreover, in case of submitting subsequent versions of the same app, previous versions stay in cache and occupy space. This is not the case when using APPLICATION visibility.

TomaszT:


clockfly:
+1

PR #1075, opened 2015-06-30
part of #1017, split the lib directory

huafengw:


core/src/main/scala/org/apache/gearpump/jarstore/JarFileContainerWrapper.scala
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ */
+package org.apache.gearpump.jarstore
+
+import java.io._
+
+import org.apache.gearpump.cluster.UserConfig
+import JarFileContainerWrapper._
+
+class JarFileContainerWrapper private (conf: UserConfig) extends JarFileContainer{
```
manuzhang: maybe some comments here on why we need this "wrapper"

core/src/main/scala/org/apache/gearpump/jarstore/JarFileContainerWrapper.scala
```diff
@@ -0,0 +1,71 @@
...
+ * limitations under the License.
+ */
+package org.apache.gearpump.jarstore
+
+import java.io._
+
+import org.apache.gearpump.cluster.UserConfig
+import JarFileContainerWrapper._
+
+/**
```
clockfly: I think the JarFileContainerWrapper itself is clear. No need to add class comments in the class definition of JarFileContainerWrapper.

We only need to explain why we use JarFileContainerWrapper at where it is used. Like JarStoreService.

clockfly:
+1

clockfly:
@huafengw 

Please be 100% sure that this PR won't break existing function before merging.

PR #1077, opened 2015-06-30
fix #1076 make core dependent on hadoop-common

### LINKED ISSUE #1076: Gearpump core should dependent on hadoop-common
We should provide an explicit way to set hadoop configuration so the core has to depend on hadoop-common

huafengw:


clockfly:
-1

Different version of hadoop common are not binary compatible with each other. That will introduce conflict.

whjiang:
@clockfly we definitely need a way to let user provides their own hadoop version. But, from my point of view, making gearpump depending on hadoop common is inevitable.

clockfly:
@whjiang 

I don't think so. Gearpump can run with zero dependency on external service besides akka.

rvvincelli:
@clockfly @whjiang Hello, I am sorry to post on this closed issue, but I've encountered a problem about this dependency in my Gearpump toy. Without `hadoop-common` my build would fail with the following: `missing or invalid dependency detected while loading class file 'HBaseSink.class'`.

In this moment I don't have time to put together a minimal non-working configuration but here's the link to the buildfile: https://github.com/rvvincelli/gearpump-wordcount-kafka-hbase/blob/master/build.sbt.

PR #1080, opened 2015-06-30
Fix #1037 recreate log files when deleted in runtime

### LINKED ISSUE #1037: remove logs when cluster is running,but program won't create again
I remove gearpump master,ui or worker log in logs directory, but program would not do as kafka logs, it won't create the log file again and append

skw1992:


core/src/main/java/org/apache/gearpump/util/RecreateFileAppender.java
```diff
@@ -0,0 +1,26 @@
+package org.apache.gearpump.util;
```
manuzhang: please add license header

core/src/main/java/org/apache/gearpump/util/RecreateFileAppender.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.apache.log4j.RollingFileAppender;
+import org.apache.log4j.spi.LoggingEvent;
+
+import java.io.File;
+
+/**
+ * Created by keweisun on 6/29/2015.
+ */
```
huafengw: Remove these comment

PR #1090, opened 2015-07-01
fix #1066: AppSubmitter will return 1 when error is occured

### LINKED ISSUE #1066: AppSubmitter exits with status 0 on failure, so it's hard to detect its failure in bash script.
[Story](https://www.pivotaltracker.com/n/projects/1345200/stories/98063734)
It should use something like System.exit(-1).

stanleyxu2005:


clockfly:
+1

PR #1092, opened 2015-07-01
fix # 1053, allow user to config Kafka through API

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1092)


manuzhang:
This PR splits kafka's own producer config (`KafkaProducerConfig`) and consumer config (`KafkaConsumerConfig`) from gearpump's config for kafka. They all can be set through both API and config file on the classpath.

`KafkaProducerConfig` and `KafkaConsumerConfig` are wrapped in `KafkaConfig` along with gearpump's config for now. They could be used individually when refactoring kafka connector in follow-up PRs.

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/lib/PropertiesConfig.scala
```diff
@@ -16,29 +16,22 @@
...
-  def update(r: (T, Long)): Unit = {
-    rankings :+= r
-  }
-  def update(obj: T, count: Long): Unit = {
-    update((obj, count))
-  }
-
-  def getTopN(n: Int): List[(T, Long)] = {
-    rankings.sortBy(_._2).reverse.take(n)
+abstract class PropertiesConfig(properties: Properties) extends java.io.Serializable {
```
whjiang: why this PropertiesConfig class is needed? Seems just a simple wrapper for Properties.

whjiang:
Several questions:
1. What's relationship between KafkaConfig and KafkaConsumerConfig/KafkaProducerConfig?
2. Can we discard KafkaConfig at all and only use KafkaConsumerConfig and KafkaProducerConfig?
3. What's additional value added by KafkaConsumerConfig/KafkaProducerConfig compared to java.util.Properties? If KafkaConsumerConfig/KafkaProducerConfig is just a list of properties, why not directly use java.util.Properties instead?

To me, the current config relationship is still hard to understand.

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/lib/PropertiesConfig.scala
manuzhang: We need to set default values for users if no `.properties` file is provided.

whjiang: If what you want is just the function ```putIfAbsent```, you can either put in an util class or inline the code at all.

manuzhang:
1.There is no relationship between them. `KafkaConsumerConfig` and `KafkaProducerConfig` needs to be passed in through `KafkaConfig` without breaking interfaces in this PR. They would be split out in refactoring kafka connectors.
2. We cannot discard `KafkaConfig` at all. `KafkaConfig` is intended for **gearpump config for kafka** while `KafkaConsumerConfig` an `KafkaProducerConfig` are kafka's own configs.
3. `KafkaConsumerConfig` and `KafkaProducerConfig` set default values for required kafka configurations. 

kafka installation has `consumer.properties` and `producer.properties` (with default values for required configurations) under its `conf` which users could use directly and be loaded into `KafkaConsumerConfig` and `KafkaProducerConfig`. If the properties files are not found, we set the default values for users. 

They are hard to understand but would be much clearer when we refactor the kafka connector.

whjiang:
@manuzhang please draw some explanation about what your ultimate kafka config will look like, the relation of each and how to use them in some psuedo code?

manuzhang:
They will be three kafka configs.

* `GearpumpKafkaConfig` - either loaded from a `kafka.conf`  or built through API.
* `KafkaConsumerConfig` - either loaded from a `consumer.properties` or built through API.
* `KafkaProducerConfig` - either loaded from a `producer.properties` or built through API.

For `KafkaSink`, there is no need to use `GearpumpKafkaConfig` and `KafkaConsumerConfig`.

```scala
    val producerConfig = new KafkaProducerConfig("producer.properties")
    producerConfig.put("compression.type", "gzip")
    val kafkaSink = new KafkaSink(producerConfig)
```

It would be more complex for `KafkaSource` which involves all three configs. 

```scala
   val gearpumpConfig = GearpumpKafkaConfig("kafka.conf").withConsumerEmitBatchSize(100)
   val producerConfig = new KafkaProducerConfig("producer.properties")
   val consumerConfig = new KafkaConsumerConfig() // we set default values for users

   val kafkaSource = new KafkaSource(gearpumpConfig, producerConfig, consumerConfig)
```

whjiang:
@manuzhang several comments:
1. better rename ```KafkaConsumerConfig``` to ```KafkaSourceConfig``` and ```KafkaProducerConfig``` to ```KafkaSinkConfig``` as they are used with KafkaSource and KafkaSink in Gearpump.
2. for sink, ```KakfaProducerConfig``` seems OK to me.
3. for source, why we need producer config here? For offset management? If so, I suggest you to put offset into another central place. And why we need ```GearpumpKafkaConfig```? Can we ONLY use ```KafkaConsumerConfig``` for data source?

manuzhang:
1. they are not bound to `KafkaSource` and `KafkaSink` and also used in `KafkaCheckpointStore` for example.
2. `KafkaOffsetManager` is coupled with `KafkaSource` now. Splitting the two is no trivial work which I'd like to defer as future optimizations.
3. For kafka source, we also have configs other than those coming from kafka "consumer.properties". For example, we will stop pulling messages from kafka and sleep for a while when the number of in-memory messages hit a threshold or there are no more incoming messages. These configs will be loaded from `kafka.conf` which users could tune without recompiling source codes.

whjiang:
@manuzhang so the consumer.properties and producer.properties are the standard Kafka configuration defined by Kafka?

whjiang:
@manuzhang 

1. I am OK to postpone the decouple of offset manager from Kafka Source. But, this may hurt other data sources as all replayable data source needs offset manager. So, I am afraid we will need to do it quite quickly after this kafka restruction.
2. I think we shall remove ```GearpumpKafkaConfig```. My understanding is that it provides:
a. the topic to fetch.
b. KafkaSource specific parameters to tune.
For a, I think we can use an explicit parameter in KafkaSource construction to pass the topic. 
For b, I think we can use either UserConfig or KafkaConsumerConfig or a map to handle it. E.g. if we use KafkaConsumerConfig, we can tell user that to further tuning Kafka source, you can put parameter xxx in consumer.properties, etc. Currently, we already have enough config (UserConfig, KafkaConsumerConfig...), I don't want to make user even more confused.

manuzhang:
@whjiang 

1. Yes. `consumer.properties` and `producer.properties` are kafka standards.
2. I will work hard to decouple the two immediately after kafka refactoring but all the changes will not come in one pr. 
3. As for `GearpumpKafkaConfig`. 
    a. I don't want to mix in `consumer.properties` with gearpump configs
    b. I could try passing parameters to constructors but not in this PR where `KafkaSource` is not yet at client side.
    c. there should be a conf file  such that users could tune the parameters after deployment.

whjiang:
@manuzhang 

My suggestion is to ensure:
1. User will not be confused by so many config types.
2. User doesn't need to manipulate a new unfamiliar config file (kafka.conf) to get the simple case work. Parameter tuning is an advanced feature that is NOT needed for new users. I want to ensure that we can help new users easily create a kafka pipeline.

manuzhang:
@whjiang 

These are good suggestions but I don't see them as blocking issues for this PR, which is to split the kafka specif configs and allow user to config kafka through API

examples/streaming/kafka/src/main/scala/org/apache/gearpump/streaming/examples/kafka/wordcount/KafkaWordCount.scala
```diff
@@ -46,7 +47,10 @@ object KafkaWordCount extends App with ArgumentsParser {
     val sumNum = config.getInt("sum")
     val kafkaStreamProcessorNum = config.getInt("kafka_stream_processor")
 
-    val kafkaConfig = KafkaConfig(ConfigFactory.parseResources("kafka.conf"))
+    val kafkaConfig = new KafkaConfig(
```
whjiang: confused. What is KafkaConfig? kafka.conf or the combination of kafka.conf, consumer.properties and producer.properties?

manuzhang: combination of the three to conform to existing interfaces.

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/lib/KafkaConfig.scala
```diff
@@ -30,42 +33,29 @@ object KafkaConfig {
...
+
+  def apply(resource: String): KafkaConfig = KafkaConfig(ConfigFactory.parseResources(resource))
+
+  def apply(): KafkaConfig = KafkaConfig(ConfigFactory.empty())
 
   private val LOG: Logger = LogUtil.getLogger(getClass)
 }
 
-class KafkaConfig(config: Config) extends Serializable  {
+class KafkaConfig(gearpumpConfig: Config,
```
whjiang: why kafka.conf is named as gearpumpConfig?

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/KafkaSource.scala
```diff
@@ -41,16 +44,38 @@ class KafkaSource private[kafka](fetchThread: FetchThread,
...
+        tp -> KafkaOffsetManager(appName, tp, kafkaConfig, consumerConfig, producerConfig)).toMap)
+
+  private[kafka] def this(appName: String,
+                          grouper: KafkaGrouper,
+                          messageDecoder: MessageDecoder,
+                          kafkaConfig: KafkaConfig,
+                          consumerConfig: ConsumerConfig,
+                          producerConfig: KafkaProducerConfig) =
+    this(appName, KafkaUtil.getTopicAndPartitions(KafkaUtil.connectZookeeper(consumerConfig)(),
+      grouper, kafkaConfig.getConsumerTopics), messageDecoder, kafkaConfig, consumerConfig, producerConfig)
```
whjiang: what the use case for this constructor?

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/KafkaSource.scala
```diff
@@ -41,16 +44,38 @@ class KafkaSource private[kafka](fetchThread: FetchThread,
...
+           messageDecoder: MessageDecoder,
+           kafkaConfig: KafkaConfig,
+           consumerConfig: KafkaConsumerConfig,
+           producerConfig: KafkaProducerConfig) =
+    this(appName, kafkaConfig.getGrouperFactory.getKafkaGrouper(taskId, taskParallelism),
+      messageDecoder, kafkaConfig, consumerConfig.config, producerConfig)
+
+  def this(appName: String, taskId: TaskId, taskParallelism: Int, kafkaConfig: KafkaConfig, messageDecoder: MessageDecoder) = {
+    this(appName, taskId, taskParallelism, messageDecoder, kafkaConfig, kafkaConfig.consumerConfig, kafkaConfig.producerConfig)
+  }
```
whjiang: What is the use case for this one? Sorry for these stupid questions.

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/lib/KafkaConfig.scala
manuzhang: as **gearpump config for kafka**

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/KafkaSource.scala
manuzhang: this is the public interface to "new" a `KafkaSource` while other constructors are hidden and used in `kafka` package.

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/lib/KafkaConfig.scala
whjiang: if so, please add kafka as prefix or postfix. GearpumpConfig is a too general term.

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/KafkaSource.scala
whjiang: So do you mean we must use kafKaconfig to select topic and partition?

And whether this interface is compatible with DataSourceProcessor or not? For example, at client side, how do I know my taskId?

manuzhang: 1. topic for kafka consumer is created beforehand by user. taskId is not part of it. 
2. we don't use KafkaConfig to select topic and partition but `KafkaGrouper`.
3. This is not compatible with DataSourceProcessor and needs refactoring.

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/lib/KafkaConfig.scala
manuzhang: it's only internal (private) to `KafkaConfig`. Given the context I think the name is clear enough to tell it part from consumer config and producer config.

manuzhang:
@whjiang 
I've created #1098 as a parent task which would finally deal with all the concerns in this PR, which is only the first step with some confusing parts and temporary work-arounds.

clockfly:
@manuzhang 

If that is temporary work-arounds, that means some code will be removed eventually?
In that case, I suggest to submit a combined version which is more easy to review.

PR #1095, opened 2015-07-02
fix ui issues #1094 #1096 and support dag change detect #1097

### LINKED ISSUE #1094: "Service unreachable" message should only be shown when health check fails
Some applicaton related api calls will return 500, which does not mean the server is unreachable. So we should only inform user when health check fails.

stanleyxu2005:
see the changes below


clockfly:
+1

PR #1102, opened 2015-07-02
Fix #1037 recreate logs when deleted in runtime

### LINKED ISSUE #1037: remove logs when cluster is running,but program won't create again

skw1992:


clockfly:
Please DON'T use gear.conf for configuration. 
Please use log4j.properties instead.


For example,
when you set in log4j.properties.
``` 
log4j.appender.RollingFileAppender.MyInterval=100
```
then appender.setMyInterval will be called.

core/src/main/java/org/apache/gearpump/util/RecreateRollingFileAppender.java
```diff
@@ -0,0 +1,61 @@
...
+
+public class RecreateRollingFileAppender extends RollingFileAppender {
+
+    {
+        Config config = ClusterConfig.load().master();
+        int period = config.getInt(Constants.GEARPUMP_LOG_FILEEXISTINTERVAL());
+        initSchedule(period);
+    }
+
+    private void initSchedule(int period) {
```
clockfly: Please DON'T create another thread in appender.

clockfly:
+1

PR #1108, opened 2015-07-03
fix #1106: cache opensource webfont lato (831kb in total) locally

### LINKED ISSUE #1106: Cache webfont locally
It might be troublesome to load particular fonts from external network. So we'd cache opensource font(s) locally.

stanleyxu2005:


clockfly:
+1

PR #1114, opened 2015-07-06
issue 1113, add a hbase datasource/sink

whjiang:
This is not a full functional PR. Just for review purpose.

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1114)

### DIFF
--- a/external/hbase/src/main/scala/org/apache/gearpump/external/hbase/HBaseDataSink.scala
+++ b/external/hbase/src/main/scala/org/apache/gearpump/external/hbase/HBaseDataSink.scala
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


experiments/hbase/src/main/scala/org/apache/gearpump/experiments/hbase/HBaseDataSink.scala
```diff
@@ -0,0 +1,64 @@
...
+ */
+class HBaseDataSink(hbaseConf : Configuration, tableName : String) extends DataSink{
+  var table : Option[HTable] = None
+  /**
+   * open connection to data sink
+   * invoked at onStart() method of [[org.apache.gearpump.streaming.task.Task]]
+   * @param context is the task context at runtime
+   */
+  def open(context: TaskContext): Unit = {
+    table = Some(new HTable(hbaseConf, tableName))
```
stanleyxu2005: As `hbaseConf` and `tableName` are used to create `HTable` instance. Do you think pass in a `HTable` instance will be more simpler?

whjiang: I prefer hbaseConf+tableName as it can provide more info. E.g. I need to create ```SCAN``` object. But for this object, we can tune the parameters for it. using a ```hbaseConf``` allows user to do so.

manuzhang: could add a private constructor that has `HTable` and `Scan` as parameters to ease unit test

Issue #1098, opened 2015-07-02
refactoring kafka connector to conform to DataSource and DataSink API


manuzhang:
Task list:

- [x] allow user to configure kafka through API (#1053)
- [x] refactoring KafkaSource and KafkaSink to conform to DataSource and DataSink API (#1115)
- [x] decoupling KafkaStorage from KafkaSource (#438)
- [ ] add documentations for DataSource and DataSink and update kafka docs on README and gearpump.io

PR #1116, opened 2015-07-06
fix #1115, refactor kafka module

### LINKED ISSUE #1115: refactor kafka module

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1116)


experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/StreamApp.scala
```diff
@@ -91,7 +88,7 @@ object StreamApp {
 class TimeReplayableProducer[T:ClassTag](timeReplayableSource: TimeReplayableSource, converter: Message => T, batchSize: Int) extends Traverse[T] {
   val source = timeReplayableSource
   override def foreach[U](fun: T => U): Unit = {
-    val list = source.pull(batchSize)
+    val list = source.read()
```
whjiang: I think it is better let caller instead of callee of ```read()``` to decide the max batch size as caller, in theory, has better understanding of the pressure it has.

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/StreamApp.scala
```diff
@@ -100,15 +97,30 @@ class TimeReplayableProducer[T:ClassTag](timeReplayableSource: TimeReplayableSou
...
+      override def scheduleOnce(initialDelay: FiniteDuration)(f: => Unit): Cancellable = null
+      override def self: ActorRef = null
+      override def output(msg: Message): Unit = {}
+      override def upstreamMinClock: TimeStamp = Message.noTimeStamp
+      override def actorOf(props: Props): ActorRef = null
+      override def actorOf(props: Props, name: String): ActorRef = null
+      override def sender: ActorRef = null
+      override def appMaster: ActorRef = null
+      override def system: ActorSystem = null
+    }
```
whjiang: why do we need to create a TaskContext here? Sorry for this dummy question. But, I don't understand why we need to create TaskContext at client side.

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/system/impl/KafkaCheckpointStore.scala
```diff
@@ -34,10 +38,13 @@ import scala.util.Try
...
   private val LOG: Logger = LogUtil.getLogger(classOf[KafkaCheckpointStore])
 
-  def apply(config: KafkaConfig, topic: String, topicExists: Boolean) = {
-    val getConsumer = () => KafkaConsumer(topic, 0, config)
-    val producer = new KafkaProducer[Array[Byte], Array[Byte]](
-      KafkaUtil.buildProducerConfig(config), new ByteArraySerializer, new ByteArraySerializer)
+  val CONSUMER_CONFIG = "consumer_config"
+  val PRODUCER_CONFIG = "producer_config"
+
+  def apply(topic: String, topicExists: Boolean, consumerConfig: ConsumerConfig, producerConfig: Properties) = {
```
whjiang: why for consumer config, using ```kafka.consumer.ConsumerConfig``` while for producer config, using ```Properties```? This is inconsistent.

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/KafkaSink.scala
```diff
@@ -20,27 +20,43 @@ package org.apache.gearpump.streaming.kafka
 
 import java.util.Properties
 
-import org.apache.gearpump.streaming.kafka.lib.{KafkaUtil, KafkaConfig}
-import org.apache.kafka.clients.producer.{ProducerRecord, KafkaProducer}
+import org.apache.gearpump.Message
+import org.apache.gearpump.streaming.kafka.lib.{KafkaConfig, KafkaUtil}
+import org.apache.gearpump.streaming.sink.DataSink
+import org.apache.gearpump.streaming.task.TaskContext
+import org.apache.kafka.clients.producer.{KafkaProducer, ProducerRecord}
```
whjiang: looks weird for mixing of ```kafka.consumer.xxx``` and ```org.apache.kafka.clients.producer.xxx```. Is this a problem of Kafka itself?

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/KafkaSource.scala
```diff
@@ -36,62 +37,105 @@ object KafkaSource {
...
 
-  private[kafka] def this(appName: String, config: KafkaConfig, topicAndPartitions: Array[TopicAndPartition], messageDecoder: MessageDecoder) =
-    this(FetchThread(topicAndPartitions, config), messageDecoder,
-      topicAndPartitions.map(tp => tp -> KafkaOffsetManager(appName, config, tp)).toMap)
-
-  private[kafka] def this(appName: String, config: KafkaConfig, grouper: KafkaGrouper, messageDecoder: MessageDecoder) =
-    this(appName, config, KafkaUtil.getTopicAndPartitions(KafkaUtil.connectZookeeper(config)(),
-      grouper, config.getConsumerTopics), messageDecoder)
+  private lazy val messageDecoder = config.getMessageDecoder
+  private lazy val timestampFilter = config.getTimeStampFilter
```
whjiang: Is timestamp filter specific to Kafka source? I assume it is general to all ```TimeReplayableSource```.

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/KafkaSource.scala
```diff
@@ -36,62 +37,105 @@ object KafkaSource {
...
       }
+      fetch.setDaemon(true)
+      fetch.start()
+    }
+  }
+
+  override def open(context: TaskContext, startTime: Option[TimeStamp]): Unit = {
+    import context.{appName, parallelism, taskId}
+
+    val grouper = config.getGrouperFactory.getKafkaGrouper(taskId, parallelism)
```
whjiang: Must this grouper factory got from KafkaConfig instead of UserConfig?

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/KafkaSource.scala
```diff
@@ -36,62 +37,105 @@ object KafkaSource {
...
+    val consumerConfig = new ConsumerConfig(config.consumerConfig)
+    val topicAndPartitions = KafkaUtil.getTopicAndPartitions(KafkaUtil.connectZookeeper(consumerConfig)(),
+      grouper, config.getConsumerTopics)
+    this.fetchThread = Some(FetchThread(topicAndPartitions, config.getFetchThreshold,
+      config.getFetchSleepMS, consumerConfig))
+    this.offsetManagers = topicAndPartitions.map { tp =>
```
whjiang: Seems offset manager is still coupled with KafkaSource. Do you plan to separate them in following PR?

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/lib/consumer/KafkaMessage.scala
```diff
@@ -16,10 +16,17 @@
...
 import kafka.common.TopicAndPartition
 
+/**
+ * wrapper over messages from kafka
+ * @param topicAndPartition where message comes from
+ * @param offset message offset on kafka queue
+ * @param key message key, could be None
+ * @param msg message payload
+ */
 case class KafkaMessage(topicAndPartition: TopicAndPartition, offset: Long,
```
whjiang: Do we really need ```topicAndPartition``` and ```offset``` in the message body? Any use case?

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/KafkaSource.scala
manuzhang: yes, that's third subtask of #1098 while this PR fixes the first two.

whjiang:
Generally this PR is good. 

My two main concerns are:
1. I think we shall remove **KafkaConfig** concept at all. 
2. We shall separate OffsetManager concept from Kafka source/sink. Of course, this can be a separate   PR.

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/KafkaSource.scala
manuzhang: `KafkaConfig` is to make kafka related configs easy to use. `kafkaConfig.getGrouperFactory` instead of `userConfig.get("kafka.grouper.factory.class").getOrElse(classOf[KafkaDefaultGrouperFactory].getName)`

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/KafkaSource.scala
manuzhang: it's general but not compulsory to use.

experiments/state/src/main/scala/org/apache/gearpump/streaming/state/system/impl/KafkaCheckpointStore.scala
manuzhang: Kafka has two sets of producers and consumers with different looking of interfaces. We are using the new kafka producer and old kafka consumer (because new kafka consumer is not ready).

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/StreamApp.scala
manuzhang: to feed `source.open`. The `KafkaSource` here would be invoked by `SourceTask` under DSL which doesn't call `source.open` for us.

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/lib/consumer/KafkaMessage.scala
manuzhang: `KafkaMessage` is not transferred down streams but for `KafkaSource`. `topicAndPartition` would dispatch message to its corresponding `KafkaOffsetManager` and `offset` is stored along with the timestamp for replay.

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/StreamApp.scala
manuzhang: the caller is the general `DataSourceTask` which has no idea of what kind of source it is dealing with

manuzhang:
@whjiang 

1. we already have a long discussion over `KafkaConfig` and I've tried to remove it but find it impossible to do.  here are my arguments
  
   * there are three kafka configs required now, kafka consumer config, kafka producer config and gearpump config for kafka. Ideally, consumer config is for `KafkaSource and producer config `KafkaSink`, and both of them could be loaded and set at client side. Nonetheless, `KafkaStorage` is created at server side and requires both producer and consumer configs. Hence, we still need to pass the configs in `UserConfig` from client to server. 
  * gearpump config for kafka should not be removed. As a kafka user, for instance, I may not know the kafka topic name to consume in advance. With the topic configuration in `kafka.conf`, I can easily reset the topic name without recompiling gearpump. Removing `KafkaConfig` and configuring all these values through API would forbid this and would be a pain to test gearpump kafka module for users. 
  * Instead of separate kafka config classes, `KafkaConfig` is a single facade for all kafka related configs in gearpump that allows users to either configure through API or files with great flexibility. By default, it loads `kafka.conf`, `consumer.properties` and `producer.properties` on the classpath, help to set required values and users are good to go with `new KafkaConfig()`. Users could also load their customized config files. More config classes only produce more confusions.
  * `KafkaConfig` makes it easier to use gearpump configs for kafka as in the `KafkaGrouperFactory` example above. 

2. yes, it will be done in follow-up PRs.

whjiang:
@manuzhang let's have some F2F discussion. Please reach me when you are available.

manuzhang:
@whjiang 
further refactoring applying your suggestions.  Please review.

examples/streaming/kafka/src/main/scala/org/apache/gearpump/streaming/examples/kafka/wordcount/KafkaWordCount.scala
```diff
@@ -34,27 +34,29 @@ object KafkaWordCount extends App with ArgumentsParser {
...
+    val sourceNum = config.getInt("source")
     val splitNum = config.getInt("split")
     val sumNum = config.getInt("sum")
-    val kafkaStreamProcessorNum = config.getInt("kafka_stream_processor")
+    val sinkNum = config.getInt("sink")
 
-    val kafkaConfig = KafkaConfig(ConfigFactory.parseResources("kafka.conf"))
     val appConfig = UserConfig.empty
-      .withValue(KafkaConfig.NAME, kafkaConfig)
+      .withInt(DataSourceConfig.SOURCE_READ_BATCH_SIZE, 1000)
```
whjiang: Can we add some default value for this? If user doesn't set it, by default it is 1000 (e.g.).

examples/streaming/kafka/src/main/scala/org/apache/gearpump/streaming/examples/kafka/wordcount/KafkaWordCount.scala
```diff
@@ -34,27 +34,29 @@ object KafkaWordCount extends App with ArgumentsParser {
...
 
-    val kafkaStreamProducer = Processor[KafkaStreamProducer](kafkaStreamProducerNum)
+    val source = new KafkaSource("topic1", "localhost:2181",
+      classOf[DefaultMessageDecoder].getName, new DefaultTimeStampFilter)
```
whjiang: 1. Can we use ClassTag to avoid using ```classOf[XXX].getName```?
2. Better make ```DefaultMessageDecoder``` and ```DefaultTimeStampFilter``` consistent in passing. E.g. both passing in class name.
3. It is better to hide DefaultTimeStampFilter as default parameter as this is rarely needed to change.

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/KafkaSource.scala
```diff
@@ -36,62 +39,131 @@ object KafkaSource {
...
+   * @param topics comma-separated string of topics
+   * @param zkConnect kafka consumer config `zookeeper.connect`
+   * @param messageDecoder decode gearpump [[Message]] from raw bytes
+   * @param timestampFilter filters out message based on timestamp
+   */
+  def this(topics: String, zkConnect: String, messageDecoder: String, timestampFilter: TimeStampFilter) = {
+    this(KafkaUtil.setZookeeperConnect(KafkaConfig.load("kafka.conf")
+        .withConsumerTopics(topics)
+        .withMessageDecoder(messageDecoder), zkConnect), timestampFilter)
+  }
```
whjiang: Better to add another constructor:
```
def this(topics: String, zkConnect: String, properties: Properties /* additional kafka configuration parameters */, messageDecoder: String = DefaultMessageDecoder, timestampFilter: TimeStampFilter=DefaultTimeStampFilter)
```

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/StreamApp.scala
whjiang: as discussed yesterday, is it possible to modify ```Traverse``` interface to pass TaskContext inside? So that we don't need to create a fake TaskContext object.

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/KafkaSink.scala
```diff
@@ -20,27 +20,51 @@ package org.apache.gearpump.streaming.kafka
...
+      new ByteArraySerializer, new ByteArraySerializer), topic)
+  }
 
-  def write(topic: String, key: Array[Byte], msg: Array[Byte]): Unit = {
-    producer.send(new ProducerRecord[Array[Byte], Array[Byte]](topic, key, msg))
+  /**
+   * @param topic producer topic
+   * @param bootstrapServers kafka producer config `bootstrap.servers`
+   */
+  def this(topic: String, bootstrapServers: String) = {
```
whjiang: Better add a default parameter: ```properties : Properties```.

external/kafka/src/main/scala/org/apache/gearpump/streaming/kafka/KafkaSource.scala
```diff
@@ -36,62 +39,131 @@ object KafkaSource {
...
-   * fetchThread will start from beginning (the earliest offset)
-   * on start offset not set
+   * @param topics comma-separated string of topics
+   * @param timestampFilter filters out message based on timestamp
    */
-  def startFromBeginning(): Unit = {
-    fetchThread.setDaemon(true)
-    fetchThread.start()
+  def this(topics: String, timestampFilter: TimeStampFilter) = {
+    this(KafkaConfig.load("kafka.conf").withConsumerTopics(topics), timestampFilter)
```
whjiang: suggest removing this constructor as it implicitly replying on kafka.conf to function. If there is no such file, it will fail with strange error. 

user must read this implementation to understand the behavior, which is not friendly.

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/StreamApp.scala
manuzhang: It's possible but I don't want to modify `Traverse` in this PR since it is a general interface which may have other use cases we haven't considered.

whjiang:
If possible, better to make ```KafkaConfig``` not relying on specific file name ```kafka.conf```. Just a suggestion. E.g. reading KafkaConfig from application.conf, etc.

manuzhang:
`kafka.conf` is an `application.conf` dedicated to kafka configurations and `application.conf` should be left to users.

manuzhang:
@whjiang 
removed `kafka.conf` and add default parameters to `KafkaSource` constructors

clockfly:
test fails.

manuzhang:
I don't think they are related to changes in this PR. Also, `KafkaWordCountSpec` passes on my laptop.

PR #1124, opened 2015-07-06
fix #1123: able to restart a running application via command-line or …

### LINKED ISSUE #1123: Be able to restart a running application
The application will be restarted with the same configuration but user data will be cleaned up. 
Expected restapi call would be `POST /api/v1/appmaster/{id}/restart`. The dashboard should also have the ability to operate.

stanleyxu2005:
…dashboard


clockfly:
+1

PR #1145, opened 2015-07-08
fix #1144, out of memory when trying to scale gearpump to 2000 task on a 2 core machine

### LINKED ISSUE #1144: [scalability] out of memory when trying to scale gearpump to 2000 task on a 2 core machine.

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1145)

### DIFF
--- a/streaming/src/main/scala/org/apache/gearpump/streaming/task/Subscription.scala
+++ b/streaming/src/main/scala/org/apache/gearpump/streaming/task/Subscription.scala
@@ -109,6 +109,11 @@ class Subscription(
         transport.transport(ackRequest, targetTask)
       }
 
+      if (messageCount(partition) % maxPendingMessageCount == 0) {
+        val probeLatency = LatencyProbe(System.currentTimeMillis())
+        transport.transport(probeLatency, targetTask)
+      }
+
       return 1
     } else {
       if (needFlush) {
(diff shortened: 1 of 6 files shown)


manuzhang:
```
[error] (gearpump-streaming/test:compile) Compilation failed
```

streaming/src/main/scala/org/apache/gearpump/streaming/task/Subscription.scala
```diff
@@ -109,6 +109,11 @@ class Subscription(
         transport.transport(ackRequest, targetTask)
       }
 
+      if (messageCount(partition) % maxPendingMessageCount == 0) {
```
manuzhang: why not make this configurable like `AckRequest` ?

clockfly: What is the value for user to config it?

clockfly: I think the story is different compared with ackRequest frequency setting.

For ackRequest setting, the size will impact the memory footprint in Mailbox. 
For ProbeLatency, it is not necessary to expose this concept to user, I think.

manuzhang: how could we know this sort of flow control is enough when hitting even higher number of task?

clockfly: The frequency of AckRequest freqneucy and ProbeLatency is binded with the frequency of normal message.

By default, AckRequest will be 1% of normal message count.
The latencyProbe will be 0.1% of normal message count.

That means if one task have send 1000 normal message, then there will be 10 AckRequest message, and 1 LatencyProbe message.

Support the flow control works for normal messge, then the same flow control mechanism should also work for AckRequest and LatencyProbe.

manuzhang:
+1

PR #1147, opened 2015-07-08
fix #1146, actor hungry when worker use block-io to wait response from FileServer

### LINKED ISSUE #1146: [availability] worker HttpClient block-IO can cause actor hungry
When worker doing blocking IO to get jar from FileServer, it will occupy the thread pool. And then FileServer may not be abled to scheduled to serve this IO request.

Solution:
Ture HttpClient block-io into a future, and use dedicated thread pool for block-io operation.

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1147)

### DIFF
--- a/daemon/src/main/scala/org/apache/gearpump/cluster/worker/Worker.scala
+++ b/daemon/src/main/scala/org/apache/gearpump/cluster/worker/Worker.scala
@@ -28,7 +28,7 @@ import com.typesafe.config.{Config, ConfigFactory}
 import org.apache.gearpump.cluster.AppMasterToMaster.{WorkerData, GetWorkerData}
 import org.apache.gearpump.cluster.AppMasterToWorker._
 import org.apache.gearpump.cluster.ClientToMaster.QueryWorkerConfig
-import org.apache.gearpump.cluster.ClusterConfig
+import org.apache.gearpump.cluster.{ExecutorJVMConfig, ClusterConfig}
 import org.apache.gearpump.cluster.MasterToClient.WorkerConfig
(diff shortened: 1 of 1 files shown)


manuzhang:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

<sup>**[daemon/src/main/scala/org/apache/gearpump/cluster/worker/Worker.scala, line 334 \[r1\]](https://reviewable.io:443/reviews/intel-hadoop/gearpump/1147#-Jtfx1tSrPpTyV0bfpzg)** ([raw file](https://github.com/intel-hadoop/gearpump/blob/d31b80402652e648c235662e4edfb5ac63a43f0a/daemon/src/main/scala/org/apache/gearpump/cluster/worker/Worker.scala#L334)):</sup>
better with `foreach` since no return value is required

---

<sup>**[daemon/src/main/scala/org/apache/gearpump/cluster/worker/Worker.scala, line 336 \[r1\]](https://reviewable.io:443/reviews/intel-hadoop/gearpump/1147#-JtfxJXnhJJx2oAvDzQD)** ([raw file](https://github.com/intel-hadoop/gearpump/blob/d31b80402652e648c235662e4edfb5ac63a43f0a/daemon/src/main/scala/org/apache/gearpump/cluster/worker/Worker.scala#L336)):</sup>
could be done in one `foreach`

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/intel-hadoop/gearpump/1147)

manuzhang:
+1. please rebase

Issue #997, opened 2015-06-04
use external argument parse library


manuzhang:
We've found a few issues on argument parse recently and the current codes are quite complicated and error-prone. I'm thinking about using an external argument parse library rather than maintaining our own.

There is already a question on [stackoverflow](http://stackoverflow.com/questions/2315912/scala-best-way-to-parse-command-line-parameters-cli), which provides a bunch of options like Scopt, Argot and Scallop.

clockfly:
Please re-open this when you have a clear proposal.

PR #1150, opened 2015-07-08
fix #1149, Shell tools printed too much detail on console

### LINKED ISSUE #1149: Shell tools printed too much detail on console.

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1150)

### DIFF
--- a/daemon/src/main/scala/org/apache/gearpump/cluster/main/Gear.scala
+++ b/daemon/src/main/scala/org/apache/gearpump/cluster/main/Gear.scala
@@ -22,69 +22,34 @@ import org.slf4j.Logger
 
 import scala.util.Try
 
-object Gear extends App {
+object Gear  {
 
   private val LOG: Logger = LogUtil.getLogger(getClass)
 
-  def usage(commandOption: Option[String]) = {
-    commandOption match {
-      case Some(command) =>
-        command match {
-          case "kill" =>
-            Kill.main(Array.empty[String])
-          case "shell" =>
-            Shell.main(Array.empty[String])
-          case "info" =>
-            Info.main(Array.empty[String])
-          case "replay" =>
-            Replay.main(Array.empty[String])
-          case "app" =>
-            AppSubmitter.main(Array.empty[String])
-          case x =>
-            throw new Exception("Unknown command " + x)
-        }
-      case None =>
-        Console.println("Usage: app|info|kill|shell|replay|mainClass ...")
-    }
+  val commands = Map("app" -> AppSubmitter, "kill" -> Kill, "shell" -> Shell,
+    "info" -> Info, "replay" -> Replay, "main" -> MainRunner)
+
+  def usage: Unit = {
+    val keys = commands.keys.toList.sorted
+    Console.println("Usage: " + "<" + keys.mkString("|") + ">")
   }
 
   def executeCommand(command : String, commandArgs : Array[String]) = {
-
-    command match {
-      case "kill" =>
-        Kill.main(commandArgs)
-      case "shell" =>
-        Shell.main(commandArgs)
-      case "info" =>
-        Info.main(commandArgs)
-      case "replay" =>
-        Replay.main(commandArgs)
-      case "app" =>
-        AppSubmitter.main(commandArgs)
-      case main =>
-        val customCommand = Try {
-          val clazz = Thread.currentThread().getContextClassLoader().loadClass(main)
-          val mainMethod = clazz.getMethod("main", classOf[Array[String]])
-          mainMethod.invoke(null, commandArgs)
-        }
-        if (customCommand.isFailure) {
-          val ex = customCommand.failed.get
-          LOG.error(s"failed to execute command $main", ex)
-          throw ex
-        }
+    commands.get(command).map(_.main(commandArgs))
+    if (!commands.contains(command)) {
+      val allArgs = (command +: commandArgs.toList).toArray
+      MainRunner.main(allArgs)
     }
   }
 
-  def start = {
+  def main(args: Array[String]) = {
     args.length match {
       case 0 =>
-        usage(None)
+        usage
       case a if(a >= 1) =>
         val command = args(0)
         val commandArgs = args.drop(1)
         executeCommand(command, commandArgs)
     }
   }
-
-  start
 }
--- a/daemon/src/main/scala/org/apache/gearpump/cluster/main/MainRunner.scala
+++ b/daemon/src/main/scala/org/apache/gearpump/cluster/main/MainRunner.scala
@@ -15,34 +15,20 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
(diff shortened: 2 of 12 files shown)


clockfly:
test failure is not related.

manuzhang:
one question. Do we want to throw exception on `gear app` ?

current behavior 
```
output/target/pack/bin/gear app
Usage:
org.apache.gearpump.cluster.main.AppSubmitter -namePrefix <application name prefix> -jar <application>.jar 
-namePrefix  (required:false, default:)
-jar  (required:true)
Exception in thread "main" java.lang.Exception: Missing option jar...
	at org.apache.gearpump.cluster.main.ArgumentsParser$$anonfun$parse$2.apply(ArgumentsParser.scala:130)
	at org.apache.gearpump.cluster.main.ArgumentsParser$$anonfun$parse$2.apply(ArgumentsParser.scala:127)
	at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:186)
	at org.apache.gearpump.cluster.main.ArgumentsParser$.parse(ArgumentsParser.scala:127)
	at org.apache.gearpump.cluster.main.ArgumentsParser$class.parse(ArgumentsParser.scala:63)
	at org.apache.gearpump.cluster.main.AppSubmitter$.parse(AppSubmitter.scala:29)
	at org.apache.gearpump.cluster.main.AppSubmitter$.start(AppSubmitter.scala:40)
	at org.apache.gearpump.cluster.main.AppSubmitter$.delayedEndpoint$org$apache$gearpump$cluster$main$AppSubmitter$1(AppSubmitter.scala:89)
	at org.apache.gearpump.cluster.main.AppSubmitter$delayedInit$body.apply(AppSubmitter.scala:29)
	at scala.Function0$class.apply$mcV$sp(Function0.scala:40)
	at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:12)
	at scala.App$$anonfun$main$1.apply(App.scala:76)
	at scala.App$$anonfun$main$1.apply(App.scala:76)
	at scala.collection.immutable.List.foreach(List.scala:381)
	at scala.collection.generic.TraversableForwarder$class.foreach(TraversableForwarder.scala:35)
	at scala.App$class.main(App.scala:76)
	at org.apache.gearpump.cluster.main.AppSubmitter$.main(AppSubmitter.scala:29)
	at org.apache.gearpump.cluster.main.Gear$.executeCommand(Gear.scala:38)
	at org.apache.gearpump.cluster.main.Gear$.main(Gear.scala:52)
	at org.apache.gearpump.cluster.main.Gear.main(Gear.scala)
```

daemon/src/main/scala/org/apache/gearpump/cluster/main/MainRunner.scala
```diff
@@ -0,0 +1,36 @@
...
+import org.slf4j.Logger
+
+import scala.util.Try
+
+object MainRunner extends App with ArgumentsParser {
+  private val LOG: Logger = LogUtil.getLogger(getClass)
+
+  override val options: Array[(String, CLIOption[Any])] = Array.empty
+
+  val mainClazz = args(0)
```
manuzhang: any usage for this ? Otherwise, `ArrayIndexOutOfBoundsException` is thrown

daemon/src/main/scala/org/apache/gearpump/cluster/main/Gear.scala
```diff
@@ -22,69 +22,34 @@ import org.slf4j.Logger
...
-        }
-      case None =>
-        Console.println("Usage: app|info|kill|shell|replay|mainClass ...")
-    }
+  val commands = Map("app" -> AppSubmitter, "kill" -> Kill, "shell" -> Shell,
+    "info" -> Info, "replay" -> Replay, "main" -> MainRunner)
+
+  def usage: Unit = {
+    val keys = commands.keys.toList.sorted
+    Console.println("Usage: " + "<" + keys.mkString("|") + ">")
```
manuzhang: maybe what we can do better here is to print the meaning of each command

manuzhang:
how running wordcount looks like. 

```
$ output/target/pack/bin/gear app -jar examples/target/scala-2.11/gearpump-examples-wordcount-assembly-0.3.8-SNAPSHOT.jar
Usage:
org.apache.gearpump.streaming.examples.wordcount.WordCount -split <how many split tasks> -sum <how many sum tasks> 
-split  (required:false, default:1)
-sum  (required:false, default:1)
```

usage is printed even no error.

daemon/src/main/scala/org/apache/gearpump/cluster/main/MainRunner.scala
clockfly: This is for usage like gear gearoump.MainClass
Parameter length is not checked.

daemon/src/main/scala/org/apache/gearpump/cluster/main/Gear.scala
clockfly: I agree.

clockfly:
@manuzhang

I agree all the comments you made above. I think they should be addressed in a different PR. It will be great if you can do it.
 
As the title indicated, the purpose of this one is to avoid printing too much info logs on console.

daemon/src/main/scala/org/apache/gearpump/cluster/main/MainRunner.scala
manuzhang: sorry, what I mean is any usage info to help users use this command.

clockfly: We call calling user.MainClass.

user.MainClass is defined by user, its user's responsiblity to provide context help.

manuzhang:
@clockfly 

Agreed. Then the only problem is the last one.  We don't print usage on correct commands before this so this PR should not introduce another issue when solving one.

clockfly:
@manuzhang 

```
We don't print usage on correct commands before this so this PR should not introduce another issue when solving one.
```

If you look into the old code, there is no place which show context help. 
Printing some help is better than nothing.

manuzhang:
+1.

What I've been thinking is #997, using some external argument parser library. Once I tried to fix some of the above issues but found it really difficult to cover all the cases.

clockfly:
@manuzhang 

I made some updates based on your comments above. Please check whether these new changes make sense.

manuzhang:
do we need to tell users that they've successfully submit the application ?

kkasravi:
I would say yes - some confirmation would typically be shown in the UI  


     On Wednesday, July 8, 2015 7:13 AM, ManuZhang <notifications@github.com> wrote:
   

 do we need to tell users that they've successfully submit the application ?—

PR #1152, opened 2015-07-08
fix #1151, reduce the memory footprint

### LINKED ISSUE #1151: scale up to 10, 000 tasks on single four-core machine
In theory, with effecitve flow control, we should be able to scale to as many tasks as we want.

But in reality, it depends on:
1. memory footprint of single task is small enough, so that there is no out of memory.
2. GC pause interval is still under control to avoid remote watching timeout.

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1152)


huafengw:
compile failed

clockfly:
fixed.

Issue #1154, opened 2015-07-08
Self contained jar applications should be able to specify arguments


kkasravi:
[Story](https://www.pivotaltracker.com/story/show/98673918)
Loading client applications via REST as jar files should provide a way to specify command line args and conf files

clockfly:
Besides jar, we should also allow user to specify:
1. MainClass [optional]. If not specified, we will use the default MainClass in manifest.in
2. Command line arguments.

stanleyxu2005:
The feature is nearly done. But I am thinking to split upload and launch in two apis. There are two reasons:
1. Upload once, launch with arguments you want. You can save bandwidth to upload repeatedly.
2. For security intensive usage, maybe only admin is privileged to deploy application. User/operator is only allowed to run application.

kkasravi:
There should be 2 APIs - we will need to surface a catalog of available apps in next demo...

clockfly:
the PR not working.

For example, when choose wordcount.jar, and use paramter
-split 2 -sum 1

PR #1155, opened 2015-07-08
Fixes #1105 Create kafka_hdfs demo under examples

### LINKED ISSUE #1105: Move kafka related examples out of intel-hadoop/gearpump into gearpump/gearpump-examples
[Story](https://www.pivotaltracker.com/story/show/98296880)
Move kafka related examples out of intel-hadoop/gearpump into gearpump/gearpump-examples

kkasravi:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1155)


kkasravi:
Build failure is unrelated

[info] MasterWatcher
[info] - should kill itself when can not get a quorum *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (5 seconds) during expectMsg: Terminated TestActor[akka://ForMasterWatcher/user/$$a]
[info]   at scala.Predef$.assert(Predef.scala:165)
[info]   at akka.testkit.TestKitBase$class.expectMsgPF(TestKit.scala:356)
[info]   at akka.testkit.TestKit.expectMsgPF(TestKit.scala:718)
[info]   at akka.testkit.TestKitBase$class.expectTerminated(TestKit.scala:368)
[info]   at akka.testkit.TestKit.expectTerminated(TestKit.scala:718)
[info]   at org.apache.gearpump.cluster.main.MasterWatcherSpec$$anonfun$1.apply$mcV$sp(MasterWatcherSpec.scala:46)
[info]   at org.apache.gearpump.cluster.main.MasterWatcherSpec$$anonfun$1.apply(MasterWatcherSpec.scala:40)
[info]   at org.apache.gearpump.cluster.main.MasterWatcherSpec$$anonfun$1.apply(MasterWatcherSpec.scala:40)
[info]   ...

conf/kafka-hdfs-pipeline.conf.template
```diff
@@ -0,0 +1,43 @@
+gearpump {
```
manuzhang: I think the conf should go to `src/main/resources/reference.conf` while kafka config file has been removed after the refactoring

examples/streaming/kafka-hdfs-pipeline/scripts/createtopic
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
manuzhang: how about getting these scripts into a README as they could not be generally applied ?

examples/streaming/kafka-hdfs-pipeline/src/main/scala/org/apache/gearpump/examples/streaming/pipeline/KafkaHdfsPipeLine.scala
```diff
@@ -0,0 +1,150 @@
...
+    val scorerNum = config.getInt("scorer")
+    val writerNum = config.getInt("writer")
+    val outputPath = config.getString("output")
+    val pipeLinePath = config.getString("conf")
+    val file = new java.io.File(pipeLinePath)
+    val pipeLineConfig = file.exists() match {
+      case true =>
+        ConfigFactory.parseFile(file)
+      case false =>
+        val fallback =
```
manuzhang: no need for the long list of kafka configs now. `val source = new KafkaSource("topic-105", "10.10.10.46:2181,10.10.10.236:2181,10.10.10.164:2181/kafka")` will do

examples/streaming/kafka-hdfs-pipeline/src/test/scala/org/apache/gearpump/examples/streaming/pipeline/KafkaHdfsPipeLineSpec.scala
```diff
@@ -0,0 +1,119 @@
...
+import org.apache.gearpump.streaming.task.{StartTime, Task, TaskContext}
+import org.apache.gearpump.streaming.transaction.api.TimeReplayableSource
+import org.apache.gearpump.util.LogUtil
+import org.scalatest.prop.PropertyChecks
+import org.scalatest.{BeforeAndAfterAll, Matchers, PropSpec}
+import org.slf4j.Logger
+
+import scala.util.{Failure, Success, Try}
+
+class SpaceShuttleReplayableSource extends TimeReplayableSource {
```
manuzhang: please checkout the new `DataSource` (`TimeReplaybaleSource`) and `DataSink` API. It has changed after latest refactoring

manuzhang:
@kkasravi 

This seems not to be based on latest trunk. We have merged several big features (dynamic dag, new kafka connector) recently. Please rebase. 

Or I could help merge the `kafka_hdfs` branch on intel-hadoop into master since I'm more familiar with the new kafka connector.

kkasravi:
@manuzhang @clockfly 

Sure although let me enumerate the issues:
1. We need to be able to submit this jar via web ui (actually all examples)
2. In order not to embed specific ip's (zookeeper.connect for example) we need an ability to either specify command line args (-D) that override conf properties or a custom conf file within the jar file.
3. The fallback val was used as a way to do this until a better solution is designed (see [#1154](https://github.com/intel-hadoop/gearpump/issues/1154) )
4. I think most of your comments above were based on an earlier version - please take a look at the latest PR.
5. The current commit does use DataProcessor 
```
    val source = new KafkaSource(kafkaConfig.getConsumerTopics.head, kafkaConfig.consumerConfig.getProperty(ZOOKEEPER_CONNECT))
    val reader = DataSourceProcessor(source, readerNum)
```
   (BTW it might be nice to add KafkaConfig.getZookeepers - similar to KafkaConfig.getConsumerTopics)
6. My recommendation is to merge as is and the design and implementation to better handle a jar that can parse an external config or command line args can be done as a follow up task in #1154. For kafka-hdfs-pipeline and kafka-hbase-pipeline we would probably want certain values to be changed easily like topics, zookeeper.connect and client.id from the command line.

kkasravi:
Note above build failure is unrelated
```
[info] MasterWatcher
[info] - should kill itself when can not get a quorum *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (5 seconds) during expectMsg: Terminated TestActor[akka://ForMasterWatcher/user/$$a]
[info]   at scala.Predef$.assert(Predef.scala:165)
[info]   at akka.testkit.TestKitBase$class.expectMsgPF(TestKit.scala:356)
[info]   at akka.testkit.TestKit.expectMsgPF(TestKit.scala:718)
[info]   at akka.testkit.TestKitBase$class.expectTerminated(TestKit.scala:368)
[info]   at akka.testkit.TestKit.expectTerminated(TestKit.scala:718)
[info]   at org.apache.gearpump.cluster.main.MasterWatcherSpec$$anonfun$1.apply$mcV$sp(MasterWatcherSpec.scala:46)
[info]   at org.apache.gearpump.cluster.main.MasterWatcherSpec$$anonfun$1.apply(MasterWatcherSpec.scala:40)
[info]   at org.apache.gearpump.cluster.main.MasterWatcherSpec$$anonfun$1.apply(MasterWatcherSpec.scala:40)
[info]   ...
```

manuzhang:
@kkasravi 

1. Many of the configs below have been moved and please check `KafkaConfig.scala` for that. 
```
kafka {
    consumer {
      # change to match cluster zookeepers
      zookeeper.connect = "127.0.0.1:2181"
      # consumer topics separated by ","
      topics = "topic-105"
      client.id = "kafka_hdfs_app"
      socket.timeout.ms = 30000
      socket.receive.buffer.bytes = 65536
      fetch.message.max.bytes = 1048576
      emit.batch.size = 100
      fetch.threshold = 5000
      fetch.sleep.ms = 100
    }

    producer {
      topic = "topic-105"
      # list of host/port pairs to use for establishing the initial connection to kafka server
      # list should be in the form "host1:port1,host2:port2,..."
      # list need not contain the full set of servers
      bootstrap.servers = "127.0.0.1:9092"
      acks = "1"
      buffer.memory = 33554432
      compression.type = "none"
      retries = 0
      batch.size = 16384
    }

    storage.replicas = 2

    grouper.factory.class = "org.apache.gearpump.streaming.kafka.lib.grouper.KafkaDefaultGrouperFactory"

    task {
      message.decoder.class = "org.apache.gearpump.streaming.kafka.lib.DefaultMessageDecoder"
      timestamp.filter.class = "org.apache.gearpump.streaming.kafka.lib.KafkaFilter"
    }
}
```

We no longer set kafka consumer/producer configs through this file which are supposed to be from the kafka standard `consumer.properties` and `producer.properties` (if you `cd` into `kafka/conf`). They corresponds to `consumerConfig` and `producerConfig` in `KafkaConfig`. Also, `kafka.task.timestamp.filter.class` has been moved (pls. refer to `DataSourceConfig.scala`) since it is general to all `DataSource`. It's my bad for not having added docs. Sorry.

2. My concern with the hardcoded backup config is we may forget to update it as the config changes.

3. We'd better not add a template to `conf` for every new examples. I think they should go to their own directories.

examples/streaming/kafka-hdfs-pipeline/src/test/scala/org/apache/gearpump/examples/streaming/pipeline/KafkaHdfsPipeLineSpec.scala
```diff
@@ -0,0 +1,119 @@
...
+      .stripMargin,
+    """
+      |{"id":"043ade58-2fbc-4fe2-8253-84ab181b8cfa","on":"2015-04-02T18:52:02.680078434Z","body": "[-0.414141,-0.0246564,-0.125,0.0140301,-0.474359,0.0256049,-0.0980392,0.463884,0.40836]"}
+    """.stripMargin,
+    """
+      |{"id":"043ade58-2fbc-4fe2-8253-84ab181b8cfa","on":"2015-04-02T18:52:02.680078434Z","body": "[-0.414141,-0.0246564,-0.125,0.0140301,-0.474359,0.0256049,-0.0980392,0.463884,0.40836]"}
+    """.stripMargin
+  )
+
+  def startFromBeginning(): Unit = {}
```
manuzhang: this has been removed

examples/streaming/kafka-hdfs-pipeline/src/test/scala/org/apache/gearpump/examples/streaming/pipeline/KafkaHdfsPipeLineSpec.scala
```diff
@@ -0,0 +1,119 @@
...
+
+  def open(context: TaskContext, startTime: Option[TimeStamp]): Unit = {}
```
manuzhang: pls. add `override` so we get compile errors if the implementation no longer conforms to interface.

clockfly:
@manuzhang 

Since you are familar with new Kafka API, can you fix this pull request?

kkasravi:
I'm making the changes @manuzhang suggested along with a README.md. Note that the three app's that take conf arguments are yarn, kafka-hdfs-pipeline and kafka-hbase-pipeline. We may want to deal with all of these in the same way eg #1154. @manuzhang is there no longer a way to start from the beginning in a kafka topic?

manuzhang:
@kkasravi 
source.open(context, None) would mean start from beginning for kafka

kkasravi:
@manuzhang how would you specify the above from DataProcessor or would you need to define your own Task that extends DataSourceTask and override onStart (or something similar)?

I've simplified KafkaHdfsPipeline, unfortunately it seems to break when running under YARN. Not sure why yet. Do you see anything that's obvious (and incorrect)?

kkasravi:
@manuzhang Below are copied from the executor logs - i'm assuming i do not need to set producer values since i'm only consuming however maybe that is incorrect.
```
[INFO] [07/10/2015 05:24:39.110] [ClientCnxn] Session establishment complete on server 10.10.10.236/10.10.10.236:2181, sessionid = 0x14e4e436be83c6e, negotiated timeout = 6000
[INFO] [07/10/2015 05:24:39.110] [ZkClient] zookeeper state changed (SyncConnected)
[INFO] [07/10/2015 05:24:39.153] [AdminUtils$] Topic creation {"version":1,"partitions":{"0":[2]}}
[INFO] [07/10/2015 05:24:39.156] [KafkaUtil$] created topic app1_KafkaHdfsPipeLine_topic-105_0
[INFO] [07/10/2015 05:24:39.156] [ZkEventThread] Terminate ZkClient event thread.
[INFO] [07/10/2015 05:24:39.158] [ZooKeeper] Session: 0x14e4e436be83c6e closed
[INFO] [07/10/2015 05:24:39.158] [ClientCnxn] EventThread shut down
[INFO] [07/10/2015 05:24:39.173] [ProducerConfig] ProducerConfig values:
        value.serializer = class org.apache.kafka.common.serialization.ByteArraySerializer
        key.serializer = class org.apache.kafka.common.serialization.ByteArraySerializer
        block.on.buffer.full = true
        retry.backoff.ms = 100
        buffer.memory = 33554432
        batch.size = 16384
        metrics.sample.window.ms = 30000
        metadata.max.age.ms = 300000
        receive.buffer.bytes = 32768
        timeout.ms = 30000
        max.in.flight.requests.per.connection = 5
        bootstrap.servers = [localhost:9092]
        metric.reporters = []
        client.id =
        compression.type = none
        retries = 0
        max.request.size = 1048576
        send.buffer.bytes = 131072
        acks = 1
        reconnect.backoff.ms = 10
        linger.ms = 0
        metrics.num.samples = 2
        metadata.fetch.timeout.ms = 60000

[WARN] [07/10/2015 05:24:39.202] [ProducerConfig] The configuration value.serializer = class org.apache.kafka.common.serialization.ByteArraySerializer was supplied but isn't a known config.
[WARN] [07/10/2015 05:24:39.202] [ProducerConfig] The configuration key.serializer = class org.apache.kafka.common.serialization.ByteArraySerializer was supplied but isn't a known config.
[INFO] [07/10/2015 05:24:39.207] [TaskActor@app1exec0TaskId(0,0)] TaskLocationReady, sending GetUpstreamMinClock to AppMaster
[WARN] [07/10/2015 05:24:39.291] [Selector] Error in I/O with localhost/127.0.0.1
java.net.ConnectException: Connection refused
        at sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)
        at sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:739)
        at org.apache.kafka.common.network.Selector.poll(Selector.java:238)
        at org.apache.kafka.clients.NetworkClient.poll(NetworkClient.java:192)
        at org.apache.kafka.clients.producer.internals.Sender.run(Sender.java:191)
        at org.apache.kafka.clients.producer.internals.Sender.run(Sender.java:122)
        at org.apache.kafka.clients.producer.internals.Sender.run(Sender.java:191)
        at org.apache.kafka.clients.producer.internals.Sender.run(Sender.java:122)
        at java.lang.Thread.run(Thread.java:744)
````

**Comment by manuzhang** [comment]
@kkasravi 

1. Yes. We need to extend `DataSourceTask` and override `onStart` method for that. Plus, `DataSourceProcessor` cannot help here.

2. `KafkaSource` persists offset-timestamp mapping to kafka. That's why there is a kafka producer. It seems that kafka broker is not available when the producer tries to send message to it. Could you check for that ?

**Comment by kkasravi** [comment]
@manuzhang 

Thanks. It would be nice to have an easier way of doing this other than extending DataSourceTask. I can see users either wanting to start from the beginning or from a particular offset. 

I did confirm that the brokers were available using kafka-console-consumer and providing the same zookeeper values. It sounds like for the consumer we need to pass broker locations so the offsets can be written. The brokers are on different nodes (not localhost) - how does one pass broker locations for KafkaSource. Does KafkaWordCount assume that the broker is running locally?

**Comment by manuzhang** [comment]
Nope. broker locations are stored in zookeeper and KafkaSource fetch those locations from zookeeper so make sure you can connect to that is configured in `kafka/conf/server.properties`.

**Comment by kkasravi** [comment]
I needed to set a BOOTSTRAP_SERVERS property in the producerConfig section for the client to work. Also I discovered that setting CONSUMER_TOPICS in the consumerConfig section actually doesn't set the topic, you need to set it in the first parameter (Config) of KafkaConfig or use KafkaConfig.withConsumerTopics. Over all it ended up being the following:

```
    val consumerProperties = new Properties()
    consumerProperties.setProperty(ZOOKEEPER_CONNECT, zookeepers)
    val producerProperties = new Properties()
    producerProperties.setProperty(BOOTSTRAP_SERVERS, "10.10.10.46:9092,10.10.10.164:9092,10.10.10.236:9092")
    val kafkaConfig = new KafkaConfig(
      consumerConfig=consumerProperties,
      producerConfig=producerProperties
    ).withConsumerTopics(topic)

    val partitioner = new ShufflePartitioner()
    val source = new KafkaSource(kafkaConfig)
    val reader = DataSourceProcessor(source, readerNum)
```

From the code it looks like localhost:9092 is used if there is no BOOTSTRAP_SERVERS_CONFIG set in the ProducerConfig. I might be wrong on that. If the brokers can be fetched from zookeeper as noted above should we use those brokers for our offsets rather than possibly using a broker running on localhost? It seems like you might have the scenario where offsets could be stored on a local broker (where ever the kafka node is initialized) but the consumption could occur from entirely different brokers. Seems like in event of checkpoint and recover the offsets could be potentially lost.

**Comment by manuzhang** [comment]
Well, consumer topics are not part of the [kafka consumer config](http://kafka.apache.org/documentation.html#consumerconfigs). That's why setting consumer topics in consumerConfig (which is for kafka consumer config) doesn't work. Kafka consumers fetch broker locations from zookeeper and then pull from them while Kafka producers require them passed in by users. You will notice the difference when you use `kafka-console-consumer.sh` compared with `kafka-producer-perf-test.sh`.

You are right that `localhost:9092` is used if no bootstrap servers provided. The problem is that currently the offset storage is coupled with `KafkaSource` which doesn't need a producer config in theory. I'll open a PR to decouple them soon.

**Comment by kkasravi** [comment]
Might be nice if feedback is provided for unrecognized ( or misplaced properties) though you would probably need to do that on client side as part or arg validation.

The exception I noted above is a bit difficult to figure why it happened unless you understand that the producer part is required for a consumer's offset storage. Seems like you did a great job of hiding offset storage from the user so it's odd to have it surface here. Decoupling producer settings from KafkaSource would really make the API simple and useful. Thanks for all the follow up.

**Comment by manuzhang** [comment]
@kkasravi  any updates on this PR ? I'd like to merge this in soon afraid of any new changes to `KafkaSource` would break things here.

**Comment by manuzhang** [comment]
build failed ?
```
[error] (gearpump-experiments-yarn/test:compile) Compilation failed
```

kkasravi:
Error doesn't appear to be related

[INFO] [07/13/2015 04:25:41.795] [SequenceFileIOSpec] Actor system is stopped, 127.0.0.1, 48826
[info] SequenceFileIOSpec:
[info] - SequenceFileIO should succeed to submit application with required arguments *** FAILED ***
[info]   AssertionError was thrown during property evaluation. (SequenceFileIOSpec.scala:57)
[info]     Message: assertion failed: timeout (25 seconds) during expectMsgClass waiting for class org.apache.gearpump.cluster.ClientToMaster$SubmitApplication
[info]     Occurred at table row 0 (zero based, not counting headings), which had values (
[info]       requiredArgs = [Ljava.lang.String;@7fa53ce6,
[info]       optionalArgs = [Ljava.lang.String;@39416182
[info]     )

manuzhang:
@kkasravi 
I noticed you upgraded versions of hadoop, hbase and cdh. How about putting that in a separate PR 'cause I think this is risky and I've also found some unwanted dependencies when building the PR ?

kkasravi:
@manuzhang Separating the cdh upgrade will be difficult since the target environment(s) for kafka-hdfs are on cdh 5.4.2 so with parquet/hdfs part of the DAG there were problems building on 5.3.2 and running on 5.4.2.

kkasravi:
SequenceFileIOSpec runs fine locally. I also took a sample sequence file and ran the example (fsio).

clockfly:
Since we are working on removing external dependencies.

Does it make sense to put this example at https://github.com/gearpump/examples?

kkasravi:
It's probably an example that many would want to see due to the Kafka + HDFS connectors, but we should probably move to separate repo.

clockfly:
@kkasravi 

It is risky to upgrade to CDH for now. Let's put this under github.com/gerapump for now. till we totally remove the dependencies of Hadoop from gearpump core.

kkasravi:
Ok, I will resubmit a new PR for github.com/gearpump/examples. I think both pipeline examples should be move to there. Closing this.

PR #1170, opened 2015-07-11
fix #1154: submit application will accept launch arguments (4th)

### LINKED ISSUE #1154: Self contained jar applications should be able to specify arguments

stanleyxu2005:
due to the form data unpack code, the launch arguments (optional) are passed as url params instead of form data

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1170)

### DIFF
--- a/services/src/main/scala/org/apache/gearpump/services/MasterService.scala
+++ b/services/src/main/scala/org/apache/gearpump/services/MasterService.scala
@@ -20,7 +20,6 @@
 package org.apache.gearpump.services
 
 import java.io.{File, IOException}
-import java.net.URLClassLoader
 
 import akka.actor.{ActorRef, ActorSystem}
 import com.typesafe.config.Config
@@ -39,10 +38,11 @@ import org.apache.gearpump.streaming.appmaster.SubmitApplicationRequest
 import org.apache.gearpump.util.ActorUtil.{askActor, askWorker}
 import org.apache.gearpump.util.Constants.GEARPUMP_APP_JAR
 import org.apache.gearpump.util.{Constants, Util}
-import spray.http.{MediaTypes, MultipartFormData}
+import spray.http.{BodyPart, MediaTypes, MultipartFormData}
 import spray.routing
 import spray.routing.HttpService
 
+import scala.collection.JavaConversions._
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success}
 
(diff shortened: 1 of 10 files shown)


services/src/main/scala/org/apache/gearpump/services/MasterService.scala
```diff
@@ -98,14 +97,16 @@ trait MasterService extends HttpService {
...
         post {
-          respondWithMediaType(MediaTypes.`application/json`) {
-            entity(as[MultipartFormData]) { formData =>
-              val jar = formData.fields.head.entity.data.toByteArray
-              onComplete(Future(MasterService.submitJar(jar, system.settings.config))) {
-                case Success(_) =>
-                  complete(write(MasterService.Status(true)))
-                case Failure(ex) =>
-                  failWith(ex)
+          anyParams('args.as[Option[String]]) { (extraArgs) =>
```
kkasravi: I'm not sure extra args should be handled in this way versus providing them within a conf file (that would be packaged in the assembly). It would be difficult to handle the various types of CLI args possible including those that override specific conf properties. We may be better off handling this within the context of jar packaging.

stanleyxu2005: There will not have this concern, in case we split it as two APIs: one for upload a jar; one for launch the jar with specified arguments.

clockfly:
Maybe a additional control to upload application.conf(optional)?

stanleyxu2005:
@clockfly @kkasravi userapp.conf is defintively more flexible than feeding arguments. So I'd like to confirm that you agree to replace optional launch arguments input box with a upload file for *.conf.

clockfly:
HI Stanley,

I think both arguments list and application.conf are needed.

kkasravi:
Look how datatorrent does it (they accept both a conf and a custom set of properties)

image

stanleyxu2005:
@clockfly @kkasravi Patch revised. Supports config file and launch args.

core/src/main/scala/org/apache/gearpump/util/HttpUtil.scala
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.gearpump.util
+
+import spray.http._
+
+object HttpUtil {
+
+  def findFirstDataByteArray(bodyParts: Seq[BodyPart], suffix: String): Option[Array[Byte]] = {
```
clockfly: This code looks very tricky to me.

Are there native API in spray to do this?

clockfly:
+1

PR #1177, opened 2015-07-14
fix #1175 refactor dsl source and sink

### LINKED ISSUE #1175: refactor dsl source and sink part
Since we have refactor the DataSource and DataSink interface, we'd better refactor the dsl related part.

huafengw:
This is a draft for review

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1177)

### DIFF
--- a/examples/streaming/kafka-hbase-pipeline/src/main/scala/org/apache/gearpump/streaming/examples/pipeline/PipeLineDSL.scala
+++ b/examples/streaming/kafka-hbase-pipeline/src/main/scala/org/apache/gearpump/streaming/examples/pipeline/PipeLineDSL.scala
@@ -28,7 +28,8 @@ import org.apache.gearpump.external.hbase.{HBaseConsumer, HBaseRepo, HBaseSinkIn
 import Messages._
 import org.apache.gearpump.streaming.dsl.StreamApp
 import org.apache.gearpump.streaming.dsl.StreamApp._
-import org.apache.gearpump.streaming.kafka.lib.KafkaConfig
+import org.apache.gearpump.streaming.kafka.dsl.KafkaDSLUtil
+import org.apache.gearpump.streaming.kafka.lib.{StringMessageDecoder, KafkaConfig}
 import org.apache.gearpump.streaming.task.TaskContext
 import org.apache.gearpump.streaming.transaction.api.TimeReplayableSource
 import org.apache.gearpump.util.{Constants, LogUtil}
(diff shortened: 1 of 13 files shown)


experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/example/WordCount.scala
```diff
@@ -30,7 +30,7 @@ object WordCount extends App with ArgumentsParser{
     val context = ClientContext()
     val app = StreamApp("dsl", context)
     val data = "This is a good start, bingo!! bingo!!"
-    app.source(data.lines.toList, 1).
+    app.sourceFromCollection(data.lines.toList, 1).
```
clockfly: Maybe still use the name source?

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/op/OP.scala
```diff
@@ -19,22 +19,13 @@
...
 
 object OpType {
-  type TraverseType[T] = {
-    def foreach[U](t: T => U): Unit
-  }
-  trait Traverse[T] extends java.io.Serializable {
-    def foreach[U](t: T => U): Unit
-  }
-
   type SinkClosure[T] = (HBaseSinkInterface,HBaseConsumer) => (T => Unit)
```
clockfly: This line should be removed?

examples/streaming/kafka-hbase-pipeline/src/main/scala/org/apache/gearpump/streaming/examples/pipeline/PipeLineDSL.scala
```diff
@@ -80,12 +81,17 @@ object PipeLineDSL extends App with ArgumentsParser {
     System.setProperty(Constants.GEARPUMP_CUSTOM_CONFIG_FILE, pipeLinePath)
 
     val app = StreamApp("PipeLineDSL", context, appConfig)
-    val producer = app.readFromKafka(kafkaConfig, msg => {
-      val jsonData = msg.msg.asInstanceOf[String]
-      val envelope = read[Envelope](jsonData)
+    val producer = KafkaDSLUtil.createStream[String](app, kafkaConfig, 1, "time-replayable-producer").map{ message =>
```
clockfly: Maybe implicit conversion to KafkaDSLUtil?

clockfly:
I think the direction is good.

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/op/OP.scala
huafengw: The sink part is related to HBase Sink, which also need to be refactored.

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/plan/OpTranslator.scala
```diff
@@ -182,39 +179,43 @@ object OpTranslator {
...
 
     override def onStart(startTime: StartTime): Unit = {
+      source.open(taskContext, Some(startTime.startTime))
       self ! Message("start", System.currentTimeMillis())
     }
 
     override def onNext(msg: Message): Unit = {
       val time = System.currentTimeMillis()
-      source.foreach(msg => {
+      source.read(1).foreach(msg => {
```
whjiang: read only 1 each iteration?

whjiang:
Have we removed the dependency of dsl on external_kafka?

examples/streaming/kafka-hbase-pipeline/src/main/scala/org/apache/gearpump/streaming/examples/pipeline/PipeLineDSL.scala
whjiang: I think KafkaDSLUtil is OK for both Java and Scala. For Scala only, whether using implicit conversion or not depends on whether we want consistency between Java API and Scala API.

experiments/dsl/src/main/scala/org/apache/gearpump/streaming/dsl/plan/OpTranslator.scala
huafengw: There supposed to be a batch size configuration, but I'm not sure we should expose this parameter on the ```source``` interface.

huafengw:
Actually what I'm doing is adding dsl dependency to external_kafka.

examples/streaming/kafka-hbase-pipeline/src/main/scala/org/apache/gearpump/streaming/examples/pipeline/PipeLineDSL.scala
kkasravi: If we have other types of source connectors, how do we extend from KafakDSLUtil? It would be nice to have either implicit conversion to a Source API or a type class mechanism where new source connectors can be added without making them adhere to the DataSource API.

manuzhang:
@whjiang Yes.

```
  lazy val dsl = Project(
    id = "gearpump-experiments-dsl",
    base = file("experiments/dsl"),
    settings = commonSettings ++
      Seq(
        libraryDependencies ++= Seq(
          "org.scalaz" %% "scalaz-core" % scalazVersion
        ) ++ hadoopDependency
      )
  ) dependsOn(streaming % "test->test;compile->compile", external_hbase % "test->test;compile->compile")
  
```

examples/streaming/kafka-hbase-pipeline/src/main/scala/org/apache/gearpump/streaming/examples/pipeline/PipeLineDSL.scala
huafengw: The gap here is that the dsl need a typed datasource, however the exsiting DataSource API is not typed.

huafengw:
Any further comment? Please note that:
1. I still choose not to use implicit conversion to KafkaDSLUtil cause I think the implicit way maybe struggle for the newbee.
2. The general data source task does not have a batch size configuration, and I'm considering should we change the ```read(batchsize: Int)``` interface in DataSource.
3. The sink part refactor is related to HBase connector, I'll finish that part in another pull request.

clockfly:
Looks good. 
+1

PR #1193, opened 2015-07-23
fix #1192 Potential message loss exception in ExpressTransport

### LINKED ISSUE #1192: Potential message lost exception in ExpressTransport
The remoteAddressMap in Express should not include localhost

huafengw:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1193)

### DIFF
--- a/streaming/src/main/scala/org/apache/gearpump/streaming/executor/Executor.scala
+++ b/streaming/src/main/scala/org/apache/gearpump/streaming/executor/Executor.scala
@@ -111,13 +111,13 @@ class Executor(executorContext: ExecutorContext, userConf : UserConfig, launcher
 
     case TaskLocations(locations) =>
       LOG.info(s"TaskLocations Ready...")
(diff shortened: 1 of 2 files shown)


streaming/src/main/scala/org/apache/gearpump/streaming/executor/Executor.scala
```diff
@@ -116,8 +116,9 @@ class Executor(executorContext: ExecutorContext, userConf : UserConfig, launcher
         taskIdList.map(taskId => (TaskId.toLong(taskId), host))
       }
       express.startClients(locations.keySet).map { _ =>
```
clockfly: startClients will still connect to local host?

huafengw: Nope, the startClients will also exclude the localhost.

clockfly:
+1

PR #1219, opened 2015-07-27
fix #1210, fix kafka examples

### LINKED ISSUE #1210: examples/streaming/kafka fails to run when submitted in UI
[Story](https://www.pivotaltracker.com/story/show/99844130)
log info will be provided

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1219)

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -170,7 +170,7 @@ object Build extends sbt.Build {
      libraryDependencies <+= (scalaVersion)("org.scala-lang" % "scala-reflect" % _),
      libraryDependencies ++= (
         if (scalaVersion.value.startsWith("2.10")) List("org.scalamacros" %% "quasiquotes" % "2.1.0-M5")
-        else Nil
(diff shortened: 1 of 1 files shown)


manuzhang:
```scala
   List("org.scala-lang.modules" %% "scala-parser-combinators" % "1.0.4")
```
The change is because the scala-parser-combinators library required by kafka but excluded in assembly is separated from scala core in 2.11.

clockfly:
Can we avoid puting this dependency in core list?

manuzhang:
here is a list of scala libraries that are excluded from assembly (sbt-assembly)

```
      "scala-actors",
      "scala-compiler",
      "scala-continuations",
      "scala-library",
      "scala-parser-combinators",
      "scala-reflect",
      "scala-swing",
      "scala-xml"
```

If not provided in core lib, they won't be on classpath and applications fail to run.
If we include them into assembled jar, they may be conflicting with scala libraries already in core lib and cause unpredictable results.

clockfly:
+1

PR #1222, opened 2015-07-27
fix #1217, Remove unnecessary dependencies from project core.

### LINKED ISSUE #1217: Remove unnecessary dependencies for project core
To minimize the classpath conflict with user application, the number of core dependencies (under lib) should be as small as possible.

We should remove dependencies of:
1. upickle.
2, storm.
3. jgrapht
4. commons-io
5. commons-logging
6. commons-lang.

and shade away:
1. guava
2. gs-collection.
3. Coda hale metrics.

We should also be easy to integrate scalajs with existing data type definition. Specifically, 
we should:
1. Refactor Graph to depends on scala library only.
2. Refactor metrics, so that it is easy to serialize/deserialize to/from json.

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1222)


clockfly:
Old review comments can be found at: 
https://github.com/intel-hadoop/gearpump/pull/1218

clockfly:
UT failure FileServcerSpec is random and unrelated.

core/src/main/scala/org/apache/gearpump/metrics/Metrics.scala
```diff
@@ -59,14 +54,23 @@ object Metrics extends ExtensionId[Metrics] with ExtensionIdProvider {
...
-      case gauge: Gauge[_] =>
-        upickle.writeJs(Map("name"-> gauge.name, "value" -> gauge.value.toString))
+    def apply(h: Histogram, c: Counter, m: Meter, t: Timer): MetricType = {
+      val result =
+        if (h != null) h
+        else if (c != null) c
+        else if (m != null) m
+        else if (t != null) t
+        else null
+      result
```
huafengw: Where is this code used, really necessary?

clockfly: Yes. It is necessary to make upickle work.

clockfly: Otherwise, we cannnot pickle the metrics to json.

core/src/main/scala/org/apache/gearpump/util/Graph.scala
```diff
@@ -17,57 +17,52 @@
...
-import org.slf4j.Logger
-
 import scala.collection.JavaConversions._
+import scala.collection.mutable
 import scala.language.implicitConversions
 
 /**
  * Application DAG
  */
+class Graph[N, E](vertexList: List[N], edgeList: List[(N, E, N)]) extends Serializable{
```
huafengw: What's the purpose of droping jgrapht cause we can also shade this library

clockfly: Frontend (scalajs) also need to use the graph implementation. 
If we still depends on jgrapht , then scalajs won't be able to use it.

huafengw:
Should we exclude libraries in storm directory? In Worker.scala, there is a function called filterDaemonLib do the similar thing.

clockfly:
@huafengw 

Yes, we should exclude storm dependency for user application. Let make it another PR.

manuzhang:
@clockfly I checked out the pr branch and built it but found that the storm example is not runnable. Have you verified with an storm example ?

experiments/storm/src/main/scala/org/apache/gearpump/experiments/storm/StormRunner.scala
```diff
@@ -18,44 +18,100 @@
...
 
+import scala.collection.JavaConverters._
 
 object StormRunner extends AkkaApp with ArgumentsParser {
   override val options: Array[(String, CLIOption[Any])] = Array(
-    "storm_topology" -> CLIOption[String]("<storm topology main class>", required = true),
-    "storm_args" -> CLIOption[String]("<storm topology name>", required = false),
-    "storm_config" -> CLIOption[String]("<storm config path>", required = false),
-    "runseconds"-> CLIOption[Int]("<how long to run this example>", required = false, defaultValue = Some(60)))
+    "jar" -> CLIOption[String]("<storm jar>", required = true),
```
manuzhang: this name conflicts with the general `-jar` option. renamed to `storm_jar` ?

clockfly:
@manuzhang 

The storm module error has been fixed.
Command I used to start a storm job:
```
bin\storm -jar C:\myData\gearpump\output\target\pack\storm-starter-0.9.3.jar storm.starter.ExclamationTopology  exclamation
```

experiments/storm/src/main/scala/org/apache/gearpump/experiments/storm/StormRunner.scala
clockfly: We don't use AppSubmitter anymore.

clockfly:
further comments?

manuzhang:
+1

PR #1255, opened 2015-07-29
fix #1254, Add error reporting service and UI

### LINKED ISSUE #1254: Add error reporting service and UI
The error reporting should report:
1. when the error happen.
2. What is the error source. From which task?
3. What is root casue of this error? A stacktrace is good.

With this feature, it is easier to user to troubleshooting what is going wrong directly in UI, without looking into every log file.

clockfly:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1255)


clockfly:
UT failure is not related.

stanleyxu2005:
+1

PR #1260, opened 2015-07-29
fix #339, add Hadoop-compatible CheckpointStore

### LINKED ISSUE #339: add hadoop-compatible storage
kafka offset storage topic is created at runtime and there will be duration of service unavailable (#338) since leader election and metadata propagation takes time. better to have hdfs as offset storage

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1260)


clockfly:
Several questions:
1. For the package output, how many new new are introduced into folder lib?
2. About the rotation, what rotation method do we recommend? Is it necessary to define different rotations?
3. For following code. User are forced to define UserConfig.withValue, it will be great if user don't need to do this.

```
+    val hadoopConfig = new Configuration
+    hadoopConfig.set("fs.defaultFS", "hdfs://localhost:9000")
+    val checkpointStoreFactory = new HadoopCheckpointStoreFactory("MessageCount", hadoopConfig, Rotation(1 day))
     val userConfig = UserConfig.empty		     val userConfig = UserConfig.empty
-      .withValue(PersistentStateConfig.NAME, stateConfig)		+      .withLong(PersistentStateConfig.STATE_CHECKPOINT_INTERVAL_MS, 1000L)
-      .withValue(KafkaCheckpointStore.PRODUCER_CONFIG, KafkaUtil.buildProducerConfig("localhost:9092"))		+      .withValue(PersistentStateConfig.STATE_CHECKPOINT_STORE_FACTORY, checkpointStoreFactory)
-      .withValue(KafkaCheckpointStore.CONSUMER_CONFIG, KafkaUtil.buildConsumerConfig("localhost:2181"))		
 		 
     val partitioner = new HashPartitioner()		     val partitioner = new HashPartitioner()
     val app = StreamApplication("MessageCount", Graph(gen ~ partitioner ~> count), userConfig)		     val app = StreamApplication("MessageCount", Graph(gen ~ partitioner ~> count), userConfig)
```

4. It will be worth exploring whether there are existing open source project, to use HDFS as a versioned key value store, so that we don't need to manage the difficulties.

manuzhang:
1. No new packages except the state jar itself. All other dependencies are assembled into the example-states jar. 
2. Rotation strategy depends on state volume, checkpoint frequency and other factors which we cannot assume.  For example, applications with low checkpoint frequency would favor `FileSizeRotation` and those with frequent small checkpoints may prefer `TimeDurationRotation`. 
3. Users need to set `PersistentStateConfig` through `UserConfig` which I don't think belongs to this story and would improve in future PR on `PersistentState`.
4. I'm been exploring for this library since I started out but fail to find a good candidate. I've also looked at Spark Streaming and Storm and they both implemented their own ways to checkpoint to HDFS, which are kind of combined into this PR.

external/hadoop/src/main/scala/org/apache/gearpump/streaming/hadoop/HadoopCheckpointStoreFactory.scala
```diff
@@ -0,0 +1,67 @@
...
+  override def getCheckpointStore(conf: UserConfig, taskContext: TaskContext): CheckpointStore = {
+    import taskContext.{appId, taskId}
+    val dirPath = new Path(dir, s"app$appId-task${taskId.processorId}_${taskId.index}")
+    val fs = HadoopUtil.getFileSystemForPath(dirPath, hadoopConfig)
+
+    val renameFile = (oldName: String, newName: String) => {
+      val fs = HadoopUtil.getFileSystemForPath(dirPath, hadoopConfig)
+      fs.rename(new Path(dirPath, oldName), new Path(dirPath, newName))
+      newName
+    }
```
huafengw: Where is ```renameFile``` used?

manuzhang: thx. should've removed this

clockfly:
```
 applications with low checkpoint frequency would favor FileSizeRotation and those with frequent small checkpoints may prefer TimeDurationRotation.
```
Why FileSizeRotation is better for low checkpoint frequency? Can the rotation be disabled?

clockfly:
Can we avoid introducing bijection and chill as dependencies?

clockfly:
For example, we can define our own bijection interface, and add an external module to integrate with twitter solutions.

This also applies to Monoid. 

Reducing the classpath dependencies to as small as possible is very important.

manuzhang:
> Why FileSizeRotation is better for low checkpoint frequency? Can the rotation be disabled?

Because we may end up with a lot of small files if using `TimeDurationRotation`.  Users can use `NoRotation` to disable `Rotation`. 

> Can we avoid introducing bijection and chill as dependencies?
Sure, I'll do so.

clockfly:
Is NoRotation prefered? Can we only use NoRatation?

manuzhang:
Rotation reduces time to find checkpoints on recovery. Every checkpoint file is in the format ```checkpoints-$startTime-$endTime.store```. To look up a timestamp, we will only look into the checkpoint file with `startTime <= timestamp <= endTime`, which helps to skip a lot of records. Users can also optionally remove very old rotation files.

clockfly:
So you are using rotation for indexing?

Should we create seprate index file to do this?

manuzhang:
Since streaming application is long running and the size of a checkpoint file will grow up infinitely,  rotation allows user to keep only recent checkpoint files (whose checkpoints older than global checkpoint time could be garbage collected) and shortens the recovery time.

I don't think it would be more efficient using a separate index file since the file will also bloat up and it is impractical to pre-load it into memory for random access. Plus, having two files to worry about in each checkpoint makes things more complicated.

clockfly:
Under which situtation TimeDurationRotation is prefered than FileSizeRotation?  and why?
Under which situation NoRotation should be used?

external/hadoop/src/main/scala/org/apache/gearpump/streaming/hadoop/lib/rotation/package.scala
```diff
@@ -0,0 +1,62 @@
...
+
+package org.apache.gearpump.streaming.hadoop.lib
+
+package object rotation {
+
+
+
+
+  implicit final class FileSizeConversions(size: Int) {
+    def B = new BSize(size)
```
clockfly: FileSizeConversions Used anywhere?

clockfly:
1. I will suggest to remove NoRation, and TimeDurationRotation. 
It doesn't make sense to me to define three Rotations in source code.
2. FileSizeConversions looks tricky to me, I will suggest to remove this.
3. the new project external_hadoop looks confusing to me. It doesn't looks like related with hadoop. 
4. Maintain our own checkpoint and recovery codebase is very fragile. For example, for the long run, there will be more and more requirements for this checkpoint, like:
  a. a dashboard to view and manage all checkpoints.
  b. Clean certain checkpoints, or clean by timestamp.
  c. Mapreduce reader
  d. version Upgrade
We can roll our checkpoint implementation for now, but keep looking for existing solution from other open source projects will still be prefered.

PR #1280, opened 2015-07-31
Fixes #1278 split services into 2 projects for scala.js

### LINKED ISSUE #1278: split services into 2 projects for scala.js
[Story](https://www.pivotaltracker.com/story/show/100222254)
Create the related subprojects so scala types can be generated for the client

kkasravi:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1280)


huafengw:
@kkasravi Hi Kam, what is ```copyJSArtifactsToOutput``` doing? Seems it is complicated, any other way to accomplish this?

services/js/src/main/scala/org/apache/gearpump/dashboard/services/RestApiService.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  def appConfigLink(appId: String): String = {
+    options.conf.restapiRoot + "/config/app/" + appId
+  }
+
+  def workerConfigLink(workerId: String): String = {
+    options.conf.restapiRoot + "/config/worker/" + workerId
+  }
+
+  def masterConfigLink: String = {
```
clockfly: Hi, Kam. 

There is update in the rest path. 
options.conf.restapiRoot + "/master/config

services/js/src/main/scala/org/apache/gearpump/dashboard/services/RestApiService.scala
```diff
@@ -0,0 +1,82 @@
...
+    val url = options.conf.restapiRoot + "/appmaster/" + appId
+    http.delete(url)
+  }
+
+  def appConfigLink(appId: String): String = {
+    options.conf.restapiRoot + "/config/app/" + appId
+  }
+
+  def workerConfigLink(workerId: String): String = {
+    options.conf.restapiRoot + "/config/worker/" + workerId
```
clockfly: options.conf.restapiRoot + "/worker/config

services/js/src/main/scala/org/apache/gearpump/dashboard/services/RestApiService.scala
```diff
@@ -0,0 +1,82 @@
...
+    future.map(_.toString)
+  }
+
+  def killApp(appId: String): HttpPromise = {
+    val url = options.conf.restapiRoot + "/appmaster/" + appId
+    http.delete(url)
+  }
+
+  def appConfigLink(appId: String): String = {
+    options.conf.restapiRoot + "/config/app/" + appId
```
clockfly: options.conf.restapiRoot + "/appmaster/1/config

clockfly:
@kkasravi 
About copyJSArtifactsToOutput, you can refer the code implementation of how we copy example jar into the output.
Please check https://github.com/kkasravi/gearpump/blob/issue_%231278/project/Build.scala#L250

project/Build.scala
```diff
@@ -411,37 +417,126 @@ object Build extends sbt.Build {
...
+  lazy val servicesjvm = Project(id = "gearpump-services-jvm", base = file("services/jvm")).
+    settings(jvmSettings : _*).
+    dependsOn(streaming % "test->test;compile->compile", daemon % "test->test;compile->compile")
+
+  lazy val servicesjs = Project(id = "gearpump-services-js", base = file("services/js")).
+    enablePlugins(ScalaJSPlugin).
+    settings(commonSettings: _*).
+    settings(jsSettings : _*)
+
+  def findDirs(currentDir: String): Array[File] = {
```
manuzhang: How about putting these scripts into a separated file such that our build file would look much cleaner ?

services/js/src/main/scala/org/apache/gearpump/dashboard/services/RestApiService.scala
kkasravi: RestApiService isn't active right now but i'll update the Uri's to match what's in restapi.js

project/Build.scala
kkasravi: @clockfly 
I'll update to use baseDirectory.value / ..
@huafengw 
CopyJSArtifacts needs to be called as part of fastOptJS actions so that ~fastOptJS works. 
```
    fastOptJS in Compile := {
      val originalResult = (fastOptJS in Compile).value
      copyJSArtifactsToOutput
      originalResult
    },
```
This mean copying specific files to output dir as well as modifying the .map file so scala debug works correctly in a browser. These actions will be triggered when any file is updated.

@manuzhang 
I can put some of the utility functions in a file similar to what has been done in PhantomJS2Env.scala.

kkasravi:
build failures are unrelated
```
[info] Worker
[info] - should register worker address to master when started. *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (25 seconds) during expectMsg while waiting for RegisterNewWorker
[info]   at scala.Predef$.assert(Predef.scala:165)
[info]   at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:338)
[info]   at akka.testkit.TestKitBase$class.expectMsg(TestKit.scala:324)
[info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:718)
[info]   at org.apache.gearpump.cluster.main.MainSpec$$anonfun$1.apply$mcV$sp(MainSpec.scala:68)
[info]   at org.apache.gearpump.cluster.main.MainSpec$$anonfun$1.apply(MainSpec.scala:54)
[info]   at org.apache.gearpump.cluster.main.MainSpec$$anonfun$1.apply(MainSpec.scala:54)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
```

```
[info] Exception encountered when attempting to run a suite with class name: org.apache.gearpump.cluster.master.MasterHAServiceSpec *** ABORTED ***
[info]   akka.ConfigurationException: Could not start logger due to [akka.ConfigurationException: Logger specified in config can't be loaded [akka.event.slf4j.Slf4jLogger] due to [akka.event.Logging$LoggerInitializationException: Logger log1-Slf4jLogger did not respond with LoggerInitialized, sent instead [TIMEOUT]]]
[info]   at akka.event.LoggingBus$class.startDefaultLoggers(Logging.scala:143)
[info]   at akka.event.EventStream.startDefaultLoggers(EventStream.scala:26)
[info]   at akka.actor.LocalActorRefProvider.init(ActorRefProvider.scala:622)
[info]   at akka.remote.RemoteActorRefProvider.init(RemoteActorRefProvider.scala:157)
[info]   at akka.cluster.ClusterActorRefProvider.init(ClusterActorRefProvider.scala:58)
[info]   at akka.actor.ActorSystemImpl.liftedTree2$1(ActorSystem.scala:618)
[info]   at akka.actor.ActorSystemImpl._start$lzycompute(ActorSystem.scala:615)
[info]   at akka.actor.ActorSystemImpl._start(ActorSystem.scala:615)
[info]   at akka.actor.ActorSystemImpl.start(ActorSystem.scala:632)
[info]   at akka.actor.ActorSystem$.apply(ActorSystem.scala:141)
```

```
[info] - should kill itself when can not get a quorum *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (5 seconds) during expectMsg: Terminated TestActor[akka://ForMasterWatcher/user/$$a]
[info]   at scala.Predef$.assert(Predef.scala:165)
[info]   at akka.testkit.TestKitBase$class.expectMsgPF(TestKit.scala:356)
[info]   at akka.testkit.TestKit.expectMsgPF(TestKit.scala:718)
[info]   at akka.testkit.TestKitBase$class.expectTerminated(TestKit.scala:368)
[info]   at akka.testkit.TestKit.expectTerminated(TestKit.scala:718)
[info]   at org.apache.gearpump.cluster.main.MasterWatcherSpec$$anonfun$1.apply$mcV$sp(MasterWatcherSpec.scala:46)
[info]   at org.apache.gearpump.cluster.main.MasterWatcherSpec$$anonfun$1.apply(MasterWatcherSpec.scala:40)
[info]   at org.apache.gearpump.cluster.main.MasterWatcherSpec$$anonfun$1.apply(MasterWatcherSpec.scala:40)
[info]   ...
```

project/Util.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  def copyJSArtifactsToOutput: Unit = {
+    ConsoleLogger().info(s"copying JS artifacts ${Build.servicesjs.base.getPath}")
+    val in = file(s"${Build.servicesjs.base.getPath}/target/scala-2.11/gearpump-services-js-fastopt.js.map")
+    val out = file(s"${Build.distDirectory}/target/pack/dashboard/gearpump-services-js-fastopt.js.map")
+    val input = IO.read(in)
+    val data = input.replaceAll("../../src","src")
+    IO.write(out, data)
+    IO.copyFile(
+      file(s"${Build.servicesjs.base.getPath}/target/scala-2.11/gearpump-services-js-fastopt.js"),
```
stanleyxu2005: Is it possible to externalize scala-2.11, it can be changed any time.

project/Util.scala
```diff
@@ -0,0 +1,60 @@
...
+      file(s"${Build.distDirectory}/target/pack/dashboard/gearpump-services-js-fastopt.js")
+    )
+    IO.copyDirectory(
+      file(s"${Build.servicesjs.base.getPath}/src"),
+      file(s"${Build.distDirectory}/target/pack/dashboard"),
+      true
+    )
+    IO.copyDirectory(
+      file(Build.servicesjs.base.getPath),
+      file(Build.distDirectory) / "target" / "pack" / "dashboard",
```
stanleyxu2005: `file(Build.distDirectory) / "target" / "pack" / "dashboard"` appears several times. Maybe externalize as a path object for these common paths.

project/Util.scala
clockfly: Can we use the same way as 
https://github.com/intel-hadoop/gearpump/blob/master/project/Pack.scala#L57

?

project/Util.scala
```diff
@@ -0,0 +1,60 @@
...
+    val out = file(s"${Build.distDirectory}/target/pack/dashboard/gearpump-services-js-fastopt.js.map")
+    val input = IO.read(in)
+    val data = input.replaceAll("../../src","src")
+    IO.write(out, data)
+    IO.copyFile(
+      file(s"${Build.servicesjs.base.getPath}/target/scala-2.11/gearpump-services-js-fastopt.js"),
+      file(s"${Build.distDirectory}/target/pack/dashboard/gearpump-services-js-fastopt.js")
+    )
+    IO.copyDirectory(
+      file(s"${Build.servicesjs.base.getPath}/src"),
```
clockfly: This will copy the scala file to distribution package? 
I don't think it is right.

project/Util.scala
```diff
@@ -0,0 +1,60 @@
...
+      file(s"${Build.distDirectory}/target/pack/dashboard"),
+      true
+    )
+    IO.copyDirectory(
+      file(Build.servicesjs.base.getPath),
+      file(Build.distDirectory) / "target" / "pack" / "dashboard",
+      true
+    )
+    val htmlFiles = findFiles(file(s"${Build.services.base.getPath}/dashboard"),
```
clockfly: This function duplicate of https://github.com/intel-hadoop/gearpump/blob/master/project/Pack.scala#L57

?

project/build.properties
```diff
@@ -16,4 +16,4 @@
 # limitations under the License.
 #
 
-sbt.version=0.13.5
+sbt.version=0.13.7
```
clockfly: Why we update sbt version?

project/Build.scala
```diff
@@ -46,10 +49,12 @@ object Build extends sbt.Build {
   val algebirdVersion = "0.9.0"
   val chillVersion = "0.6.0"
 
+  val distDirectory = "output"
+
   override def projects: Seq[Project] = (super.projects.toList ++ BuildExample.projects.toList
       ++ Pack.projects.toList).toSeq
 
-  val commonSettings = Defaults.defaultSettings ++ Seq(jacoco.settings:_*) ++ sonatypeSettings  ++ net.virtualvoid.sbt.graph.Plugin.graphSettings ++
+  val commonSettings = Seq(jacoco.settings:_*) ++ sonatypeSettings  ++ net.virtualvoid.sbt.graph.Plugin.graphSettings ++
```
clockfly: Why we remove Defaults.defaultSettings?

project/Build.scala
```diff
@@ -223,36 +228,76 @@ object Build extends sbt.Build {
...
-          "org.webjars" % "select2" % "3.5.2",
-          "org.webjars" % "select2-bootstrap-css" % "1.4.6",
-          "org.webjars.bower" % "ng-file-upload" % "5.0.9",
-          "org.webjars.bower" % "vis" % "4.5.1"
-        ).map(_.exclude("org.scalamacros", "quasiquotes_2.10")).map(_.exclude("org.scalamacros", "quasiquotes_2.10.3"))
-      )
-  ) dependsOn(streaming % "test->test; compile->compile", daemon % "test->test; compile->compile")
+  lazy val services = Project(id = "gearpump-services", base = file("services")).
+    settings(commonSettings: _*).
+    aggregate(servicesjs, servicesjvm).
```
clockfly: Can we enable scalajs cross-project? 
Seems now we are treating jvm and js as two seperate projects.

As I think there will be some scala files shared between js project and jvm project.

project/Build.scala
```diff
@@ -223,36 +228,76 @@ object Build extends sbt.Build {
...
+  )
+
+  lazy val jsSettings = Seq(
+    scalaVersion := scalaVersionNumber,
+    checksums := Seq(""),
+    requiresDOM := true,
+    resolvers += "Sonatype OSS Snapshots" at "https://oss.sonatype.org/content/repositories/snapshots",
+    libraryDependencies ++= Seq(
+      "org.scala-js" %%% "scalajs-dom" % "0.8.1",
+      "com.greencatsoft" %%% "scalajs-angular" % "0.5-SNAPSHOT",
```
clockfly: Can we avoid using SNAPSHOT version?

services/js/src/main/scala/org/apache/gearpump/dashboard/DashboardApp.scala
```diff
@@ -0,0 +1,54 @@
...
+package org.apache.gearpump.dashboard
+
+import com.greencatsoft.angularjs.Angular
+import org.apache.gearpump.dashboard.services._
+
+import scala.scalajs.js.JSApp
+import scala.scalajs.js.annotation.JSExport
+
+@JSExport
+object DashboardApp extends JSApp {
```
stanleyxu2005: So scalajs is not included as a special js library? It seems later every piece of view/control has to be written in scala? Could you hold off this file be committed, until we make it clear?

project/Util.scala
kkasravi: It is correct - it shows up correctly in the browsers debug window.

project/Util.scala
kkasravi: The problem is we need to invoke the copy prior to the compile step, i don't think the pack replacement will do that.

project/build.properties
kkasravi: This is the minimum version that scalajs projects were using, i ran into problems with 0.13.5.

project/Build.scala
kkasravi: See this thread. https://github.com/scala-js/scala-js/issues/1450

project/Build.scala
kkasravi: Cross-project did not work entirely, there were a number of problems mostly around cross-project expects one jvm project and one js project. We have many jvm projects.

services/js/src/main/scala/org/apache/gearpump/dashboard/DashboardApp.scala
kkasravi: This is required for fastopt to be generated (it looks for a main). Note this does not replace index.html or restapi.js or anything else at this time.

project/Util.scala
kkasravi: For the .map files we actually need to edit prior to copy.

services/js/src/main/scala/org/apache/gearpump/dashboard/DashboardApp.scala
kkasravi: No requirement to write all view/control in scala. Old DashboardApp.scalas from #952 included many external controllers/views as pure js.

project/Build.scala
clockfly: Can we enable cross-project just for project services?

services/js/src/main/scala/org/apache/gearpump/dashboard/DashboardApp.scala
clockfly: Maybe just extends JSApp without introducing greencatsoft angular dependencies for now?

We can add the greencatsoft as a new seperate PR.

project/Build.scala
clockfly: The thread you mentioned is about jetty, do we encounter same problem here?

project/Build.scala
kkasravi: My recollection was no. Also I think a number of things would need to be moved under dashboard to js.

services/js/src/main/scala/org/apache/gearpump/dashboard/DashboardApp.scala
kkasravi: I think that's possible but we should then also remove it from Build.scala as well as remove RestApiService and ConfService.scala. Is there a reason to keep all angular dependencies separate? Perhaps only introduce angular dependencies where they have actually replaced the js equivalents? I think angular dependencies would probably be introduced in stages anyway - several controllers / services at a time.

kkasravi: For services like RestApiService, ConfService. I think we would add the service and then begin to use the service from either javascript or scala(js). That is the intent here - to first introduce the services and then begin to use them.

project/Build.scala
kkasravi: Yes we will get an error about jetty. See in the thread reference above the following comments

screen shot 2015-08-02 at 11 30 32 pm

project/Build.scala
kkasravi: If we are copying some files from core, some from streaming, to servicesjs this is not easy to do in cross-compile. My recollection is that cross-compile wants a shared project where these files live - but this would necessitate moving types out of core/.../ClusterMessage.scala etc to a shared project.

kkasravi:
Note build failures are not related
```
[info] Passed: Total 24, Failed 0, Errors 0, Passed 24
[error] (gearpump-daemon/test:test) sbt.TestsFailedException: Tests unsuccessful
[error] (gearpump-streaming/test:test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 1187 s, completed Aug 3, 2015 6:47:28 AM
```

services/js/src/main/scala/org/apache/gearpump/dashboard/DashboardApp.scala
clockfly: @kkasravi 

The RestService now depends on greencatsoft HttpService. I think for this point, we should avoid introducing any greencatsoft Angular stuff, until the discussion is clear.

For RestApiService, you can choose to use XmlHttpRequest to do ajax call. 
Or 
http://lihaoyi.github.io/hands-on-scala-js/#dom.extensions

clockfly: So I will suggest create a dummy JsApp for now, and strip all dpenendencies about greencatsoft .

If you'd like to introduce RestApiService, then using avoid using greencatsoft.HttpService is prefered.

project/Build.scala
clockfly: I wonder why seperate jvm/js is possible, but make it cross-project is not possible. 

IMHO, cross-project is no more than adding a common shared source folder services/shared. This allow us to define some shared class in project services/. 

For message definition in core or streaming,  you can still choose to copy them to js/ subfolder.

services/js/src/main/scala/org/apache/gearpump/dashboard/DashboardApp.scala
clockfly: Here is an example about how to do ajax call with upickle
http://www.wayofquality.de/scala/play/scalajs/reactjs/castillo-04-Initial-Webclient/

kkasravi: Angular HttpService wraps the JavaScript $http code which is what restapi.js calls. Then shouldn't restapi.js also use XmlHttpRequest? Why are we forcing non angular use for RestApiService.scala but are ok with $http for restapi.js? This may introduce many subtle bugs.

clockfly: Frankly speaking, scala binding for angular is not that promsing. Until recently(1 day ago), it has a big change of the HttpService interface. 
https://github.com/greencatsoft/scalajs-angular/pull/50/files

Depending on this changing library is riskly. 

I can see your argument that restapi.js depends on javascript version of angular. Then  in that case, RestApiService.scala should not be incldued until the discussion is clear. 

Maybe for now the scalajs part should only expose the serializer and deserializer which is implemented with upickle.

kkasravi: Agreed, there is some risk on SNAPSHOT's. What issues are relevant to the discussion? Is the discussion whether to use scalajs-angular at all or whether xmlhttprequest should be always used regardless of whether we're using angular, reactive, etc? Changing to XmlHttpRequest and then changing back to $http is not a very efficient use of time if the discussion can be resolved without doing this. Perhaps we need to ask the author of scalajs-angular when he plans on 0.5 or are there other concerns with using HttpService or $http?

project/Build.scala
kkasravi: I will revisit why cross-project did not work and provide the reasons here.

kkasravi: For cross-project there are problems getting fastOptJS and fullOptJS to work correctly, including getting the correct files copied to output/. I don't think cross-project works well with the pack related  tasks since pack tasks are invoked after compile. I've built a cross-compile type of project under my fork issue_#1278_1. Still working on it to see if I can get the desired behavior.

kkasravi:
Removed RestApiService, ConfService, scalajs-angular references.

services/js/src/main/scala/org/apache/gearpump/dashboard/DashboardApp.scala
clockfly: @kkasravi 

Ok, here is my points:
1. For now, remove angular stuff completely from scalajs.
2. In scalajs, only expose two util which do serialization and deserialization with upickle. Neither XmlHttpRequest or angular HttpService is used.

project/Util.scala
```diff
@@ -0,0 +1,60 @@
...
+      IO.listFiles(dir, pattern)
+    })
+  }
+
+  def copyJSArtifactsToOutput: Unit = {
+    ConsoleLogger().info(s"copying JS artifacts ${Build.servicesjs.base.getPath}")
+    val in = file(s"${Build.servicesjs.base.getPath}/target/${Build.scalaVersionMajor}/gearpump-services-js-fastopt.js.map")
+    val out = file(s"${Build.distDashboardDirectory}/gearpump-services-js-fastopt.js.map")
+    val input = IO.read(in)
+    val data = input.replaceAll("../../src","src")
```
clockfly: I see here you repalce "../../src" with "src". This replacement is used for source mapping for debugging?

If that is the case, I suggest removing this part.
1. In the distributation package, there should be no .scala source files.

kkasravi: What target should we use to allow debugging?

services/js/src/main/scala/org/apache/gearpump/dashboard/DashboardApp.scala
kkasravi: All angular and all scalajs files have been removed. I don't believe any util's are required for upickle unless they provide handling of ActorRef or Config which are types referenced in ./core/src/main/scala/org/apache/gearpump/cluster/ClusterMessage.scala.

kkasravi:
Note build failures are unrelated
```
[info] Passed: Total 24, Failed 0, Errors 0, Passed 24
[error] (gearpump-core/test:test) sbt.TestsFailedException: Tests unsuccessful
[error] (gearpump-daemon/test:test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 548 s, completed Aug 4, 2015 2:18:08 AM
```

project/Util.scala
kkasravi: For distribution if we remove the ability to step through scala code, similar functionality would be retained by any pure javascript solution unless all js code is minified and obfuscated. We need some kind of debug target that allows source maps, scala and non-minified js. I'm not sure if pack provides for something like this.

clockfly: Here is my proposal:
1. remove this function copyJSArtifactsToOutput.
2. In sbt/pack target, copy gearpump-services-js-fastopt.js and gearpump-services-js-fastopt.js.map to target\pack\dashboard\views\scalajs\
3. If developer want to debug, they can copy services/js/src/  to target/pack/dasboard/

Here is my reasoning:
1. no scala source file sould exist in final package
2. Maintianing stuff like findDirs, findFile, or copyJSArtifactsToOutput will complicate the build script.

project/Build.scala
clockfly: I think cross-project shall work. For cross-project, there are still project created.
One is services.js, the other is services.jvm.

In your current code, you are operating on project servicesjs. After enabling scalajs, the only difference is that you are operating on project "service.js" (note the additonal dot ".")

kkasravi: Correct. But I did not get fastOptJS to work. Having serviceJVM depend on fastOptJS in serviceJS results in a recursion error. The result is that fastOptJS is never invoked.

project/Util.scala
kkasravi: Ok, this will likely break ~fastOptJS, so files after editing must be manually copied to output.

project/Build.scala
clockfly: Yes, I know there are difficulties.

I think project services will need to share some class between servicejs and servicejvm. So cross-project is a must-have feature.

Let's find way to solve the recursion error you mentioned.

project/Util.scala
clockfly: Have you tried artifactPath?
http://stackoverflow.com/questions/25990605/how-to-add-the-output-path-of-javascripts-from-scala-js-to-play

kkasravi: artifactPath looks promising. Thanks! I will followup with sjrd in scala.js gitter room if needed.

kkasravi:
Resubmitting on cleaned up branch using crossProject

PR #1291, opened 2015-08-03
fix #1290, purify state dependencies

### LINKED ISSUE #1290: state should not introduce extra dependencies into core lib

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1291)


clockfly:
Do we need to differentiate system api and user api?

To me, Monoid, Serializer, and PersistentTask are both user api.

clockfly:
Looks good!

+1

Issue #1197, opened 2015-07-23
Gearpump Storage Framework


whjiang:
In general, a Gearpump application requires following storage support:
1. Jar-file storage to store the application jar file(s).
2. application log. Currently we store logs in each node which makes application log analysis difficult.
3. application metrics. 
4. application configuration.
5. data source offset store (for at-least once semantics of streaming application)
6. application state checkpoint store (for transaction semantics)

The general idea is:
1. Provide a storage system satisfied the above requirements.
2. Assume this storage is highly available. That means, it is user's duty to provide such kind of a storage. For test purpose, user can use some non-HA storage system. But, in product use, it shall be HAed.
3. Isolate usage from implementation. That is, Gearpump doesn't rely on Hadoop-common or HDFS or one specific implementation to provide such storage. User is free to implement its own storage.
4. This is a daemon provided functionality and can be used by every Gearpump application.
5. This storage shall provide data retentation functionality and access control.
6. This storage provides a set of API to meet the above requirements instead of one low-level API.
7. User can override the system setting to provide dedicated implementation for certain sub-storage system, e.g. chekcpoint store.
8. Akka replication shall store minimal info for an application and leave the majority to this storage system. I.e. akka replication is more like a seed to this storage system.
9. In release, each storage implementation (e.g. storage-hdfs) is a standalone module/artifact.

The draft of this storage looks like (quite initial, tentative to change):
```
trait Storage {
	def createAppStorage(AppName, AppId) : AppStorage
	def getAppStorage(AppId) : Option[AppStorage]
}

trait AppStorage {
	def open
	def close
	def getJarStore: JarStore
	def getMetricsStore: AppMetricsStore
	def getKVStore: KVStore
	def getLogAppender: LogAppender
	def getConfiguration(ProcessorId): UserConfig
	def setConfiguration(ProcessorId, UserConfig)
}

trait JarStore {
	def copyFromLocal(localPath, remotePath)
	def copyToLocal(remotePath, localPath)
}

///assume K is sortable
trait KVStore[K,V] {
	def append(key, value)
	def read(key): Try[Option[V]]
}
```

kkasravi:
Perhaps the different types of storage could share a common API or aspects of collections so AppStorage doesn't need to change every time a new type of storage is added. Also at least with JarStore see #1204 where lack of streaming causes Master to exit. 

Should AppStorage contain write, read methods?

whjiang:
[...]

PR #1293, opened 2015-08-03
implement #1197 gearpump storage framework

whjiang:
This is not a full PR. Just a review request for the API.

[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1293)

### DIFF
--- a/storage/storage-hdfs/src/main/scala/org/apache/gearpump/appstorage/hadoopfs/rotation/TimeDurationRotation.scala
+++ b/storage/storage-hdfs/src/main/scala/org/apache/gearpump/appstorage/hadoopfs/rotation/TimeDurationRotation.scala
@@ -1,0 +1,49 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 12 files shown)


clockfly:
Can you rebase on trunk?

storage/storage-hdfs/src/main/scala/org/apache/gearpump/appstorage/hadoopfs/rotation/TimeDurationRotation.scala
```diff
@@ -0,0 +1,49 @@
...
+import scala.concurrent.duration.Duration
+
+class TimeDurationRotation(duration: Duration) extends Rotation {
+
+  private[rotation] val maxDuration = duration.toMillis
+  private var startTime: Option[TimeStamp] = None
+  private var curTime: Option[TimeStamp] = None
+
+  override def mark(timestamp: TimeStamp, offset: Long): Unit = {
+    if (startTime.isEmpty) {
```
stanleyxu2005: Line 47 indicates that the duration will NOT be restarted by calling `rotate` but by calling `mark`. It means you have to call `mark` after `rotate`, otherwise the duration will be longer than the actual value. Is it by design or have I missed something?

manuzhang: this is no longer used.

PR #1295, opened 2015-08-04
Fixes #1278 split services into 2 projects for scala.js

### LINKED ISSUE #1278: split services into 2 projects for scala.js

kkasravi:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1295)


project/Build.scala
```diff
@@ -223,36 +229,79 @@ object Build extends sbt.Build {
...
+    jvmSettings(jvmSettings: _*). 
+    jsSettings(jsSettings: _*)
+
+  lazy val js2jvmSettings = Seq(fastOptJS, fullOptJS, packageJSDependencies) map { packageJSKey =>
+    crossTarget in(serviceJS, Compile, packageJSKey) := file(distDashboardDirectory)
+  }
+
+  lazy val serviceJVM: Project = service.jvm.
+    settings(js2jvmSettings: _*).
+    settings(compile in Compile <<= (compile in Compile) dependsOn (fastOptJS in (serviceJS, Compile))).
```
clockfly: Do we need serviceJVm depends on serviceJS?
How about using examples in http://www.scala-js.org/doc/sbt/cross-building.html?

clockfly: To my understanding, crossProject is a util which help us to define the source folder of each project.

Take following example:
```
name := "Foo root project"

lazy val root = project.in(file(".")).
  aggregate(fooJS, fooJVM).
  settings(
    publish := {},
    publishLocal := {}
  )

lazy val foo = crossProject.in(file(".")).
  settings(
    name := "foo",
    version := "0.1-SNAPSHOT",
    scalaVersion := "2.11.5"
  ).
  jvmSettings(
    // Add JVM-specific settings here
  ).
  jsSettings(
    // Add JS-specific settings here
  )

lazy val fooJVM = foo.jvm
lazy val fooJS = foo.js
```

What is meanful is project fooJvm and fooJs, project foo is only a Util to generate settings for fooJvm and fooJs.

If we respect this rule, then there are several suggestions:
1). for project lazy val services, since now it only becomes a util, it should be referenced anywhere in Build.sbt. We should only reference serviceJs or serviceJVm.

So the following code doesn't make sense to me
```
  lazy val services = Project(id = "gearpump-services", base = file("services")).
    aggregate(serviceJS, serviceJVM).

// All lines below doesn't make sense to exist.
    settings(commonSettings: _*).
    settings(compile in Compile <<= (compile in Compile) dependsOn (compile in (serviceJVM, Compile))).
    dependsOn(serviceJVM % "test->test;compile->compile", 
      streaming % "test->test;compile->compile", 
      daemon % "test->test;compile->compile"
    )
```

clockfly:
@kkasravi 

here is an example about how to define custom output path of fastOpts output.
http://stackoverflow.com/questions/29371654/scala-js-compilation-destination

project/Build.scala
kkasravi: There are 2 projects services and service. The latter is defined as 
```
  lazy val service = crossProject.in(file("services")).
    settings(commonSettings: _*).
    jvmSettings(jvmSettings: _*).
    jsSettings(jsSettings: _*)
```
and actually isn't referenced anywhere else.

The former 'services' is a project whose compilation depends on the compilation in serviceJVM.
The serviceJVM project's compilation in turn depends on fastOptJS in serviceJS.

clockfly: Do we still need project services? We already have servicesJVm. 
Other projects should depends on servicesJVm directly.

kkasravi: Not having serviceJVM compilation depend on serviceJS will not generate the fastOptJS step. See http://stackoverflow.com/questions/23023026/how-to-make-scalatra-compile-task-depend-on-scala-js-packagejs-task-in-sbt where sjrd explains how to specify a task dependency (not a project dependency).

project/Build.scala
```diff
@@ -223,36 +229,79 @@ object Build extends sbt.Build {
...
+    dependsOn(streaming % "test->test;compile->compile", daemon % "test->test;compile->compile")
+
+  lazy val serviceJS: Project = service.js.
+    dependsOn(streaming % "test->test;compile->compile", daemon % "test->test;compile->compile")
```
clockfly: Why serviceJS depends on streaming?

project/Build.scala
clockfly: Yes, we need project serviceJVM, but don't need project services.

project/Build.scala
kkasravi: I do redefine the 'artifactPath' and used code from http://stackoverflow.com/questions/29371654/scala-js-compilation-destination eg
```
  lazy val js2jvmSettings = Seq(fastOptJS, fullOptJS, packageJSDependencies) map { packageJSKey =>
    crossTarget in(serviceJS, Compile, packageJSKey) := file(distDashboardDirectory)
  }
```

I think your correct there probably isn't a project dependency of serviceJS on streaming unless we're copying common type definitions from there - although we would probably copy into the shared directory anyway.

kkasravi:
Looks like there is an error in Pack.scala about an ambiguous reference to dsl (imported from both sbt._ and Build._). I'll correct.

kkasravi:
2 references that I've borrowed from are the following:

https://github.com/vmunier/play-with-scalajs-example/blob/master/build.sbt
https://github.com/ochrons/scalajs-spa-tutorial/blob/master/project/build.scala

Both do a fair amount with IO.copy, etc - the former doing it in the Play plugin to copy files from the shared directory. Both use crossProject.

project/Build.scala
kkasravi: Yes, I think we could probably remove services entirely. Let me try that change.

kkasravi: It would result in 'serviceJVM, serviceJS' in place of services where we do things like aggregate - in root project for example.

clockfly: Yes. we want serviceJVM, servieJs in place of services. 
When we do aggregate, we should aggregrate serviceJVM.

project/Build.scala
clockfly: I am thinking whether using artifactPath directly is easier.

clockfly: Like here:
http://stackoverflow.com/questions/29371654/scala-js-compilation-destination

project/Build.scala
kkasravi: I tried removing services and replacing it with serviceJVM in places like Pack.scala. It resulted in no fastOptJS step which prevent these files from being generated
```
./output/target/pack/dashboard/service-fastopt.js
./output/target/pack/dashboard/service-fastopt.js.map
```

clockfly:
@kkasravi 
This takes us a lot of time. I will submit a fix PR as followup.

+1

kkasravi:
@clockfly for the choice of either doing
```
artifactPath in fastOptJS :=
  (resourceManaged in jvm in Compile).value /
    ((moduleName in fastOptJS).value + "-fastopt.js"))
```
or
```
  lazy val js2jvmSettings = Seq(fastOptJS, fullOptJS, packageJSDependencies) map { packageJSKey =>
    crossTarget in(serviceJS, Compile, packageJSKey) := file(distDashboardDirectory)
  }
```
The later takes care of all steps fastOptJS, fullOptJS and packageJSDependencies.

kkasravi:
Yes - if you see better alternative please make the required changes. Thanks for detailed review.

PR #1300, opened 2015-08-06
fix #339, add Hadoop-compatible CheckpointStore

### LINKED ISSUE #339: add hadoop-compatible storage

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1300)


clockfly:
+1

PR #1174, opened 2015-07-13
fix #1013, track task checkpoint time at ClockService

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/intel-hadoop/gearpump/1174)

PR #1306, opened 2015-08-11
fix #1013, track task checkpoint time at ClockService

### LINKED ISSUE #1013: track task checkpoint time at ClockService

manuzhang:


manuzhang:
Previously on #1174, 

> There may be some problems when message ReportCheckpointClock is lost during transfer.

now, the global checkpoint time is only advanced to a value that is reported by all `PersistentTask`s. Say Task A and Task B both reports timestamp 100 and global checkpoint time is set to 100. Then Task A reports 200 but that of Task B is lost. In that case, the global checkpoint time remains at 100. It will only advance when a newer checkpoint time is both reported by Task A and Task B.

clockfly:
Is MinClock still used?

clockfly:
travis failed.

manuzhang:
1. related tests are fixed
2. minClock is used for at-most-once and at-least-once applications and also for dashboard.

clockfly:
+1

PR #1307, opened 2015-08-11
fix #1305 remove JarFileContainer

### LINKED ISSUE #1305: change the JarStore implementaion to remove JarFileContainer

huafengw:


core/src/main/scala/org/apache/gearpump/jarstore/JarStoreService.scala
```diff
@@ -0,0 +1,73 @@
...
+  def copyToLocalFile(localFile: File, remotePath: FilePath)
+}
+
+object JarStoreService {
+  lazy val jarstoreServices: List[JarStoreService] = {
+    ServiceLoader.load(classOf[JarStoreService]).asScala.toList
+  }
+
+  def getJarStoreService(rootPath: String): JarStoreService = {
+    jarstoreServices.find(_.matchScheme(rootPath)).get
```
manuzhang: what if no match is found

huafengw: An exception is expected.

manuzhang: but it will be difficult to find the cause without looking at the source code if we don't provide any help info

huafengw:
updated, please help to review

core/src/main/scala/org/apache/gearpump/cluster/client/ClientContext.scala
```diff
@@ -49,9 +48,12 @@ class ClientContext(config: Config, sys:Option[ActorSystem], _master: Option[Act
...
   LOG.info(s"Starting system ${system.name}")
 
-  import system.dispatcher
-
   private val master = _master.getOrElse(system.actorOf(MasterProxy.props(masters), s"masterproxy${system.name}"))
+  private val jarStoreService = JarStoreService.get(config)
+  jarStoreService match {
+    case needConfig: ConfigRequired => needConfig.init(config)
+    case needSystem: ActorSystemRequired => needSystem.init(system)
+  }
```
manuzhang: really cannot relate a thing `JarStoreService` to a property `xxxRequired`.
`jarStoreService.isConfigRequired` feels more intuitive to me.

clockfly: It is a trait, not a method.

clockfly: What if a  service require both ActorSystem and Config?

case match will only do a partial match.

huafengw: config can be fetched by system.settings.config

clockfly:
Please rebase on trunk

daemon/src/main/scala/org/apache/gearpump/jarstore/local/LocalJarStoreService.scala
```diff
@@ -0,0 +1,90 @@
...
+      client.save(address.url + remotePath, bytes)
+    }(system.dispatcher)
+    Await.ready(future, Duration(15, TimeUnit.SECONDS))
+  }
+
+  /**
+   * Close the JarStoreService
+   */
+  override def close(): Unit = {
+    system.shutdown()
```
clockfly: Why it need to close the system, since it doesn't create the system.

clockfly:
+1

clockfly:
merged

PR #1309, opened 2015-08-12
fix #606, add more rest service for new UI(version 2)

### LINKED ISSUE #606: New UI
We need to define the requirements and apply them to implementable wireframes for UI. 

##### Structure
Here is the sitemap:
- Dashboard 
- Cluster
  - Master
  - Workers
  - Configuration
- Applications
  - Overview
  - Application Detail
    - Summary
    - Processors
    - Tasks
    - Configuration
- Tool
- Administration

##### Dashboard 
A place to look at the overall cluster and application status

##### Cluster
The section is used to view and manage the cluster. It consists of three parts as follows.

##### Master
Master is the brain of the cluster. It can be one master or a master group. We need to:
- View current runtime information
- View current resource usage information
- View log and config files
- View recent warning and failure alerts (parsed from log) as a list
- View recent resource usage in charts

###### Wireframe (in 6-6 column layout)
master

1. On top left, the summary box lists master’s runtime information. *Equivalent to the existing master page*
2. Below, the alerts box lists recent warnings and errors. Alerts are sorted by timestamp.
3. On top right, there are two circles. The first circle indicates the number of worker instances. The second circle indicates the usage of slots. 
4. Below, there will be couple of line charts (e.g. number of errors) to indicate the cluster running status in a particular past period.

##### Workers
Worker instances are a set of computation resources. We need to:
- View current runtime information
- View current resource usage information (optional)
- View aggregated resource usage information
- View individual worker instance details

###### Wireframe
workers
1. On top there is a fuzzy filter to identify worker instance (optional)
2. Below there is a table of worker instances. (Pagination control is optional) 

##### Worker "X" Detail
Worker is running on a specified physical machine. Worker can created a set of executors to run particular applications. When applications are stopped, executors will be destroyed. And the worker will become idle. We need to:
- View the current runtime information of worker
- View the current executors' activities
- View recent warning and failure alerts (parsed from log) as a list
- View recent resource usage in charts

###### Wireframe (in 6-6 column layout)
workerx

1. On top left, the Summary box lists the information of the worker.
2. Below, there is a list of executors with corresponding actorpath, etc.
3. Below, the alerts box lists recent warnings and errors. Alerts are sorted by timestamp.
4. On the right side there will be couple of charts, e.g. CPU time, Memory usage, Disk IO, Network IO.

##### Configuration
We need to:
- View cluster's configuration
- Download config file
- Modify and provisioning configuration changes (Not in scope)
- View configuration change history (Not in scope)

###### Wireframe
apps_config

##### Applications
The section is used to view and manage the applications. It consists of three parts as follows.

##### Submit New Application
Not implemented yet. It will be a wizard to submit new application.

##### Application Overview
We need to:
- View all running or completed applications as a list

###### Wireframe
apps

1. On top there is a fuzzy filter
2. Right to filter there is a button to submit a new application (Not in scope)
3. Below there is a table of applications.

##### App "X" Detail
Detail will be broken down into four parts: Summary, Processors, Tasks, Configuration.

##### App "X" Summary
This is the overview of a running application. We need to:
- View current runtime information
- View important message processing metrics
- View recent resource usage in charts

###### Wireframe (in 6-6 column layout)
appx

1. On top left there is a summary box.
2. Below there is a list of important message processing metrics (1min, 10min, 1hour, etc.)
3. On top right, there is a button back to Applications overview page
4. On the right side there are couple of line charts (throughput, processing time, latency at application level)

##### App "X" Processors (Type: streaming application)
Application can be described as a DAG. Nodes are processors. Edge between two nodes describes how data flow goes and its throughput, etc. We need to:
- View the topology of the processors (DAG control). 
- View the current throughput, working load, etc. on the DAG live.
- The DAG control should support add/modify/remove nodes/edges live with context menu.
- View the metrics data of every single processors

###### Wireframe (in 6-6 column layout)
apps_processes

1. On top left there is a DAG control. Above the DAG control, the number of processed messages should be shown.
2. Below there is a selector for different metrics with a table of metrics data (the table can be moved to right, if we want to bind the table with dag when hovering over a row or a dag node)
3. On top right, there is a button back to Applications overview page
4. On the right side there are charts of different metrics 

##### App "X" Tasks (When user click one processor of DAG)
Tasks describe the internal relationship of a processor. We need to:
- View the details of tasks.
- View the metrics of tasks.
- Allow user to add/remove tasks to adjust processors work performance. (Not in scope)

###### Wireframe (in 6-6 column layout)
apps_tasks

1. On top left there is a pie chart of tasks’ throughput.
2. Below there is table of tasks information
3. On top right, there is a button back to Applications overview page
4. On the right side there are charts of metrics at task level

##### App "X" Configuration
Configuration of the application. Similar to Cluster's configuration.

clockfly:


clockfly:
Changes:

1. add rest service for executor summary:

```
+  case class ExecutorSummary(
+    id: Int,
+    workerId: Int,
+    actorPath: String,
+    logFile: String,
+    status: String,
+    taskCount: Int,
+    tasks: Map[ProcessorId, List[TaskId]]
+  )
```

REST: http://127.0.0.1:8090/api/v1.0/appmaster/1/executor/0

2. add config rest api for executor process

REST: http://127.0.0.1:8090/api/v1.0/appmaster/1/executor/0/config

3. log activities for admin operation 
https://github.com/gearpump/gearpump/pull/1309/files#diff-7ee9bd0bcbf26dee4d4f5347bf9dda30R135

4. add more info in processor summary. Like executors used by processor.

5. add more info in StreamAppMasterSummary.

core/src/main/scala/org/apache/gearpump/cluster/master/MasterSummary.scala
```diff
@@ -23,7 +23,14 @@ object MasterStatus {
...
+  leader: (String, Int),
+  cluster: List[(String, Int)],
+  aliveFor: Long,
+  logFile: String,
+  jarStore: String,
+  masterStatus: MasterStatus.Type,
+  homeDirectory: String,
+  activities: List[MasterActivity])
+
+case class MasterActivity(time: Long, event: String)
```
kkasravi: Do we want the *Summary.scala files copied into services/shared/... so that services/{jvm,js} can access them from the shared directory?

clockfly: I don't know what is the best approach. 

1. we don't want to create a package named shared, that means we need to copy everything under this package.
2. We don't want to modify the copy-logic for every new class file we created.

kkasravi: We could copy into just services/js and keep the package name intact. We could probably automate what classes are copied by looking for @JSExport which is a dummy tag for java and does nothing.

kkasravi:
Compilation error? ...
[INFO] [08/12/2015 15:19:48.775] [Metrics$] Metrics is enabled...,  false
[INFO] [08/12/2015 15:19:48.935] [AppMasterSpec] Actor system is started, 127.0.0.1, 46570
[error] /home/travis/build/gearpump/gearpump/services/jvm/src/test/scala/org/apache/gearpump/services/WorkerServiceSpec.scala:43: not found: value WorkerSummary
[error]           sender ! WorkerData(WorkerSummary.empty)
[error]                               ^

streaming/src/main/scala/org/apache/gearpump/streaming/appmaster/StreamAppMasterSummary.scala
```diff
@@ -19,26 +19,42 @@
...
-    executors: Map[ExecutorId, String] = null,
-    tasks: Map[TaskId, ExecutorId] = null,
+    executors: List[ExecutorBrief] = null,
     status: AppMasterStatus = MasterToAppMaster.AppMasterActive,
     startTime: TimeStamp = 0L,
-    user: String = null)
-  extends AppMasterDataDetail
\ No newline at end of file
+    user: String = null,
+    appType: String = "streaming",
```
stanleyxu2005: `appType = "streaming"` appears twice. I thought it is a readonly property (should not have a setter) and it should be unique (should set only once internally)

clockfly:
Thanks.

All review comments solved.

stanleyxu2005:
+1

PR #1323, opened 2015-08-19
fix #924, fix storm wordcount example

### LINKED ISSUE #924: Storm WordCount cannot run correctly on gearpump
The value of "SINK PROCESSORS RECEIVE THROUGHPUT" is always 0 msgs/s.

manuzhang:


experiments/storm/src/main/scala/org/apache/gearpump/experiments/storm/processor/StormBoltOutputCollector.scala
```diff
@@ -18,20 +18,23 @@
...
+import org.apache.gearpump.util.LogUtil
+import org.slf4j.Logger
 
 private[storm] class StormBoltOutputCollector(outputFn: (String, JList[AnyRef]) => Unit) extends IOutputCollector {
+  private val LOG: Logger = LogUtil.getLogger(getClass)
 
   override def emit(streamId: String, anchors: JCollection[Tuple], tuple: JList[AnyRef]): JList[Integer] = {
     outputFn(streamId, tuple)
-    null
+    new JArrayList[Integer](0)
```
clockfly: Should we reuse the empty array?

clockfly:
+1

PR #1325, opened 2015-08-19
fix #1246 isolate processors' classpath

### LINKED ISSUE #1246: realtime machine learning scoring engine
User story:

1. Allow user to submit a machine learning topology by REST.
2. The gearpump environment classpath should be clean and not affect user app.
3. For different processor, it may use different algorithm, and the algorithm is hosted in different jars. So the topology should be able to support multiple jars.
4. User can dynamically swap one algorithm implementation by providing a new jar, without impacting other algorithm.
5. Need to support a RPC like question and answer. For each data point, the topology should be able to return a score.

huafengw:
The backend is ready, need to coordinate with the frontend


clockfly:
Can you also finish the UI part?

huafengw:
Not sure, I'll try

clockfly:
Leave the UI fix to another PR.

+1

PR #1328, opened 2015-08-19
fix #1327, move project state and dsl to under streaming

### LINKED ISSUE #1327: move project state and dsl to streaming

clockfly:


manuzhang:
+1

PR #1333, opened 2015-08-20
fix #1330, Worker cannot init when setting HDFS as jarStore

### LINKED ISSUE #1330: Worker cannot init when setting HDFS as jarStore
My hadoop version is 2.7.1, gaerpump is deployed on yarn, and when I change jarStore to HDFS the worker cannot init.

clockfly:


manuzhang:
+1

PR #1342, opened 2015-08-20
fix #1341, Cannot submit application in UI if one master is killed an…

### LINKED ISSUE #1341: Cannot submit application in UI if one master is killed and another master started
When killing a master in cluster and another master started, user cannot submit application in UI.

clockfly:
…d another master started


clockfly:
test pass on 3 nodes.

huafengw:
+1

PR #1348, opened 2015-08-21
fix #1344 replicate application history

### LINKED ISSUE #1344: When a master is killed, the running applications disappear in UI
When a master in cluster is killed, the applications are still running but disappear in UI

huafengw:


clockfly:
+1

PR #1360, opened 2015-08-23
fix #1358: properties will be updated 

### LINKED ISSUE #1358: Property table will not update sometimes
Property values in new UI might not update

stanleyxu2005:
(2) moved roboto font and echarts to dashing-deps
(3) fixed timestamp /1000 problem
(4) split life time into birth time and death time


clockfly:
+1

PR #1362, opened 2015-08-23
fix #1361: able to show multiple tasks in chart 

### LINKED ISSUE #1361: In processor view, chart cannot show multiple task data series
Seems to be a new introduced issue...

stanleyxu2005:
also fixed JVM metric chart y-axis label issue


manuzhang:
+1

PR #1365, opened 2015-08-24
fix #1364, add default value for BYTES_PER_MESSAGE of SOLStreamProducer

### LINKED ISSUE #1364: fail to replace source with SOLStreamProducer

manuzhang:


huafengw:
+1 
So we should also allow user provide a related config when replacing a processor

manuzhang:
yes, could you open an issue for that ?

PR #1369, opened 2015-08-24
fix #1359, split metric source with metric item

### LINKED ISSUE #1359: Split metrics owner name and metrics name
Currently its inconvenience for dashboard to handle metrics name. The metric owner name and real metric name are mixed together. For instance, `worker0.executor1.physical.memory.total`, Dashboard will have to split the string to understand it is a  `physical.memory.total`. 

Better split them where it is manipulated, so that dashboard do not need to parse the name everything.

clockfly:


stanleyxu2005:
Please use https://github.com/gearpump/gearpump/pull/1370

PR #1388, opened 2015-08-31
fix #1238, adds BroadcastPartitioner

### LINKED ISSUE #1238: Support broadcast partitioner

manuzhang:


whjiang:
+1.

streaming/src/main/scala/io/gearpump/streaming/task/Subscription.scala
```diff
@@ -74,17 +75,29 @@ class Subscription(
     transport.transport(ackRequest, allTasks: _*)
   }
 
+  def sendMessage(msg: Message): Int = {
+    partitioner match {
+      case up: UnicastPartitioner =>
```
clockfly: Can we avoid do the pattern match every time?

Will there be possible performance degradtion?

manuzhang:
@clockfly no pattern matching for every new message. Further comments ?

clockfly:
+1

Issue #491, opened 2015-02-15
Frontend requires a javascript graph API


kkasravi:
In addition to the JSON model that is visualized by the visdag widget, we should also have a javascript graph api so adding, deleting nodes, defining new edges, etc.

kkasravi:
One candidate is cytoscape which does visualization and provides a graph api. We already have visualization using vis.js, however we should look at the Graph API. See http://js.cytoscape.org/

clockfly:
@kkasravi 

There is a nice demo in visjs. 
http://visjs.org/examples/network/21_data_manipulation.html

Maybe we can just follow its approach.

kkasravi:
Yes, this is nice. We will need the related REST/WS API to send the DAG.

clockfly:
Yes, I agree.

This issue should be solved together with https://github.com/intel-hadoop/gearpump/issues/53

clockfly:
In this feature, we can allow user to compose a DAG completely in UI.

huafengw:
A json example of SubmitApplicationRequest:
```
{
  "appName": "complexdag",
  "processors": [
    [
      0,
      {
        "id": 0,
        "taskClass": "A",
        "parallelism": 1,
        "description": "processor A",
        "jar": {
          "name": "file1",
          "filePath": {
            "path": "path1"
          }
        }
      }
    ],
    [
      1,
      {
        "id": 1,
        "taskClass": "B",
        "parallelism": 1,
        "description": "processor B",
        "jar": {
          "name": "file2",
          "filePath": {
            "path": "path3"
          }
        }
      }
    ],
    [
      2,
      {
        "id": 2,
        "taskClass": "C",
        "parallelism": 1,
        "description": "processor C",
        "jar": {
          "name": "file3",
          "filePath": {
            "path": "path3"
          }
        }
      }
    ]
  ],
  "dag": {
    "vertexList": [
      0,
      1,
      2
    ],
    "edgeList": [
      [
        1,
        "io.gearpump.partitioner.ShufflePartitioner",
        2
      ],
      [
        0,
        "io.gearpump.partitioner.HashPartitioner",
        1
      ]
    ]
  },
  "userconfig": {
    "_config": {
      "key1": "value1"
    }
  }
}
```

stanleyxu2005:
The PR #1452 is ready. Here are some screenshots.

[...]

PR #1391, opened 2015-09-01
part of #491 allow user to compose a DAG in UI, backend part

huafengw:


services/jvm/src/main/scala/io/gearpump/services/MasterService.scala
```diff
@@ -157,14 +152,46 @@ trait MasterService extends HttpService {
...
+            }
+          }
+        }
+      } ~
+      path("uploadedtasks") {
+        onComplete(askActor[UploadedTasks](master, GetUploadedTasks)) {
+          case Success(tasks: UploadedTasks) =>
+            complete(write(tasks))
+          case Failure(ex) =>
+            failWith(ex)
```
kkasravi: Does this result in appropriate HTTP error code and message?

daemon/src/main/scala/io/gearpump/cluster/master/AppManager.scala
```diff
@@ -214,6 +213,12 @@ private[cluster] class AppManager(kvService: ActorRef, launcher: AppMasterLaunch
         case AppMasterNonExist =>
           sender ! AppMasterData(AppMasterNonExist)
       }
+
+    case uploadTasks: UploadTaskJar =>
+      uploadTasks.tasks.foreach(uploadedTasks += _ -> uploadTasks.jar)
+      sender ! UploadedTasks(uploadedTasks)
+    case GetUploadedTasks =>
+      sender ! UploadedTasks(uploadedTasks)
```
kkasravi: @huafengw @clockfly 
For really big jars (60M+) I found that it killed the master with OOM. I concluded that we needed to stream data in the multipart form which although much more involved would not endanger the master stability. Perhaps check with @clockfly - I think Sean was working on converting spray to http (akka-streams). This PR could be accepted with a note to convert to streaming IO.

services/jvm/src/main/scala/io/gearpump/services/MasterService.scala
stanleyxu2005: @kkasravi It will return an HTTP 500 back in general.

core/src/main/scala/io/gearpump/cluster/ClusterMessage.scala
```diff
@@ -78,6 +82,8 @@ object MasterToClient {
   case class HistoryMetrics(path: String, metrics: List[HistoryMetricsItem])
 
   case class LastFailure(time: TimeStamp, error: String)
+
+  case class UploadedTasks(tasks: Map[String, AppJar])
```
clockfly: What UploadedTasks mean?

huafengw: Request by frontend, return all the uploaded task classes and corresponding jar

services/jvm/src/main/scala/io/gearpump/services/RestServices.scala
```diff
@@ -43,10 +46,24 @@ trait RestServices extends
...
+  def getJarStoreService: JarStoreService
+}
+
 class RestServicesActor(masters: ActorRef, sys:ActorSystem) extends Actor with RestServices {
   def actorRefFactory = context
   implicit val system: ActorSystem = sys
   implicit val eh = RoutingSettings.default(context)
+  lazy val jarStoreService = {
+    val service = JarStoreService.get(system.settings.config)
+    service match {
```
clockfly: can the case match be part of JarStoreService.get?

services/jvm/src/main/scala/io/gearpump/services/MasterService.scala
```diff
@@ -157,14 +152,46 @@ trait MasterService extends HttpService {
...
 
-            import upickle.default.{read, write}
+            import upickle.default.write
             val submitApplicationResultValue = SubmitApplicationResultValue(appId)
             val jsonData = write(submitApplicationResultValue)
             complete(jsonData)
           }
         }
+      } ~
+      path("uploadtasks") {
```
clockfly: Should this be more general uploadJar?

As the name uploadTasks doesn't make any sense, why a task can be uploaded?

I suggest use general UploadJar to do only one thing, which is to upload a jar file, this API will return the JarPath for this jar.

The frontend will be responsible to bind task class name to jarPath, and submit the DAG.

daemon/src/main/scala/io/gearpump/cluster/master/AppManager.scala
```diff
@@ -214,6 +213,12 @@ private[cluster] class AppManager(kvService: ActorRef, launcher: AppMasterLaunch
         case AppMasterNonExist =>
           sender ! AppMasterData(AppMasterNonExist)
       }
+
+    case uploadTasks: UploadTaskJar =>
+      uploadTasks.tasks.foreach(uploadedTasks += _ -> uploadTasks.jar)
```
clockfly: It doesn't make any sense to create another mapping uploadTasks....

As ProcessorDescription already contains all the information.
in ProcessorDescription, there is taskClass name, and AppJar.

huafengw: I know your point, Stanley and I discussed about it before. We choose to save these info on master side because we hope user don't need to reupload the jar file if the composing failed accidentally.

services/jvm/src/main/scala/io/gearpump/services/MasterService.scala
stanleyxu2005: +1

services/jvm/src/main/scala/io/gearpump/services/AppMasterService.scala
```diff
@@ -72,7 +59,7 @@ trait AppMasterService extends HttpService {
                 if(jar.nonEmpty) {
                   dagOperation match {
                     case replace: ReplaceProcessor =>
-                      val description = replace.newProcessorDescription.copy(jar = uploadJar(jar.get, jarStoreService))
+                      val description = replace.newProcessorDescription.copy(jar = Util.uploadJar(jar.get, getJarStoreService))
```
clockfly: This will block the IO

core/src/main/scala/io/gearpump/jarstore/JarStoreService.scala
```diff
@@ -59,8 +61,17 @@ object JarStoreService {
   }
 
   def get(config: Config): JarStoreService = {
```
clockfly: Is get(config: Config) still used?
Should we only expose one public method?

clockfly:
How about this

In JarStoreService, define a function called init
```
def init(conf: Config, actorRefFactory: ActorRefFactory): Unit
```

clockfly:
Then we don't need getAndInit in JarStoreService

huafengw:
Updated

clockfly:
+1

PR #1393, opened 2015-09-01
fix #1392, add TimeExactor to TypedDataSource

### LINKED ISSUE #1392: can't extract app time from source in DSL API

manuzhang:


streaming/src/main/scala/io/gearpump/streaming/dsl/StreamApp.scala
```diff
@@ -78,15 +78,35 @@ object StreamApp {
...
   override def read(batchSize: Int): List[Message] = {
-    val size = Math.min(seq.length - index % seq.length, batchSize)
-    index += size
-    seq.slice(index % seq.length - size, size).toList.map(msg => Message(msg.asInstanceOf[AnyRef], System.currentTimeMillis()))
+    if (seq.isEmpty) {
+      List.empty[Message]
+    } else {
+      val size = Math.min(seq.length - index % seq.length, batchSize)
+      index += size
+      seq.slice(index % seq.length - size, size).toList.map(msg => Message(msg.asInstanceOf[AnyRef], timeExtractor(msg)))
```
stanleyxu2005: Can the calculation of `startIndex` and `stopIndex` of `slice` be more straightforward?

huafengw:
What kind of scenario will using TimeExtractor?

manuzhang:
When user wants to inject app time from external source.

For example, count the number of tweets of last second for every second. 
```scala
   val twitter = TwitterStream
   source.setTimeExtractor(TimeExtractor[Status](_.getTimeStamp)
   app.source(twitter, 1, "twitter stream").window(1000, 1000).sum
```

PR #1394, opened 2015-09-02
fix #1392, add TimeExactor to TypedDataSource

### LINKED ISSUE #1392: can't extract app time from source in DSL API

manuzhang:
previously in #1393

### DIFF
--- a/streaming/src/main/scala/io/gearpump/streaming/dsl/StreamApp.scala
+++ b/streaming/src/main/scala/io/gearpump/streaming/dsl/StreamApp.scala
@@ -18,14 +18,14 @@
 
 package io.gearpump.streaming.dsl
 
-import akka.actor.{ActorRef, ActorSystem, Cancellable, Props}
+import akka.actor.ActorSystem
+import io.gearpump.cluster.UserConfig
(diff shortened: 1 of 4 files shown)


streaming/src/main/scala/io/gearpump/streaming/dsl/StreamApp.scala
```diff
@@ -78,15 +78,37 @@ object StreamApp {
...
+  def apply[T](fn: T => TimeStamp): TimeExtractor[T] = new TimeExtractor[T] {
+    override def apply(t: T): TimeStamp = fn(t)
+  }
+}
+trait TimeExtractor[T] {
+  def apply(t: T): TimeStamp
+}
+
+trait TypedDataSource[T] extends DataSource {
+  def setTimeExtractor(extractor: TimeExtractor[T]): Unit
```
clockfly: It doesn't feel right to put setTimeExtractor as part of interface.

clockfly:
I think setTimeExtractor should not be part of DataSource interface.

It will be strange if we call setTimeExtrator, but it is not effective as user choose not use it when implementing function.

```
def read(batchSize: Int): List[Message]
```

PR #1398, opened 2015-09-07
fix #1397, allow KafkaSink to receive Message of bytes

### LINKED ISSUE #1397: would it better to support on external KafkaSink to support single Array[Byte] rather than (Array[Byte], Array[Byte])
I have to make my sink source message format to tuple2(Array[Byte], Array[Byte]) 
if not match that format, it would throw class cast exception in backend.
but it would be better just support Array[Byte] and then serialize it to an Object?

manuzhang:


clockfly:
+1

external/kafka/src/test/scala/io/gearpump/streaming/kafka/KafkaSinkSpec.scala
```diff
@@ -39,12 +39,17 @@ class KafkaSinkSpec extends PropSpec with PropertyChecks with Matchers with Mock
...
+        val (topic, key, msg) = data
+        val kafkaSink = new KafkaSink(() => producer, topic)
+        kafkaSink.write(Message((key, msg)))
+        verify(producer).send(MockUtil.argMatch[ProducerRecord[Array[Byte], Array[Byte]]](
+          r => r.topic == topic && (r.key sameElements key) && (r.value sameElements msg)))
+        kafkaSink.write(Message(msg))
+        verify(producer).send(MockUtil.argMatch[ProducerRecord[Array[Byte], Array[Byte]]](
+          r => r.topic() == topic && (r.key == null) && (r.value() sameElements msg)
+        ))
+        kafkaSink.close()
```
stanleyxu2005: Will Kafka sink always be closed even an exception is thrown here?

PR #1400, opened 2015-09-09
fix #1399: in application overview page the task count of executor wa…

### LINKED ISSUE #1399: incorrect Executor.TaskCount on Application overview tab
To reproduce, please launch a local SOL application with args `-streamProducer 3 -streamProcessor 3`. Usually, executor 1 and executor 2 will only have 1 or 2 tasks instead of 3.

stanleyxu2005:
…s wrong

### DIFF
--- a/services/dashboard/services/models/models.js
+++ b/services/dashboard/services/models/models.js
@@ -156,16 +156,9 @@ angular.module('dashboard')
                 {source: parseInt(item[0]), target: parseInt(item[2]), type: item[1]}];
             });
           }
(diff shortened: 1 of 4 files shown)


manuzhang:
now the `Executor.TaskCount` on each processor is incorrectly computed to total count of tasks in an executor.

stanleyxu2005:
Indeed. Backend sends an array of executors not a map of executors, so that the index does not mean the executor id. In the patch, I convert the executor array to an executor map. Problem solved in this way.

manuzhang:
+1

PR #1406, opened 2015-09-10
fix #1405: now the number of executors is back

### LINKED ISSUE #1405: The number of executors disappear from application header
After the PR #1400, the `app.executors` has been changed from array to object. However there is no object.length to get the number of properties of an object. This is the reason why the number disappears.

stanleyxu2005:


manuzhang:
+1

PR #1413, opened 2015-09-11
fix #1412: add 'no-cache' header in ajax responses

### LINKED ISSUE #1412: Need 'no-cache' header in RESTful api response
IE will cache response, so that the dashboard contents will NEVER get updated. I found a workaround is to add `no-cache` header into every ajax responses.
```
    $httpProvider.defaults.headers.get['If-Modified-Since'] = new Date(0);
    $httpProvider.defaults.headers.get['Cache-Control'] = 'no-cache';
    $httpProvider.defaults.headers.get['Pragma'] = 'no-cache';
```

stanleyxu2005:
Without the patch MSIE 8/9/10/11 will not receive page changes, because the response of ajax calls are cached...

### DIFF
--- a/services/dashboard/dashboard.js
+++ b/services/dashboard/dashboard.js
@@ -44,6 +44,15 @@ angular.module('dashboard', [
     cfpLoadingBarProvider.latencyThreshold = 200;
   }])
 
+  // disable caching for ajax calls to make MSIE happy
+  .config(['$httpProvider', function($httpProvider) {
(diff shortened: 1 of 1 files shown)


clockfly:
+1

stanleyxu2005:
After further digging, as the patch will modify response header. It will break CORS (cross-origin resource sharing), which at least breaks my dashboard debugging environment.

PR #1415, opened 2015-09-14
fix #1373 add duplicated edge detecting

### LINKED ISSUE #1373: MessageLoss for a DAG where two nodes are connected with two edges
Reproducible with a valid DAG like `Graph(a ~ p1 ~> b, a ~ p2 ~> b)`. The issue is that there is only **one** b processor but **two** b subscribers of a

huafengw:


clockfly:
+1

PR #1416, opened 2015-09-14
part of #1373 add duplicated edge detecting

huafengw:
This piece of code is accidentally removed when merging the last pr.

### DIFF
--- a/services/jvm/src/main/scala/io/gearpump/services/MasterService.scala
+++ b/services/jvm/src/main/scala/io/gearpump/services/MasterService.scala
@@ -154,6 +154,10 @@ trait MasterService {
                   PartitionerDescription(new PartitionerByClassName(edge))
                 }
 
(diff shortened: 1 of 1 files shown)


services/jvm/src/main/scala/io/gearpump/services/MasterService.scala
```diff
@@ -154,6 +154,10 @@ trait MasterService {
                   PartitionerDescription(new PartitionerByClassName(edge))
                 }
 
+                if(graph.hasDuplicatedEdge()){
```
clockfly: This is not right.

1. The frontend will not check this value.
2. new StreamApplication() will throw exception, the route will already return a 500 error.

huafengw: OK, I'll close this pr

PR #1420, opened 2015-09-14
fix #1228, add java interface for Graph API

### LINKED ISSUE #1228: Stablize the java interface and scala interface

clockfly:


streaming/src/main/java/io/gearpump/streaming/javaapi/Processor.java
```diff
@@ -0,0 +1,79 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.gearpump.streaming.javaapi;
+
+import io.gearpump.cluster.UserConfig;
+
+public class Processor<T extends Task> implements io.gearpump.streaming.Processor<T> {
+  private Class<T> _taskClass;
```
stanleyxu2005: LGTM. +1

The only one thing could be the leading underscore of all private properties. Is it a common java name convention, isn't it?

stanleyxu2005:
+1

PR #1421, opened 2015-09-14
fix #1228, add java interface for Graph API

### LINKED ISSUE #1228: Stablize the java interface and scala interface

clockfly:


manuzhang:
Could we have a simple example using java api ?

clockfly:
The example will be hosted at https://github.com/gearpump/gearpump-java-example

manuzhang:
I think a simple in-gearpump java example that evolves with gearpump api and is always up to date would be valuable. Besides, it would be a good guide for whoever wants to contribute to https://github.com/gearpump/gearpump-java-example

clockfly:
Good suggestion.

Patch updated.

examples/streaming/wordcount-java/src/main/java/io/gearpump/streaming/examples/wordcountjava/WordCount.java
```diff
@@ -0,0 +1,63 @@
...
+    // For split task, we config to create two tasks
+    int splitTaskNumber = 2;
+    Processor split = new Processor(Split.class).withParallelism(splitTaskNumber);
+
+    // For sum task, we have two summer.
+    int sumTaskNumber = 2;
+    Processor sum = new Processor(Sum.class).withParallelism(sumTaskNumber);
+
+    UserConfig conf = UserConfig.empty();
+    StreamApplication app = new StreamApplication("wordcountJava", conf, system);
```
huafengw: Why user have to construct an ActorSystem to build a StreamApplication?

clockfly: Good point

clockfly:
@huafengw 

Please check the new API design.

examples/streaming/wordcount-java/src/main/java/io/gearpump/streaming/examples/wordcountjava/Split.java
```diff
@@ -0,0 +1,59 @@
...
+    this.userConf = userConf;
+  }
+
+  private Long now() {
+    return System.currentTimeMillis();
+  }
+
+  @Override
+  public void onStart(StartTime startTime) {
+    self().tell(new Message("start", now()), self());
```
manuzhang: maybe wrap this in a method of super class like

```java
   public void outputToSelf(Message msg) {
      self().tell(msg, self());
   }
```

clockfly: I don't think it is cleaner. I think we should align with how akka does this.

examples/streaming/wordcount-java/src/main/java/io/gearpump/streaming/examples/wordcountjava/WordCount.java
```diff
@@ -0,0 +1,67 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.gearpump.streaming.examples.wordcountjava;
+
+import akka.actor.ActorSystem;
```
manuzhang: please clean up the imports

huafengw:
Kind of inconsistent that we have ```ClientContext.submit(app)``` in scala while ```app.submit(clientContext)``` in java. But generally I'll +1 with this PR.

clockfly:
@huafengw

Please check the new look and feel. Now it is consistent with scala.

huafengw:
+1

PR #1425, opened 2015-09-15
fix #1422: removed if-modified-since header. it is not relevant and i…

### LINKED ISSUE #1422: UI issue about if-modified-since
```
WARN ActorSystemImpl: Illegal request header: Illegal 'if-modified-since' header: Invalid input '7', expected ':' (line 1, column 14): Thu Jan 01 1970 08:00:00 GMT+0800 (China Standard Time)
             ^

```

stanleyxu2005:
…s handled very different by different web servers


clockfly:
+1

PR #1426, opened 2015-09-15
fix #1228, further simplify the java interface

### LINKED ISSUE #1228: Stablize the java interface and scala interface

clockfly:


huafengw:
+1

PR #1427, opened 2015-09-15
fix #1228, add data source and data sink for java api

### LINKED ISSUE #1228: Stablize the java interface and scala interface

clockfly:


clockfly:
test failure is irrelevant.

streaming/src/main/java/io/gearpump/streaming/javaapi/Processor.java
```diff
@@ -35,6 +40,29 @@ public Processor(Class<T> taskClass, int parallelism) {
     this._parallelism = parallelism;
   }
 
+  public Processor(DataSink dataSink, int parallelism, String description,  UserConfig taskConf, ActorSystem system) {
```
manuzhang: This requires user to create an `ActorSystem` ?

streaming/src/main/java/io/gearpump/streaming/javaapi/Processor.java
```diff
@@ -35,6 +40,29 @@ public Processor(Class<T> taskClass, int parallelism) {
...
+    io.gearpump.streaming.Processor p = DataSinkProcessor.apply(dataSink, parallelism, description, taskConf, system);
+    this._taskClass = p.taskClass();
+    this._parallelism = p.parallelism();
+    this._description = p.description();
+    this._userConf = p.taskConf();
+  }
+
+  public Processor(DataSource source, int parallelism, String description,  UserConfig taskConf, ActorSystem system) {
+    io.gearpump.streaming.Processor p = DataSourceProcessor.apply(source, parallelism, description, taskConf, system);
```
manuzhang: why not reuse the copy constructor ?

streaming/src/main/java/io/gearpump/streaming/javaapi/Processor.java
```diff
@@ -35,6 +40,29 @@ public Processor(Class<T> taskClass, int parallelism) {
...
 
+  public Processor(DataSink dataSink, int parallelism, String description,  UserConfig taskConf, ActorSystem system) {
+    io.gearpump.streaming.Processor p = DataSinkProcessor.apply(dataSink, parallelism, description, taskConf, system);
+    this._taskClass = p.taskClass();
+    this._parallelism = p.parallelism();
+    this._description = p.description();
+    this._userConf = p.taskConf();
+  }
+
+  public Processor(DataSource source, int parallelism, String description,  UserConfig taskConf, ActorSystem system) {
```
manuzhang: would a factory method `source` be more informative ?

streaming/src/main/java/io/gearpump/streaming/javaapi/Processor.java
clockfly: Do you mean call other constructor by this()?
It looks like JDK6 don't support that.

streaming/src/main/java/io/gearpump/streaming/javaapi/Processor.java
clockfly: Good point.

streaming/src/main/java/io/gearpump/streaming/javaapi/Processor.java
clockfly: This maps to DataSinkProcessor constructor api.

clockfly:
updated.

clockfly:
Example of the java API can be found at:
https://github.com/gearpump/gearpump-java-example/commit/532083e49fb94b7486e3c983a3ff3be5d044926e

streaming/src/main/scala/io/gearpump/streaming/StreamApplication.scala
```diff
@@ -29,7 +29,7 @@ import io.gearpump.util.{Graph, ReferenceEqual}
 import scala.language.implicitConversions
 import scala.reflect.ClassTag
 
-trait Processor[+T <: Task] extends ReferenceEqual {
+trait  Processor[+T <: Task] extends ReferenceEqual {
```
stanleyxu2005: Avoid unnecessary code change (an extra space found)

streaming/src/main/java/io/gearpump/streaming/javaapi/StreamApplication.java
```diff
@@ -34,8 +34,10 @@
    * @param conf
    */
   public StreamApplication(String name, UserConfig conf, Graph graph) {
+    //by pass the tricky type check in scala 2.10
+    io.gearpump.util.Graph graphHelp = graph;
     this.app = io.gearpump.streaming.StreamApplication.apply(
-        name, graph, conf);
+        name, graphHelp, conf);
```
stanleyxu2005: Can it be `(io.gearpump.util.Graph) graph`? I think a `graphHelp` is wired

streaming/src/main/java/io/gearpump/streaming/javaapi/Processor.java
```diff
@@ -36,6 +43,41 @@ public Processor(Class<T> taskClass, int parallelism) {
...
   /**
+   * Create a Source Processor
+   * @param dataSink
+   * @param parallelism
+   * @param description
+   * @param taskConf
+   * @param system
+   * @return
+   */
+  public static Processor<DataSinkTask> sink(DataSink dataSink, int parallelism, String description,  UserConfig taskConf, ActorSystem system) {
```
manuzhang: maybe passing in `ClientContext` and hiding `ActorSystem` from users is better ?

clockfly: I don't think so.

clockfly:
@stanleyxu2005  updated.

streaming/src/main/java/io/gearpump/streaming/javaapi/Processor.java
huafengw: Or can we avoid the ActorSystem here? Seems it is only used for serialization.

clockfly: It depends on the scala interface DataSinkProcessor which requires an ActorSystem. If we want to do this, then we need to change DataSinkProcessor first, then maybe changing UserConfig serialization mechanism eventually. It will be a big change. 

I can live with current method, as long as it is compatible with the scala interface(implicit in scala should not be considered as a escape).

huafengw:
+1

stanleyxu2005:
+1

PR #1440, opened 2015-09-18
fix #1387 add a serializer delegate when serialize Message

### LINKED ISSUE #1387: user provided kryo serializer cannot be used
Since the `Serializer` interface is shaded, any user provided serializer that extends the original interface would throw a ClassCastException unless that seriailzer is shaded as well.

huafengw:


core/src/main/scala/io/gearpump/serializer/FastKryoSerializer.scala
```diff
@@ -74,7 +76,18 @@ class FastKryoSerializer(system: ExtendedActorSystem) {
...
   def deserialize(msg : Array[Byte]): AnyRef = {
-      kryoSerializer.fromBinary(msg)
+    kryoSerializer.fromBinary(msg)
+  }
+
+  private def getDelegator(): Option[SerializationDelegate] = {
+    Try(config.getString(Constants.GEARPUMP_SERIALIZER_DELEGETE_FACTORY)).map { factoryClass =>
+      if (factoryClass != "") {
+        val factory = Class.forName(factoryClass).newInstance()
+        Some(factory.asInstanceOf[IDelegateFactory].getDelegate(system))
```
manuzhang: what if factory is not instance of IDelegateFactory ? Any fallback strategy ?

huafengw: You mean user provide a non-IDelegateFactory class?

manuzhang: yes

huafengw: Suppose user provide a non-IDelegateFactory class, the getDelegtor function will throw exception. If we fall back to non-delegator, user's message still can not be serialized correctly, so maybe throw the exception earlier  is better.

core/src/main/scala/io/gearpump/util/Constants.scala
```diff
@@ -36,6 +36,7 @@ object Constants {
   val GEARPUMP_SCHEDULING_SCHEDULER = "gearpump.scheduling.scheduler-class"
   val GEARPUMP_SCHEDULING_REQUEST = "gearpump.scheduling.requests"
   val GEARPUMP_SERIALIZERS = "gearpump.serializers"
+  val GEARPUMP_SERIALIZER_DELEGETE_FACTORY = "gearpump.serializer.delegate.factory"
```
manuzhang: typo "DELEGATE"

core/src/main/scala/io/gearpump/serializer/IMessageSerializer.scala
```diff
@@ -0,0 +1,29 @@
...
+package io.gearpump.serializer
+
+import io.gearpump.esotericsoftware.kryo.Serializer
+
+/**
+ * This trait should only be used in Gearpump and for message serialization.
+ * When a Class extends this trait. Gearpump will construct the corresponding object
+ * using reflection and pass in a delegate.
+ */
+abstract class IMessageSerializer[T] extends Serializer[T]{
```
manuzhang: then why not use a trait if you mean to ? You could restrict its visibility with `private [gearpump]`

Issue #1377, opened 2015-08-26
allow user to output to a specific subscriber


manuzhang:
Currently, message from a task is output to all its subscribers but a user may want to send some messages to one of the subscribers while other messages to another. The different subscribers could end up as the same Processor through different Partitioners.

Hence, I'm considering allowing user to `output` to a specific processor / subscriber with `output("b", "p1", message)` or `output("c", "p2", message)` where 

```scala
  val a = Processor[A]("a", 1)
  val b = Processor[B]("b", 1)
  val c = Processor[C]("c", 1)
  val p1 = Partitioner[P1]("p1")
  val p2 = Partitioner[P2]("p2")

  val graph = Graph(a ~ p1 ~> b, a ~ p2 ~> c)
```

"a", "b", "c", "p1" and "p2" are globally unique ids to identify processors and partitioners respectively.

Just my immature thoughts to open further discussions.

kkasravi:
Do you think there is benefit in typing output even if inferred? EG output[T:ClassTag]

manuzhang:
by T you mean the class type of a Processor ?

clockfly:
I think we need to think whether the strong binding here will impact feature "dynamic dag".

manuzhang:
superseded by #1456

PR #1441, opened 2015-09-22
fix #1343, support Storm emitDirect and DirectGrouping

### LINKED ISSUE #1343: support Storm OutputCollector.emitDirect and DirectGrouping

manuzhang:
remaining work

1. #1442 support Storm CustomStreamGrouping and CustomObject
2. full UTs
3. performance test
4. update documentation

all examples from storm-starter should work except `TridentWordCount` because of 1


manuzhang:
Summary of this PR:

1. return the target `TaskId`s from output and allow user to output to a specific `TaskId.
2. expose subscriptions from `TaskActor` to user task through `TaskContext`  and add `getParallelism` to `Subscription`.
3. storm partitioners now also extends `StormPartitioner` which carries downstream component and stream information such that storm tasks could map a subscription to component.
4. #1373 happens because Storm uses `DirectGrouping` to send the number of emitted tuples between two `CoordinatedBolt`s for transaction besides the regular message flow. DirectGrouping is now ignored at graph build time and `emitDirect` is supported through the new output method from 1 and the mapping available from 2, 3.
5. storm has the use case where a processor only outputs to one of its subscribers (#1377) while gearpump processor broadcasts to all. This is resolved by filtering the input message at receiver side. The returned `TaskId`s are also filtered at sender side.

huafengw:
why need to return the target TaskIds from output?

streaming/src/main/scala/io/gearpump/streaming/task/Subscription.scala
```diff
@@ -69,14 +70,18 @@ class Subscription(
...
       }
     case mp: MulticastPartitioner =>
       (msg: Message) => {
         val partitions = mp.getPartitions(msg, parallelism, taskId.index)
-        partitions.map(partition => sendMessage(msg, partition)).sum
+        partitions.foldLeft((List.empty[TaskId], 0)) { (accum, partition) =>
+          val (tasks, msgsSent) = accum
+          val num = sendMessage(msg, partition)
+          (TaskId(processorId, partition) +: tasks) -> (msgsSent + num)
+        }
```
huafengw: In Subscription.scala, if ```life.contains(msg.timestamp)```, then the number of messages were sent is the length of ```partitions```, maybe you could avoid using foldLeft here.

streaming/src/main/scala/io/gearpump/streaming/task/TaskActor.scala
```diff
@@ -109,12 +105,25 @@ class TaskActor(
 
   def onStop() : Unit = task.onStop()
 
-  def output(msg : Message) : Unit = {
-    var count = 0
-    this.subscriptions.foreach{ subscription =>
-      count += subscription._2.sendMessage(msg)
+  def output(taskId: TaskId, msg: Message): Unit = {
+    subscriptions.find(_._1 == taskId.processorId) match {
```
huafengw: change subscriptions to Map?

streaming/src/main/scala/io/gearpump/streaming/task/Subscription.scala
manuzhang: yes for now. My concern is that the return value of sendMessage could be changed to be more than 1 sometime.

manuzhang:
@huafengw requirement from storm's CoordinatedBolt where messages are sent to a specific task

streaming/src/main/scala/io/gearpump/streaming/task/TaskActor.scala
manuzhang: I'm not sure. @clockfly any comments ?

clockfly:
Can we avoid changing the Task interface?

manuzhang:
@clockfly do you mean the existing `output` method ?

clockfly:
I mean this one
https://github.com/gearpump/gearpump/pull/1441/files#diff-6732f2c78b22ca11a7680ec8153c1030

Task belongs to user interface, I am worried about the changes in class Task.

manuzhang:
@clockfly no changes to `Task` interface now.  Please review again.

PR #1443, opened 2015-09-22
Fixes #1383 Enable a general typeclass for processors

### LINKED ISSUE #1383: Enable a general type class for processors
[Story](https://www.pivotaltracker.com/story/show/102310700)
This enables a type class for TypedProcessor. The TypedTask is similar to a SourceTask but doesn't impose a particular API and allows implicit overrides which is what the type class pattern requires.

kkasravi:


kkasravi:
Build failure is unrelated
```
[info] - should accept worker RegisterNewWorker when started *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (25 seconds) during expectMsgClass waiting for class io.gearpump.cluster.MasterToWorker$WorkerRegistered
[info]   at scala.Predef$.assert(Predef.scala:179)
[info]   at akka.testkit.TestKitBase$class.expectMsgClass_internal(TestKit.scala:423)
[info]   at akka.testkit.TestKitBase$class.expectMsgType(TestKit.scala:405)
[info]   at akka.testkit.TestKit.expectMsgType(TestKit.scala:718)
[info]   at io.gearpump.cluster.main.MainSpec$$anonfun$2.apply$mcV$sp(MainSpec.scala:102)
[info]   at io.gearpump.cluster.main.MainSpec$$anonfun$2.apply(MainSpec.scala:82)
[info]   at io.gearpump.cluster.main.MainSpec$$anonfun$2.apply(MainSpec.scala:82)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
```

streaming/src/main/scala/io/gearpump/streaming/task/Task.scala
```diff
@@ -183,4 +186,12 @@ abstract class Task(taskContext : TaskContext, userConf : UserConfig) extends Ta
     case msg =>
       LOG.error("Failed! Received unknown message " + "taskId: " + taskId + ", " + msg.toString)
   }
+}
+
+class TypedTask[T](val typedTask: T, taskContext : TaskContext, userConf : UserConfig) extends Task(taskContext, userConf) {
```
manuzhang: could we have an example for `TypedTask` ?

clockfly:
Why an where is this needed?

kkasravi:
This addresses the need to allow an arbitrary type T to be added to a Task, thus TypedTask[T]. In general we cannot assume all Sources will look like DataSource or all Sinks will look like DataSink and shouldn't need to add new sources, sinks and **nodes** in general by using subtype polymorphism. This will proliferate the types we need to add to core and streaming. For example now we have DataSource, DataSourceTask, DataSourceProcessor, DataSink, DataSinkTask, DataSinkProcessor along with many other similar types in the DSL. A better approach is to use ad-hoc polymorphism via [typeclasses](https://tpolecat.github.io/2015/04/29/f-bounds.html) rather than subtype polymorphism. For example, ATK will have many types [T] that need to be bound to a Task, but we shouldn't need to add KMeans, KMeansTask, KMeansProcessor, RandomForest, RandomForestTask, RandomForestProcessor, etc to core and streaming. See ATK example under gearpump-examples. Note this is not yet integrated into the GearPump DSL where a new parameter OP will need to be defined. This approach has the following advantages:
   1. Avoids subtype proliferation in core and streaming for new types of Tasks
   2. Transparently adds (and serializes) the type [T] in TypedProcessor.taskConf for later retrieval when the TypedTask is initialized
   3. Corrects the definition of Processor.taskClass to be Class[_ <: T]
   4. Could potentially reduce lots of boilerplate code elsewhere

streaming/src/main/scala/io/gearpump/streaming/task/Task.scala
kkasravi: @manuzhang 
See [PR](https://github.com/gearpump/gearpump-examples/pull/16) in gearpump-examples

huafengw:
My concern is that the object T is constructed by reflection, what if the T need some configs when initiated? 
For pull request in https://github.com/gearpump/gearpump-examples/pull/16 , user still need to write a KMeansTask, maybe we can do this follow the DataSource way:
1. Construct the KMeansDataProcessor at client side, and serialize the object into task's userconf.
2. At server side, when a task is launched, it fetch the KMeansDataProcessor from conf.

clockfly:
@kkasravi 

1). For the API compatibility level:

Task and Processor are so fundamental that we should change them unless for absolute reason.
 
2). From the abstraction level:
I saw your example at https://github.com/gearpump/gearpump-examples/pull/16/files#diff-9c272b30b212a6b12db5ba5675cc6612R94

It is strange to me that The T in TypedTask has nothing to do with message type. From the abstraction level, Task is to process message, and maybe the message type is the only thing that make sense as the template type. After looking at the code, it looks like T is actaully a Processor Type?

I don't think this abstraction is right.

3). From the implementation level.
TypedTask[T] will construct the value of type T by reflection, and reading config gearpump.streaming.dsl.processor. 
It is strange that there is no contract or interface definition about for how this processor should be implemented. Then why it is a processor? not a attribute?

What if user want more attribute of different type? Should we define TypedTask[A, B]?
What if the attribute cannot de instantiated by default constructor?

4) summary

From your example, it is more of a application level logic. I don't see there is value to make it part of core. So, I am -1 for this.

kkasravi:
@huafengw 
+1 type [T] is now stored as an instance in userConfig.

@clockfly
1). For the API compatibility level:
When @manuzhang did the Connector framework as suggested by @whjiang he noted that there were a fair amount of redundancy in the DSL and the connectors. These redundancies have been addressed. See below.

@clockfly, @manuzhang 
2). From the abstraction level:
TypedTask[T] was a poor name. I've renamed this to HandlerTask[H] which is more description of what it is doing. Task subtypes DataSourceTask, DataSinkTask were redundant with SourceTask[T,OUT], SinkTask[T]. For example below is DataSinkTask
```
class DataSinkTask (context: TaskContext, conf: UserConfig) extends Task(context, conf) {
  import DataSinkTask._

  private val sink = conf.getValue[DataSink](DATA_SINK).get

    LOG.info("opening data sink...")
    sink.open(context)
  }

  override def onNext(message: Message): Unit = {
    sink.write(message)
  }

  override def onStop(): Unit = {
    LOG.info("closing data sink...")
    sink.close()
  }

}
```
and SinkTask[T]
```
  class SinkTask[T](dataSink: TypedDataSink[T], taskContext: TaskContext, userConf: UserConfig) extends Task(taskContext, userConf) {
    def this(taskContext: TaskContext, userConf: UserConfig) = {
      this(userConf.getValue[TypedDataSink[T]](GEARPUMP_STREAMING_SINK)(taskContext.system).get, taskContext, userConf)
    }

      dataSink.open(taskContext)
    }

      dataSink.write(msg)
    }

    override def onStop() = {
      dataSink.close()
    }
  }
```
Enabling higher-kinded types also eliminates DataSourceProcessor, DataSinkProcessor and the following constants: GEARPUMP_STREAMING_SINK, GEARPUMP_STREAMING_SOURCE. Several warnings in OpTranslator have been eliminated due to unsafe casts.

3). From the implementation level.

HandlerTask[T] establishes a contract that enforces the operations on handler which are called in Task.{onStart, onNext, onStop}

- [x] Avoids subtype proliferation in core and streaming for new types of Tasks
- [x] Adds (and serializes) Processor.taskConf and reifies this Task when the Processor is initialized
- [x] Corrects the definition of Processor.taskClass to be Class[_ <: T]
- [x] Merges SinkTask, DataSinkTask, SourceTask, DataSourceTask
- [x] Removes redundant types DataSourceProcessor, DataSinkProcessor
- [x] Enables the type class pattern on the Task type which allows new Task bindings to be added later

kkasravi:
A good article that explains why the adapter design pattern applies to Processor and Task is [here](https://staticallytyped.wordpress.com/2013/03/24/gang-of-four-patterns-with-type-classes-and-implicits-in-scala-part-2/).
TL;DR 
Task is a general interface we provide to run tasks. Anyone wishing to run a task needs to minimally implement onNext. They may also need to implement onStart for initialization and onStop for termination. Those external tasks whose API we do not control will need an adapter to pair Tasks methods with the external task API. This is done using HandlerTask[T] where T is the handler whose API is called from onStart, onNext and onStop. This enables types of applications. For example HandlerTask[SQL] could enable a general class of applications that are SQL based and need to make some SQL calls in their onNext method. HandlerTask[MQTT] could enable a general class of applications that use MQTT and need to establish some MQTT channels in their onStart and send some MQTT messages in their onNext.

streaming/src/main/scala/io/gearpump/streaming/StreamApplication.scala
```diff
@@ -37,9 +39,11 @@ trait Processor[+T <: Task] extends ReferenceEqual {
 
   def description: String
 
-  def taskClass: Class[_ <: Task]
+  def taskClass: Class[_<:T]
```
clockfly: This is wrong.

1. 
For example, if we have task ABCTask, then the taskClass should return Class[ABCTask] instead of Class[? <: ABCTask]

2. The reason that we don't use Class[T] but Class[_ <: Task] is that Class is a class defined in java, which don't have the concept of covariant or contravariant. By erasuing the actual task type, it gives us some flexibility.

clockfly:
I think we should break it into two problems:
1. whether we need to change the core interface. 
2. Whether the core need need to be improved to remove some duplicate code.

For the second problems, I agree that some improvement needed to be done about the type system about in DSL API, as you stated.

For the first problem, 
I don't think the change belongs to the core API, it is better to put it in your application logic. Here is my reason:

1.It will not work. what if I need to define two "handlerTask" in same DAG, will them conflict with each other? Seems they are using same configuration.
2. It is not generic. You made a point about Task[MQTT] and Task[SQL], how about my task need to interact with both MQTT and SQL? Then should we extends the core API again?
3. There is no necessity to change the core API. Take the MQTT and SQL example, the application have full power to do whatever it want without the core API "handlertask". If you want to save the code, then maybe puting it as a application util is better.

clockfly:
For 
```
1). For the API compatibility level:
When @manuzhang did the Connector framework as suggested by @whjiang he noted that there were a fair amount of redundancy in the DSL and the connectors. These redundancies have been addressed. See below.
```

I think the SinkTask used in DSL belongs to the implementation but not API, what user see is DataSink, which is consistent across the graph API and DSL API.

kkasravi:
Thanks for the feedback @clockfly. Below is my feedback.
```
I think we should break it into two problems:
1. whether we need to change the core interface. 
```
+1. The core interface - if changed at all - should be done once so all subsequent Tasks will not need to modify anything in core. The only change to core (if we do not consider DataSource, DataSink) is TypedProcessor, HandlerTask and Processor.apply. From these 3 changes no additional core changes are required. These could be called AdapterProcessor and AdapterTask (see my last comment below) since they are providing an [Adapter](http://geekswithblogs.net/subodhnpushpak/archive/2009/09/18/the-23-gang-of-four-design-patterns-.-revisited.aspx) design pattern using typeclasses. The remaining issue is whether we should provide a separate Processor.apply. This also is not technically necessary (see below).
```
2. Whether the DSL implementation need need to be improved to remove some duplicate code.
```
+1 We should remove redundant code within the DSL and make it easier to extend the DSL.
```
1.It will not work. what if I need to define two "handlerTask" in same DAG, will them conflict with each other? Seems they are using same configuration.
```
Each handlerTask has it's own taskConf. 
```
2. It is not generic. You made a point about Task[MQTT] and Task[SQL], how about my task need to interact with both MQTT and SQL? Then should we extends the core API again?
```
For MQTT and SQL one would define a class that implemented both MQTT and SQL and then override methods as needed. But this does not change the core API. Things like ATK, MQTT, SQL would be defined in the application or external or a separate library altogether. They are all resolved via implicit type resolution. If we needed something that combined DataSource and DataSink how would we do it? It's a similar problem. However it would require changes to the DSL and a new subdirectory under streaming called source_sink or something. I thing the goal is the same the design choice is between subtyping or ad-hoc polymorphism. 
```
3. There is no necessity to change the core API. Take the MQTT and SQL example, the application have full power to do whatever it want without the core API "handlertask". If you want to save the code, then maybe puting it as a application util is better.
```
My suggestion is I will define something like Adapter.scala which will hold AdapterProcessor AdapterTask and an implicit that could take Processor and return TypedProcessor which would define its own apply. However it would be desired to have this class in core somewhere but not critical. I can place this in gearpump/gearpump-adapter or gearpump/gearpump-atk for now. I will close this PR and the related issue.

PR #1444, opened 2015-09-25
fix #1343, support Storm emitDirect and DirectGrouping

### LINKED ISSUE #1343: support Storm OutputCollector.emitDirect and DirectGrouping

manuzhang:
previously in #1441.

Summary of this PR,

1. supports all types of Storm Grouping (including CustomStreamGrouping)
2. covers the following cases not natively supported by gearpump
      * sends message to a specific task
      * returns target tasks to storm
      * a processor has two edges (streams in storm) to another processor
      * a processor only sends message to one of its subscriber rather than broadcasts to all
3. no changes to `Task` interface

Remaining issue I'd like to fix in follow-ups,

1. trident example is not working yet
2. user provided config file is not read at storm side

More future tasks,

1. full UTs
2. performance test
3. documentation


experiments/storm/src/main/scala/io/gearpump/experiments/storm/partitioner/StormPartitioner.scala
```diff
@@ -18,18 +18,21 @@
...
 package io.gearpump.experiments.storm.partitioner
 
-import clojure.lang.RT
 import io.gearpump.Message
-import io.gearpump.partitioner.UnicastPartitioner
+import io.gearpump.experiments.storm.util.StormUtil
+import io.gearpump.partitioner.{MulticastPartitioner, UnicastPartitioner}
 
-import scala.util.Random
+class StormUnicastPartitioner(target: String) extends UnicastPartitioner {
```
clockfly: What "target" mean here?
Can you add an comment in source code?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/GraphBuilder.scala
```diff
@@ -36,118 +37,90 @@ import scala.collection.JavaConversions._
...
-    spoutTasks.foreach { case (sourceId, sourceProducer) =>
-      val sourceSpec = spouts.get(sourceId)
-      getTargets(sourceId, topology).foreach { case (streamId, targets) =>
-        val outFields = new Fields(sourceSpec.get_common().get_streams().get(streamId).get_output_fields())
-        targets.foreach { case (targetId, grouping) =>
-          boltTasks.get(targetId).foreach { targetProcessor =>
-            processorGraph.addEdge(sourceProducer, groupingToPartitioner(outFields, grouping), targetProcessor)
-          }
-        }
+    val spoutProcessors = spouts.map { case (id, spout) => id -> spoutToProcessor(id, spout) }.toMap
```
clockfly: Can you break down this function to smaller private ones to make the purpose explicit?

streaming/src/main/scala/io/gearpump/streaming/task/Subscription.scala
```diff
@@ -100,7 +100,7 @@ class Subscription(
   def sendMessage(msg: Message, partition: Int): Int = {
 
     // only send message whose timestamp matches the lifeTime
-    if (life.contains(msg.timestamp)) {
+    if (partition >= 0 && life.contains(msg.timestamp)) {
```
clockfly: Is partition expected to have minus value?
Will require(http://www.scala-lang.org/api/2.10.2/index.html#scala.Predef$) better?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/StormTuple.scala
```diff
@@ -18,17 +18,26 @@
...
-  def toTuple(topologyContextBuilder: TopologyContextBuilder): Tuple = {
-    new TupleImpl(topologyContextBuilder.buildContext(taskId, componentId), tuple, taskId, streamId, null)
+    streamId: String,
+    stormTaskId: Int,
+    @transient targetPartitions: Map[String, List[Int]]) {
+
+  def toTuple(topology: StormTopology, stormConfig: JMap[_, _],
+      taskToComponent: JMap[Integer, String], componentToTasks: JMap[String, JList[Integer]],
+      componentToStreamFields: JMap[String, JMap[String, Fields]]): Tuple = {
+    val topologyContext = StormUtil.buildTopologyContext(topology, stormConfig, taskToComponent, componentToTasks,
```
clockfly: Do we need to construct buildTopologyContext for every call of toTuple?

clockfly: It looks like this will be called for every message.

experiments/storm/src/main/scala/io/gearpump/experiments/storm/processor/StormProcessor.scala
```diff
@@ -73,15 +78,21 @@ private[storm] class StormProcessor(taskContext : TaskContext, conf: UserConfig)
       })
     }
```
clockfly: Why is it different for getTickTupleContext and normal Context?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/processor/StormProcessor.scala
```diff
@@ -73,15 +78,21 @@ private[storm] class StormProcessor(taskContext : TaskContext, conf: UserConfig)
...
     val delegate = new StormBoltOutputCollector(outputCollector)
-    val topologyContext = topologyContextBuilder.buildContext(pid, boltId)
     bolt.prepare(stormConfig, topologyContext, new OutputCollector(delegate))
   }
 
   override def onNext(msg: Message): Unit = {
     val stormTuple = msg.msg.asInstanceOf[StormTuple]
-    outputCollector.setTimestamp(msg.timestamp)
-    bolt.execute(stormTuple.toTuple(topologyContextBuilder))
+    val tuple = stormTuple.toTuple(topology, stormConfig, taskToComponent.asJava, componentToSortedTasks,
```
clockfly: stormTuple.toTuple looks like have the same dependency of TopologyContext?

```
buildTopologyContext(topology, stormConfig, taskToComponent.asJava, componentToSortedTasks,
+      componentToStreamFields, boltId, stormTaskId, bolt.isInstanceOf[ShellBolt])
```

Shoould we pass the Topologycontext directly?

Will every new tuple need a new TopologyContext?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/processor/StormProcessor.scala
```diff
@@ -47,17 +52,17 @@ private[storm] class StormProcessor(taskContext : TaskContext, conf: UserConfig)
...
-  private val topologyContextBuilder = TopologyContextBuilder(topology, stormConfig,
-    multiLang = bolt.isInstanceOf[ShellBolt])
-  private val outputCollector = new StormOutputCollector(taskContext, pid, boltId)
+  private val taskToComponent = conf.getValue[Map[Integer, String]](TASK_TO_COMPONENT).get
+  private val componentToSortedTasks = getComponentToSortedTasks(taskToComponent)
+  private val stormTaskId = getStormTaskId(boltId, componentToSortedTasks, taskContext.taskId)
+  private val componentToStreamFields = getComponentToStreamFields(topology)
+  private val inputs: JMap[GlobalStreamId,  Grouping] = StormUtil.getInputs(boltId, topology)
+  private var outputCollector: StormOutputCollector = null
   private var scheduler: Cancellable = null
```
clockfly: The initialization list seems pretty long to me, do we need them all?

Whether it is a good idea to use TopologyContext to represent all? In that case, we can avoid passing so many values.

experiments/storm/src/main/scala/io/gearpump/experiments/storm/processor/StormProcessor.scala
```diff
@@ -73,15 +78,21 @@ private[storm] class StormProcessor(taskContext : TaskContext, conf: UserConfig)
...
     bolt.prepare(stormConfig, topologyContext, new OutputCollector(delegate))
   }
 
   override def onNext(msg: Message): Unit = {
     val stormTuple = msg.msg.asInstanceOf[StormTuple]
-    outputCollector.setTimestamp(msg.timestamp)
-    bolt.execute(stormTuple.toTuple(topologyContextBuilder))
+    val tuple = stormTuple.toTuple(topology, stormConfig, taskToComponent.asJava, componentToSortedTasks,
+      componentToStreamFields)
+    if (inputs.containsKey(tuple.getSourceGlobalStreamid)) {
```
clockfly: Is this check necessary? It may impact perfomance as basically you are doing Set.contain for every message.

experiments/storm/src/main/scala/io/gearpump/experiments/storm/processor/StormProcessor.scala
```diff
@@ -91,8 +102,8 @@ private[storm] class StormProcessor(taskContext : TaskContext, conf: UserConfig)
   }
 
   override def receiveUnManagedMessage: Receive = {
-    case tickTuple: TupleImpl =>
-      bolt.execute(tickTuple)
+    case tuple: Tuple if isTickTuple(tuple) =>
```
clockfly: Why not wrapping Tuple inside Message?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormProducer.scala
```diff
@@ -18,32 +18,35 @@
...
   import StormUtil._
 
-  private val topology = getTopology(conf)
-  private val pid = taskContext.taskId.processorId
-
-  private val spoutId = conf.getString(GraphBuilder.COMPONENT_ID)
-      .getOrElse(throw new RuntimeException(s"Storm spout id not found for processor $pid"))
-  private val spoutSpec = conf.getValue[SpoutSpec](GraphBuilder.COMPONENT_SPEC)
-      .getOrElse(throw new RuntimeException(s"Storm spout spec not found for processor $pid"))
+  private val topology = conf.getValue[StormTopology](TOPOLOGY).get
```
clockfly: This list seems pretty long.

Is it possible for we to wrap all in to one?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/GraphBuilder.scala
```diff
@@ -36,118 +37,90 @@ import scala.collection.JavaConversions._
...
+    val boltProcessors = bolts.map { case (id, bolt) => id -> boltToProcessor(id, bolt) }.toMap
+
+    spouts.foreach { case (sourceId, spout) =>
+      val sourceProcessor = spoutProcessors(sourceId)
+      val sourceCommon = spout.get_common()
+      val sourceParallelism = getParallelism(sourceCommon)
+      val tasks = getTasks(sourceId, taskId, sourceParallelism)
```
clockfly: To be honest, I don't like the way you generate the task id.
It is very complicated. 

Have you thought about convert type TaskId to a Int?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/Grouper.scala
```diff
@@ -0,0 +1,105 @@
...
+import backtype.storm.generated.GlobalStreamId
+import backtype.storm.grouping.CustomStreamGrouping
+import backtype.storm.task.TopologyContext
+import backtype.storm.tuple.Fields
+import clojure.lang.RT
+
+import scala.util.Random
+import scala.collection.JavaConversions._
+
+case class GroupingContext(topologyContext: TopologyContext, globalStreamId: GlobalStreamId, sourceTasks: JList[Integer],
```
clockfly: Why this is needed? better name?

It doesn't feel like a "Context" to me..

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/Grouper.scala
```diff
@@ -0,0 +1,105 @@
...
+      outFields: Fields, taskToPartition: Map[Integer, Int])
+
+trait Grouper {
```
clockfly: Better add a description about the purpose of this class.

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/Grouper.scala
```diff
@@ -0,0 +1,105 @@
...
+  def getPartition(taskId: Int, values: JList[AnyRef]): Int
+  def getPartitions(taskId: Int, values: JList[AnyRef]): List[Int] = {
```
clockfly: This implementation make no sense to me. As getPartitions: List[Int] is more general than getPartition: Int

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/Grouper.scala
```diff
@@ -0,0 +1,105 @@
...
+  }
+}
+
+class GlobalGrouper extends Grouper {
+  override def getPartition(taskId: Int, values: JList[AnyRef]): Int = 0
+}
+
+class NoneGrouper(numTasks: Int) extends Grouper {
+  val random = new Random
+  lazy val mod = RT.`var`("clojure.core", "mod")
```
clockfly: Can we replace this with pure scala?b

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/Grouper.scala
clockfly: And what you expect user to use eventually? 

Do you expect user to call getPartiton: Int or getPartitions: List[Int] ???

What if these two are in-consistent, which one is prefered?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/StormOutputCollector.scala
```diff
@@ -18,34 +18,78 @@
...
+    if (targets.containsKey(streamId)) {
+      val target = taskToComponent(id)
+      val partition = taskToPartition(id)
+      val targetPartitions = Map(target -> List(partition))
+      val tuple = StormTuple(values.toList, componentId, streamId, stormTaskId, targetPartitions)
+      context.output(Message(tuple, timestamp))
+    }
   }
 
+  def fail(tuple: Tuple): Unit = {}
```
clockfly: Can you comment why here it is empty?
Is this expected? or a TODO? 

Will there impacts?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/StormTuple.scala
```diff
@@ -18,17 +18,26 @@
...
 
-import backtype.storm.tuple.{Tuple, TupleImpl}
+import java.util.{List => JList, Map => JMap}
+
+import backtype.storm.generated.StormTopology
+import backtype.storm.tuple.{Fields, Tuple, TupleImpl}
 
 import scala.collection.JavaConversions._
 
 private[storm] case class StormTuple(
```
clockfly: Why case class?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/StormUtil.scala
```diff
@@ -75,4 +81,185 @@ object StormUtil {
...
+    destDir + File.separator + "resources"
+  }
+
+  def buildTopologyContext(topology: StormTopology, stormConf: JMap[_, _],
+      taskToComponent: JMap[Integer, String], componentToSortedTasks: JMap[String, JList[Integer]],
+      componentToStreamFields: JMap[String, JMap[String, Fields]],
+      componentId: String, stormTaskId: Int, multiLang: Boolean = false): TopologyContext = {
+    var codeDir: String = null
+    var pidDir: String = null
+    if (multiLang) {
```
clockfly: Why only multilang require codeDir or pidDir?

for non-multilang app, can user see the two fields in API? What is the expected behavior when user want to use them?

clockfly:
I doubt this patch will have serious performance degrading. Please do some performance verification.

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/GraphBuilder.scala
clockfly: And this is also not stable. What if I added a new processor in runtime?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/processor/StormProcessor.scala
manuzhang: tick tuple is system tuple which uses system stream, task and component

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/GraphBuilder.scala
manuzhang: I'm afraid it's not possible with such things as `CoordinatedBolt`. `CoordinatedBolt` needs to know both downstream and upstream storm task ids, which cannot be easily worked out given the current interfaces of Gearpump. That's why I generate storm task ids at graph build time.

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/Grouper.scala
manuzhang: clojure `mod` is different from scala `%` and I'm not sure how `mod` is implemented in clojure

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/StormUtil.scala
manuzhang: multilang needs `codeDir` for source codes for other languages like `splitsentence.py` and `pidDir` for subprocess to run those codes. users shoud not see them. I'd better hide those fields at all.

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/GraphBuilder.scala
clockfly: Can you do a mapping like this?
TaskId(processorId, taskIndex)

map processorId -> int16
map taskIndex -> int16

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/Grouper.scala
clockfly: Check the doc of Non-grouping at 
```
https://storm.apache.org/documentation/Concepts.html
```

I think it is OK to use ShuffleGrouping to replace it.

clockfly: Check https://clojuredocs.org/clojure.core/mod

basically, the logic is same as
```
(-13 % 5 + 5) % 5
```

streaming/src/main/scala/io/gearpump/streaming/task/Subscription.scala
manuzhang: gearpump broadcasts messages to all subscribers which will cause error at storm's side. I'm trying to solve this by returning partition -1 for those unintended subscribers and filter it out when `sendMessage`

manuzhang:
@clockfly I've addressed most of your comments. I'll follow up with performance verification.

streaming/src/main/scala/io/gearpump/streaming/task/Subscription.scala
clockfly: Where you return partition = -1? 
Can you make that explicit so that we don't accidentially delete the check in future refactory?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/processor/StormProcessor.scala
```diff
@@ -19,96 +19,89 @@
...
+
+    setupTickTuple(topologyContext)
   }
 
   override def onNext(msg: Message): Unit = {
     val stormTuple = msg.msg.asInstanceOf[StormTuple]
     outputCollector.setTimestamp(msg.timestamp)
-    bolt.execute(stormTuple.toTuple(topologyContextBuilder))
+    val sourceId = stormTuple.stormTaskId
+    val topologyContext = taskToTopologyContext.getOrElse(sourceId, topologyContextBuilder.build(sourceId))
```
clockfly: This will call topologyContextBuilder.build(sourceId) everytime/

clockfly: Why will you create different topologyContext? Is this different from the toplogyContext in onStart?

clockfly: It doesn't make sense to have multiple ToplogyContext in single task.

experiments/storm/src/main/scala/io/gearpump/experiments/storm/processor/StormProcessor.scala
```diff
@@ -19,96 +19,89 @@
...
-    val tasks = new JArrayList[Integer](1)
-    tasks.add(SYSTEM_TASK_ID)
-    componentToSortedTasks.put(SYSTEM_COMPONENT_ID, tasks)
-    val streamToFields = new JHashMap[String, Fields]
-    streamToFields.put(SYSTEM_STREAM_ID, new Fields(SYSTEM_COMPONENT_OUTPUT_FIELDS))
-    val componentToStreamToFields = new JHashMap[String, JMap[String, Fields]]
-    componentToStreamToFields.put(SYSTEM_COMPONENT_ID, streamToFields)
-    new GeneralTopologyContext(topology, stormConfig, taskToComponent, componentToSortedTasks,
-      componentToStreamToFields, null)
+  private def setupTickTuple(topologyContext: TopologyContext): Unit = {
```
clockfly: This is ugly.

Is Tick a constant message?

clockfly: Maybe we can create a constant mesasge Tick somewhere.

val Tick = 

And changing this function to something like scheduleTick
And for flow control purpose, maybe we should use scheduleOnce instead.

experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormProducer.scala
```diff
@@ -18,32 +18,37 @@
...
 
   override def onStart(startTime: StartTime): Unit = {
-    val topologyContext = topologyContextBuilder.buildContext(pid, spoutId)
+    implicit val timeout = Constants.FUTURE_TIMEOUT
+    val dagFuture = askAppMasterForDAG(taskContext.appMaster)
+    val dag = Await.result(dagFuture, timeout.duration)
+    val taskToComponent = getTaskToComponent(dag)
+    val topologyContextBuilder = new TopologyContextBuilder(topology, stormConfig, taskToComponent)
+    val topologyContext = topologyContextBuilder.build(stormTaskId)
+    collector = new StormOutputCollector(taskContext, topologyContext)
```
clockfly: I believe the topology part is same with StormProcessor. Can we resue the code?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/GraphBuilder.scala
```diff
@@ -18,136 +18,105 @@
...
-        val inputStreamId = globalStreamId.get_streamId()
-        val inputComponentId = globalStreamId.get_componentId
-        if (inputComponentId.equals(componentId)) {
-          val curr = allTargets.getOrElse(inputStreamId, Map.empty[String, Grouping])
-          allTargets + (inputStreamId -> (curr + (otherComponentId -> grouping)))
-        } else {
-          allTargets
-        }
-    }
+  def getTasks(sourceId: String, startId: Integer, num: Int): JList[Integer] = {
```
clockfly: This function is not used.

experiments/storm/src/main/scala/io/gearpump/experiments/storm/util/StormUtil.scala
```diff
@@ -75,4 +72,205 @@ object StormUtil {
...
+    }
+  }
+
+  def isGlobalGrouping(grouping: Grouping): Boolean = {
+    grouping.getSetField == Grouping._Fields.FIELDS &&
+        grouping.get_fields.isEmpty
+  }
+
+  def getParallelism(component: ComponentCommon): Int = {
+    val componentConf = parseJsonStringToMap(component.get_json_conf)
```
clockfly: This part seems unclear to me, I don't know what ComponentCommon means, and what the schema of the json_conf used in this function.

I think we will want some clear "line" between storm part and gearpump part. 

One way is to define a wrapper for StormTopology, like GearpumpStormTopology(storm: StormTopology) {
  def raw: StormTolology
}

and define some useful function name that is more clear in this context.

clockfly: About the clear "line", I think the design of Grouper  is a good example.

manuzhang:
@clockfly updated. Please review.

clockfly:
Good work

+1

PR #1446, opened 2015-09-25
fix #1445 refactor serialization implementation

### LINKED ISSUE #1445: Refactor Gearpump's serialization implementation in transport layer

huafengw:
This is a sketch for #1445 and the code is need to be refined. I raised it for the advise of interface design. Please ignore the code style issue.


clockfly:
Can you use the old class layout?

Like AckRequest, AckRequestSeriaizer.

streaming/src/main/scala/io/gearpump/streaming/task/InitialAckRequest.scala
```diff
@@ -0,0 +1,51 @@
...
+import io.gearpump.transport.netty.TransportMessage
+import org.jboss.netty.buffer.ChannelBuffer
+
+/*
+ * Initial AckRequest
+ */
+case class InitialAckRequest(taskId: TaskId, sessionId: Int) extends TransportMessage{
+  override def getLength(): Int = taskId.getLength() + 8 // classId + sessionId
+
+  override def getClassId: Int = InitialAckRequest.ID
```
clockfly: Can we avoid defining getClassId?

streaming/src/main/scala/io/gearpump/streaming/task/InitialAckRequest.scala
```diff
@@ -0,0 +1,51 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gearpump.streaming.task
+
+import java.io.DataOutput
+import io.gearpump.streaming.task.AckRequest._
+import io.gearpump.transport.netty.TransportMessage
```
clockfly: It doesn't look good to depends on TransportMessage.

How about define a getLength call in Serializer?

huafengw: When serializing the objects, we have to know the exact length of buffer to allocate. The serializer has to serialize the object so that it would know the length of the object.

streaming/src/main/resources/reference.conf
```diff
@@ -3,12 +3,6 @@ gearpump {
   serializers {
     ## Follow this format when adding new serializer for new message types
     ##    "io.gearpump.Message" = "io.gearpump.streaming.MessageSerializer"
```
whjiang: Better remove this comment line as it is out of date.

streaming/src/main/scala/io/gearpump/streaming/MessageSerializer.scala
```diff
@@ -18,96 +18,92 @@
...
+import org.jboss.netty.buffer.ChannelBuffer
 
-  override def read(kryo: Kryo, input: Input, typ: Class[Message]): Message = {
-    val timeStamp = input.readLong()
-    val msg = kryo.readClassAndObject(input)
-    new Message(msg, timeStamp)
-  }
-}
+class TaskIdSerializer extends TaskMessageSerializer[TaskId] {
+  override def getLength(obj: TaskId): ProcessorId = 8
```
whjiang: Why return type as `ProcessorId` instead of `Int`?

streaming/src/main/scala/io/gearpump/streaming/MessageSerializer.scala
```diff
@@ -18,96 +18,92 @@
...
     new TaskId(processorId, index)
   }
 }
 
-class AckRequestSerializer extends Serializer[AckRequest] {
-  val taskIdSerialzer = new TaskIdSerializer()
+class AckSerializer extends TaskMessageSerializer[Ack] {
+  val taskIdSerializer = new TaskIdSerializer
+
+  override def getLength(obj: Ack): ProcessorId = taskIdSerializer.getLength(obj.taskId) + 8
```
whjiang: same comment for return type.

streaming/src/main/scala/io/gearpump/streaming/task/StreamingTransportSerializer.scala
```diff
@@ -0,0 +1,68 @@
...
+  serializers.register(classOf[Ack], new AckSerializer)
+  serializers.register(classOf[AckRequest], new AckRequestSerializer)
+  serializers.register(classOf[InitialAckRequest], new InitialAckRequestSerializer)
+  serializers.register(classOf[LatencyProbe], new LatencyProbeSerializer)
+  serializers.register(classOf[SerializedMessage], new SerializedMessageSerializer)
+
+  override def serialize(dataOutput: DataOutput, obj: Object): Unit = {
+    val registration = serializers.getRegistration(obj.getClass)
+    if(registration != null) {
+      dataOutput.writeInt(registration.id)
```
whjiang: How to ensure that the registration ID is consistent for the whole cluster? I mean, the registration id is generated according to register order. However, how to ensure each node has the same registration order? It is highly possible that some node may have different orders. Maybe we shall put such `SerializerResolver` as part of `AppMaster` so that, for one app, we can have consistent ordering. And of course, as the order is monotonic, you can cache at each worker to speed up.

core/src/test/scala/io/gearpump/transport/MockTransportSerializer.scala
```diff
@@ -0,0 +1,43 @@
...
+ */
+package io.gearpump.transport
+
+import java.io.DataOutput
+
+import io.gearpump.transport.MockTransportSerializer.NettyMessage
+import io.gearpump.transport.netty.ITransportMessageSerializer
+import org.jboss.netty.buffer.ChannelBuffer
+
+class MockTransportSerializer extends ITransportMessageSerializer{
```
whjiang: why do we need a specific `MockTransportSerializer`? Can't we use `StreamingTransportSerializer` in general?

streaming/src/main/scala/io/gearpump/streaming/task/SerializerResolver.scala
```diff
@@ -0,0 +1,30 @@
+package io.gearpump.streaming.task
+
+import io.gearpump.esotericsoftware.kryo.util.ObjectMap
+import io.gearpump.esotericsoftware.kryo.util.IntMap
+import io.gearpump.streaming.task.SerializerResolver.Registration
+
+private[task] class SerializerResolver {
```
whjiang: Why it is `private`?

How do user to register for his own serializer? Can we provide user mechanism to register some new serializer for this solver? 

And further, is it possible to register some general serializer? E.g. `java.io.Serializable` or `hadoop.io.Writable` or kyro? That is, if there is a specific serializer, we can use it. otherwise, we can check whether there is any general serializer suitable for this message type? If so, we can use it as a last refugee instead of throw error.

core/src/test/scala/io/gearpump/transport/MockTransportSerializer.scala
huafengw: This class is in folder ```test``` because the ```NettySpec``` is under ```core``` while ```StreamingTransportSerializer``` in under ```streaming```. So the test need a mock implementation.

streaming/src/main/scala/io/gearpump/streaming/task/SerializerResolver.scala
huafengw: Hi Weihua, generally this refactor is transparent to user. 
We have 5 type of messages will be transported by the netty layer, including ```Message Ack AckRequest InitialAckRequest LatencyProbe```. However, user's message is boxed in ```Message``` and it will still be serialized by Kryo, which means user only need to care about registering their serializers when running a streaming application, so the ```SerializerResolver``` is actually only be used in Gearpump internally.

streaming/src/main/scala/io/gearpump/streaming/task/TaskControlMessage.scala
```diff
@@ -20,23 +20,24 @@ package io.gearpump.streaming.task
 
 import io.gearpump.TimeStamp
 
-
 /*
- * Initial AckRequest
+  Here the seq field represents the expected number of received messages
+    and the actualReceivedNum field means the actual received number since start
  */
```
clockfly: Unnecessary change here

streaming/src/main/scala/io/gearpump/streaming/task/TaskControlMessage.scala
```diff
@@ -56,6 +57,4 @@ object GetStartClock
 
 case class StartClock(clock: TimeStamp)
 
-case class LatencyProbe(timestamp: Long)
```
clockfly: Where is this message?

huafengw: Still there, in the same file.

core/src/main/java/io/gearpump/transport/netty/TaskMessage.java
```diff
@@ -60,6 +58,6 @@ public boolean equals(Object obj)  {
     return this.sessionId() == other.sessionId()
```
clockfly: PLease check whether the equal is used anymore.

If not, please remove it.

streaming/src/main/resources/reference.conf
```diff
@@ -1,16 +1,5 @@
 gearpump {
 
-  serializers {
```
clockfly: Why removing this config?

I think GearpumpSerialization is still reading tihs configuration for custom messages.

huafengw: This config is not exposed to user because it's in streaming's jar file. User still can config this in gear.conf

clockfly:
+1

PR #1448, opened 2015-09-28
fix #809: static resources are now sent to client gzipped

### LINKED ISSUE #809: Investigate compressed response
Make sure we can compress server responses (in particular large json objects).
http://spray.io/documentation/1.2.2/spray-routing/encoding-directives/compressResponse/

stanleyxu2005:
This patch will reduce dashboard js files from 1MB to 300kb


clockfly:
+1

PR #1452, opened 2015-09-30
Added support for composing DAG from scratch

stanleyxu2005:
Also changed other dialogs form control style (regarding new css styles)
Also switched the position of "Cancel" and "Save" button in all dialogs
Also replaced the library underscore with lodash, which is 100% compatible but has more utils.


clockfly:
+1

PR #1454, opened 2015-09-30
fix #1453: allows use DEL to remove processor/edge

### LINKED ISSUE #1453: Support pressing DEL key to delete processor/edge

stanleyxu2005:
also makes node label consistent as it shows in application's dag view
also fixed the click behavior on a disabled button (span button)


clockfly:
+1

PR #1461, opened 2015-10-07
Fixes #1460 Provide YARN related option that can disable creating the…

### LINKED ISSUE #1460: Provide YARN related option that can disable creating the services UI when the YARN application master creates the Master and Workers
For integration with TAP we need to be able to NOT spawn the services UI, since this will be created within TAP as a user facing process by the GearPump broker at the same time the broker creates YARN GearPump applicationmaster

kkasravi:
… services UI when the YARN application master creates the Master and Workers

### DIFF
--- a/conf/yarn.conf
+++ b/conf/yarn.conf
@@ -37,6 +37,7 @@ gearpump {
     logname = "worker.log"
   }
   services {
+    enabled = false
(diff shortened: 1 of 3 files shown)


clockfly:
+1

PR #1467, opened 2015-10-08
fix #1463, fix storm config

### LINKED ISSUE #1463: storm config file is not read
the config file is set to the system property `storm.conf.file` at client side, which is not put on the classpath when launching executors

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/gearpump/gearpump/1467)


clockfly:
+1

PR #1477, opened 2015-10-13
fix #1476 fix serialization in transport layer

### LINKED ISSUE #1476: serialization works incorrectly in cluster mode

huafengw:


stanleyxu2005:
+1

core/src/main/scala/io/gearpump/transport/netty/Client.scala
```diff
@@ -200,11 +200,11 @@ object Client {
     }
   }
 
-  class ClientPipelineFactory(name: String, decoder: MessageDecoder, encoder: MessageEncoder) extends ChannelPipelineFactory {
+  class ClientPipelineFactory(name: String, conf: NettyConfig) extends ChannelPipelineFactory {
```
manuzhang: I think the original interface is better for testing where decoder and encoder could be mocked. Same holds for ServerPipelineFactory

stanleyxu2005: He wants factory produce encoder decoder internally on demand. There is a thread unsafe issue.

manuzhang: With `conf.newTransportSerializer`, each client has its own serializer for ChannelPipeline. Also, I wonder why a client requires a decoder and server an encoder.

huafengw: There is no test cover for the pipeline factory, but there is a Netty.spec which will test the Client and Server, so the pipeline factory is embedded.

PR #1483, opened 2015-10-19
fix #1482, fix TaskScheduler over schedules when executors down

### LINKED ISSUE #1482: TaskScheduler over schedules when multiple executors down at once

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/gearpump/gearpump/1483)


manuzhang:
One issue I found is that the same worker is allocated when restarting multiple crashed executors with `ONEWORKER` relaxation

clockfly:
+1

PR #1484, opened 2015-10-19
fix #1481, fix storm performance regression

### LINKED ISSUE #1481: storm over gearpump doesn't perform well
I've found several hot spots

1. We are now using storm's kryo serializer where GearpumpTuple is not registered and is serded by java serializer. The solution is to replace storm's kryo serializer and register GearpumpTuple.
2. Each StormOutputCollector.emit call will create a new grouper, which should be created beforehand.
3. GearpumpTuple.toTuple is not free and is invoked on every received message. We could implement a customized `Tuple` instead of `ITuple` but the benefit remains to be tested out.

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/gearpump/gearpump/1484)

### DIFF
--- a/experiments/storm/src/main/scala/io/gearpump/experiments/storm/topology/GearpumpTuple.scala
+++ b/experiments/storm/src/main/scala/io/gearpump/experiments/storm/topology/GearpumpTuple.scala
@@ -18,22 +18,24 @@
 
 package io.gearpump.experiments.storm.topology
 
+import java.lang.{Integer => JInteger}
+import java.util.{List => JList}
+
 import backtype.storm.task.TopologyContext
 import backtype.storm.tuple.{Tuple, TupleImpl}
 
-import scala.collection.JavaConversions._
 
 private[storm] class GearpumpTuple(
-    tuple: List[AnyRef],
-    componentId: String,
-    streamId: String,
-    stormTaskId: Int,
+    tuple: JList[AnyRef],
+    sourceTaskId: JInteger,
+    sourceStreamId: String,
     @transient val targetPartitions: Map[String, List[Int]]) extends Serializable {
 
+  def this() = this(null, null, null, null)
+
   def toTuple(topologyContext: TopologyContext): Tuple = {
-    new TupleImpl(topologyContext, tuple, stormTaskId, streamId, null)
+    new TupleImpl(topologyContext, tuple, sourceTaskId, sourceStreamId)
   }
 }
-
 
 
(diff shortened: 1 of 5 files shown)


experiments/storm/src/main/scala/io/gearpump/experiments/storm/topology/GearpumpTuple.scala
```diff
@@ -18,22 +18,25 @@
 
 package io.gearpump.experiments.storm.topology
 
+import java.lang.{Integer => JInteger}
+
 import backtype.storm.task.TopologyContext
-import backtype.storm.tuple.{Tuple, TupleImpl}
+import backtype.storm.tuple.{TupleImpl, Tuple}
+
+import java.util.{List => JList}
```
stanleyxu2005: Are you sure all imports are sorted by IDE?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/topology/GearpumpTuple.scala
```diff
@@ -18,22 +18,25 @@
 
 package io.gearpump.experiments.storm.topology
 
+import java.lang.{Integer => JInteger}
```
stanleyxu2005: In this PR, as a new code reader, I miss the context, why we need to create a name alias for Java primitive types. Can you describe in code comment or somewhere else (e.g. in the PR)

manuzhang: To explicitly distinguish it from scala AnyVal types

stanleyxu2005:
+1

manuzhang:
The changes include

1. replaces storm's `DefaultKryoFactory` with gearpump's `StormKryoFactory` such that we have a chance to register `GearpumpTuple` in Kryo. Note that we only register `GearpumpTuple` when the storm configuration `topology.fall.back.on.java.serialization` is false, when we are sure values in `GearpumpTuple` could be serialized with Kryo. Otherwise, `GearpumpTuple` will be serialized with Java serialization which has bad performance. 
2. initializes grouper beforehand rather than on each `emit`
3. removes `componentId` from `GearpumpTuple` since it can be retrieved from `TopologyContext`
4. adds `application.conf` so storm applications use `StormKryoPool` by default.

The performance is on par with our previous tests where Storm over Gearpump has a throughput of 4,500,000 messages/s on a three-node cluster running SOL benchmark (36 producers, 36 processors).

clockfly:
+1

PR #1486, opened 2015-10-20
fix #1479: add uptime field in service interface

### LINKED ISSUE #1479: incorrect "Up Time" when server clock is not the same as client clock
The cause is that startTime is from the server clock but the current time is client (browser) clock. The two might not be the same.

stanleyxu2005:
The `uptime` is no longer calculated from the browser. If the application is terminated, the appdetail will not return any meaningful data.


clockfly:
+1

PR #1487, opened 2015-10-21
fix #1481, improve storm over gearpump perf

### LINKED ISSUE #1481: storm over gearpump doesn't perform well

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/gearpump/gearpump/1487)


clockfly:
+1

manuzhang:
Changes include

1. set `TupleImpl.messageId` to `null` to avoid map initialization
2. avoid converting java collections to scala
3. turn `foldLeft` into `tailrec

Now storm over gearpump reaches 5 million/s on 3-node(72 tasks) cluster

Issue #1401, opened 2015-09-09
Executor for dead processor is still alive?


stanleyxu2005:
While troubleshooting #1399, we found the the application (also in worker section) will return all executors, including those work for dead processors. Actually, dead processor stops processing any new messages, but the process is still alive. 

It needs to be distinguish the following things in the dashboard:
1. The processor page is still accessible by typing url? (considering yes, and provide links in a new history page)
2. The impacted executors should still appear in executor list table (in application overview)? (considering no. The backend should consider how to define the status of such an executor. Currently it is still `running`.)
3. The executors should still appear in worker page? (considering yes but with a grey indicator color. Currently the worker does have executor status information.)

clockfly:
@stanleyxu2005 

This is related with dynamic-dag. 
1. Sometimes user still want to query the in-memory state of processor that has been repalced. That is why we don't shutdown it down.
2. another reason is that it is tricky to handle the recovery when we want to recover to state of old DAG(before dag replacement)

Maybe We can make that configurable(whether to shut down old dag processor that has been replaced).

manuzhang:
The problem is when user sum up the `Executor.TaskCount` they find the number is not consistent with that on the overview tab. We could explicitly display counts for **both live and dead tasks** on the overview tab such that users won't be confused.

manuzhang:
I also find that executors that have been shutdown are still displayed on the application overview page. Is this expected ?

stanleyxu2005:
So we need to a filter to switch active/all executors

stanleyxu2005:
@manuzhang A simplified solution can be: When a processor is replaced, we add a "replaced" flag to this processors, so that we are able to distinguish such processor with normal processors.

manuzhang:
It's actually new processors and processor id is different and users don't know which processors have been replaced from dashboard.  I'm not sure whether it conveys right information to users. And what about executors ?

manuzhang:
I guess sometime we would allow users to query history dag versions from a pull-down menu and they get all the statistics (average throughput, average latency, etc) of that dag version. Well, that's just off top of my head. What I mean is that we not only solve the problem for now, we'd better have a blueprint to account for future usages.

stanleyxu2005:
I marked dead processors with a gray stripe at the left side of processor table in [this PR](https://github.com/gearpump/gearpump/pull/1765). Close the issue.

PR #1490, opened 2015-10-21
fix #1489, remove terminated executor from ExecutorManager

### LINKED ISSUE #1489: terminated executor is not removed from ExecutorManager

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/gearpump/gearpump/1490)

### DIFF
--- a/streaming/src/main/scala/io/gearpump/streaming/appmaster/ExecutorManager.scala
+++ b/streaming/src/main/scala/io/gearpump/streaming/appmaster/ExecutorManager.scala
@@ -21,15 +21,15 @@ package io.gearpump.streaming.appmaster
 import akka.actor._
 import akka.remote.RemoteScope
 import com.typesafe.config.Config
-import io.gearpump.streaming.{ExecutorId, ExecutorToAppMaster}
-import io.gearpump.streaming.executor.Executor
 import io.gearpump.cluster.AppMasterToWorker.ChangeExecutorResource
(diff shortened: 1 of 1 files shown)


clockfly:
+1

stanleyxu2005:
The patch will help to hide executors when related processor is terminated, right?
Can I reference it to https://github.com/gearpump/gearpump/issues/1401

clockfly:
There is slight difference.

Here actually means the executor is really dead. But in #1401, the executor still alive, but hidden because it will not handle any message in future due to its lifetime definition.

streaming/src/main/scala/io/gearpump/streaming/appmaster/ExecutorManager.scala
```diff
@@ -121,9 +121,12 @@ private[appmaster] class ExecutorManager(
   def terminationWatch : Receive = {
     case Terminated(actor) =>
       val executorId = Try(actor.path.name.toInt)
-      LOG.error(s"Executor $executorId is down")
       executorId match {
-        case scala.util.Success(id) => taskManager ! ExecutorStopped(id)
+        case scala.util.Success(id) => {
+          executors -= id
+          LOG.error(s"Executor $id is down")
```
stanleyxu2005: Not big thing, but I'd like to get clear. `LOG.error` means the executor is terminated **unexpectedly**. Expected behavior should be logged with `INFO`.

manuzhang: yes. it means the executor is terminated unexpectedly

PR #1493, opened 2015-10-22
fix #1485, fix send pending messages

### LINKED ISSUE #1485: NoSuchElementException when flushing pending messages

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/gearpump/gearpump/1493)

### DIFF
--- a/streaming/src/main/scala/io/gearpump/streaming/task/ExpressTransport.scala
+++ b/streaming/src/main/scala/io/gearpump/streaming/task/ExpressTransport.scala
@@ -94,7 +94,9 @@ class SendLater(express: Express, serializerPool: SerializerPool, sender: ActorR
       flushPendingMessages(localActor.get, transportId)
     } else {
       val remoteAddress = express.lookupRemoteAddress(transportId)
-      flushPendingMessages(remoteAddress.get, transportId)
+      if (remoteAddress.isDefined) {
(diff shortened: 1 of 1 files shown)


streaming/src/main/scala/io/gearpump/streaming/task/ExpressTransport.scala
```diff
@@ -94,7 +94,9 @@ class SendLater(express: Express, serializerPool: SerializerPool, sender: ActorR
       flushPendingMessages(localActor.get, transportId)
     } else {
       val remoteAddress = express.lookupRemoteAddress(transportId)
-      flushPendingMessages(remoteAddress.get, transportId)
+      if (remoteAddress.isDefined) {
```
clockfly: Why remoteAddress is empty?

manuzhang: it's cleared on recovery and `RestartTasks`

clockfly: Will this impact strong message ordering?

manuzhang: this is dropping messages which is bad. I'd better fix the empty remoteAddress issue.

PR #1495, opened 2015-10-24
fix #1485, fix NoSuchElementException when flushing pending messages

### LINKED ISSUE #1485: NoSuchElementException when flushing pending messages

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/gearpump/gearpump/1495)


manuzhang:
changes include,

1. remove clearing `remoteAddressMap` on `RestartTasks` which could be received later than `TaskLocations` and causes NoSuchElementException (detailed cause at #1485)
2. log errors if this exception happens for any reason (bugs) not yet found

PR #1500, opened 2015-10-29
fix #1485 no more pending messages in ExpressTransport

### LINKED ISSUE #1485: NoSuchElementException when flushing pending messages

huafengw:


streaming/src/main/scala/io/gearpump/streaming/task/TaskActor.scala
```diff
@@ -281,6 +259,7 @@ class TaskActor(
         queue.add(msg)
         handler()
       case None =>
+        LOG.error(s"Task $taskId drop message $msg")
```
clockfly: There is one problem, when there is a upstream zombie task keeping sending message, it will flood the log file. 

Maybe we should find some solution which meet:
1. indicate there is unexpected error happening.
2. avoid flooding the log file.

streaming/src/main/scala/io/gearpump/streaming/task/ExpressTransport.scala
```diff
@@ -67,63 +60,11 @@ trait ExpressTransport {
 
         val remoteAddress = express.lookupRemoteAddress(transportId)
         if (remoteAddress.isDefined) {
-          if(sendLater.hasPendingMessages){
-            sendLater.flushPendingMessages(remoteAddress.get, transportId)
-          }
           express.transport(taskMessage, remoteAddress.get)
         } else {
-          sendLater.addMessage(transportId, taskMessage)
+          LOG.error(s"Can not find target task $remote")
```
clockfly: Maybe you want to put some inline document about on what ocassion this error will happen?

streaming/src/main/scala/io/gearpump/streaming/executor/Executor.scala
```diff
@@ -179,11 +187,28 @@ class Executor(executorContext: ExecutorContext, userConf : UserConfig, launcher
...
+      tasks.get(rejected.taskId).foreach(_ ! PoisonPill)
+  }
+
+  def taskMsgHandler: Receive = {
+    case register: RegisterTask =>
+      sendMsgWithTimeOutCallBack(appMaster, register, registerTaskTimout, registerTaskTimeOut())
+  }
+
+  private def registerTaskTimeOut(): Unit = {
+    LOG.error(s"Failed to register task to AppMaster of application $appId, executor id is $executorId")
```
clockfly: How can we know which task trigger the timeout error?

streaming/src/main/scala/io/gearpump/streaming/executor/Executor.scala
```diff
@@ -179,11 +187,28 @@ class Executor(executorContext: ExecutorContext, userConf : UserConfig, launcher
 
     case query: QueryExecutorConfig =>
       sender ! ExecutorConfig(systemConfig)
+
+    case registered: TaskRegistered =>
+      LOG.debug(s"Register task ${registered.taskId} succeed")
+    case rejected: TaskRejected =>
+      LOG.info(s"Task ${rejected.taskId} is rejected by AppMaster, shutting down myself...")
+      tasks.get(rejected.taskId).foreach(_ ! PoisonPill)
```
clockfly: Should we clean out the reference in field "tasks"?

As executor has watched the task, if task is killed by PoisonPill, executor will restart it automatically.

streaming/src/main/scala/io/gearpump/streaming/executor/Executor.scala
clockfly: What else do we need to do to recover from failure besides logging?

streaming/src/main/scala/io/gearpump/streaming/executor/Executor.scala
```diff
@@ -179,11 +187,28 @@ class Executor(executorContext: ExecutorContext, userConf : UserConfig, launcher
 
     case query: QueryExecutorConfig =>
       sender ! ExecutorConfig(systemConfig)
+
+    case registered: TaskRegistered =>
+      LOG.debug(s"Register task ${registered.taskId} succeed")
```
clockfly: Is this log important? Do we need to make it a INFO level?

What if some TaskRegistered  message is lost ,will there be some problem?

streaming/src/main/scala/io/gearpump/streaming/executor/Executor.scala
```diff
@@ -147,11 +152,14 @@ class Executor(executorContext: ExecutorContext, userConf : UserConfig, launcher
...
-        express.remoteAddressMap.send(result)
-        express.remoteAddressMap.future().map{_ =>
-          tasks.foreach { task =>
-            task._2 ! TaskLocationReady
+      getStartClock.map { startClock =>
+        express.startClients(locations.keySet).map { _ =>
+          express.remoteAddressMap.send(result)
+          express.remoteAddressMap.future().map { _ =>
+            tasks.foreach { task =>
+              val sessionId = ids.newSessionId
```
clockfly: Is ids.newSessionId thread safe?

streaming/src/main/scala/io/gearpump/streaming/executor/Executor.scala
```diff
@@ -147,11 +152,14 @@ class Executor(executorContext: ExecutorContext, userConf : UserConfig, launcher
         val (host, taskIdList) = kv
         taskIdList.map(taskId => (TaskId.toLong(taskId), host))
       }
-      express.startClients(locations.keySet).map { _ =>
-        express.remoteAddressMap.send(result)
-        express.remoteAddressMap.future().map{_ =>
-          tasks.foreach { task =>
-            task._2 ! TaskLocationReady
+      getStartClock.map { startClock =>
```
clockfly: Should the start clock be passed along with TaskLocations?

streaming/src/main/scala/io/gearpump/streaming/executor/Executor.scala
```diff
@@ -57,10 +60,12 @@ class Executor(executorContext: ExecutorContext, userConf : UserConfig, launcher
 
   private val LOG: Logger = LogUtil.getLogger(getClass, executor = executorId, app = appId)
 
+  implicit val timeOut = FUTURE_TIMEOUT
   private val address = ActorUtil.getFullPath(context.system, self.path)
-  val systemConfig = context.system.settings.config
-
+  private val systemConfig = context.system.settings.config
+  private val ids = new SessionIdFactory(executorId)
```
clockfly: How to gurantee that the session id will not have duplicated value?

streaming/src/main/scala/io/gearpump/streaming/appmaster/TaskManager.scala
```diff
@@ -98,6 +85,7 @@ private[appmaster] class TaskManager(
         val resourceRequestDetail = subDAGManager.executorFailed(executorId)
         executorManager ! StartExecutors(resourceRequestDetail.requests, resourceRequestDetail.jar)
       } else {
+        appMaster ! ShutdownApplication(appId)
```
clockfly: instead of telling AppMaster to shutdown, maybe it is better to tell AppMaster what is going wrong. It is AppMaster's responsibility to decide how to react to that error.

streaming/src/main/scala/io/gearpump/streaming/appmaster/TaskManager.scala
```diff
@@ -215,9 +200,12 @@ private[appmaster] class TaskManager(
       }
 
     case Event(MessageLoss(executorId, cause), state) =>
-      if (state.taskRegistry.isTaskRegisteredForExecutor(executorId)) {
+      if (state.taskRegistry.isTaskRegisteredForExecutor(executorId) &&
+            executorRestartPolicy.allowRestartExecutor(executorId)) {
         goto(Recovery) using state.recoverState
       } else {
+        LOG.error(s"Task fails too many times to recover, shut down the application")
```
clockfly: Better need to document which task is restarting too many times...

streaming/src/main/scala/io/gearpump/streaming/executor/Executor.scala
```diff
@@ -179,11 +187,28 @@ class Executor(executorContext: ExecutorContext, userConf : UserConfig, launcher
...
+      LOG.info(s"Task ${rejected.taskId} is rejected by AppMaster, shutting down myself...")
+      tasks.get(rejected.taskId).foreach(_ ! PoisonPill)
+  }
+
+  def taskMsgHandler: Receive = {
+    case register: RegisterTask =>
+      sendMsgWithTimeOutCallBack(appMaster, register, registerTaskTimout, registerTaskTimeOut())
+  }
+
+  private def registerTaskTimeOut(): Unit = {
```
manuzhang: better name the function to timeOutBack to differentiate it from the val `registerTaskTimout`

streaming/src/main/scala/io/gearpump/streaming/executor/Executor.scala
```diff
@@ -57,10 +60,12 @@ class Executor(executorContext: ExecutorContext, userConf : UserConfig, launcher
...
   private val LOG: Logger = LogUtil.getLogger(getClass, executor = executorId, app = appId)
 
+  implicit val timeOut = FUTURE_TIMEOUT
   private val address = ActorUtil.getFullPath(context.system, self.path)
-  val systemConfig = context.system.settings.config
-
+  private val systemConfig = context.system.settings.config
+  private val ids = new SessionIdFactory(executorId)
   private val serializerPool = getSerializerPool()
+  private val registerTaskTimout = systemConfig.getLong(GEARPUMP_STREAMING_REGISTER_TASK_TIMEOUT_MS)
```
manuzhang: typo `Timout`

streaming/src/main/scala/io/gearpump/streaming/task/ExpressTransport.scala
manuzhang: what if an unexpected case causing the error to happen ? How about throwing exception here ?

streaming/src/main/scala/io/gearpump/streaming/appmaster/TaskManager.scala
```diff
@@ -260,6 +248,8 @@ private[appmaster] class TaskManager(
       }
   }
 
+  when(Recovery)(startTasks orElse ignoreClockEvent)
```
manuzhang: `TaskRegistry` is not cleared if we accept `RegisterTask` on recovery

clockfly:
Also need to consider the case of "at most once".

If user only need "at most once", then we should avoiding rebooting the whole app. 
For example, A -> B, A  -> C (A send data to B)
In  scenario of at most once, when B stops,  we don't want A to be rebooted, as C is still there.

streaming/src/main/scala/io/gearpump/streaming/task/ExpressTransport.scala
huafengw: I think this error will eventually cause the MessageLoss exception

clockfly:
Do you want to look at the UT failure?

streaming/src/main/scala/io/gearpump/streaming/appmaster/ExecutorManager.scala
```diff
@@ -70,6 +72,11 @@ private[appmaster] class ExecutorManager(
       context.become(service orElse terminationWatch)
   }
 
+  override val supervisorStrategy = OneForOneStrategy(maxNrOfRetries = 10, withinTimeRange = 1 minute) {
+    case ex: Throwable =>
+      Stop
```
clockfly: We need to LOG this

streaming/src/main/scala/io/gearpump/streaming/appmaster/TaskManager.scala
```diff
@@ -362,4 +361,8 @@ private [appmaster] object TaskManager {
       sessionId
     }
   }
+}
+
+object Test extends App {
```
clockfly: Remove test code?

streaming/src/main/scala/io/gearpump/streaming/executor/Executor.scala
```diff
@@ -179,11 +175,31 @@ class Executor(executorContext: ExecutorContext, userConf : UserConfig, launcher
 
     case query: QueryExecutorConfig =>
       sender ! ExecutorConfig(systemConfig)
+
+    case registered: TaskRegistered =>
+      taskToSessionId += registered.taskId -> registered.sessionId
+    case rejected: TaskRejected =>
+      tasks.get(rejected.taskId).foreach(_ ! PoisonPill)
```
clockfly: Do we want to combine taskToSessionId  with tasks, so that it is easier to manage consistency

streaming/src/main/scala/io/gearpump/streaming/appmaster/TaskManager.scala
huafengw: When recovery, I think it's using the recovery state, whose TaskRegistry is empty.

clockfly:
@huafeng,

please check
1. whether supervisorStrategy will use the main inbox.
2. check the executor state machine to make sure nothing goes wrong.

huafengw:
@clockfly updated

clockfly:
+1

PR #1520, opened 2015-11-09
fix #1519 fix performance regression

### LINKED ISSUE #1519: serious performance regression on Java 6
Also found regression on Java 7, while it's not significant

huafengw:


whjiang:
+1.

PR #1531, opened 2015-11-11
exclude auto-generated files from version control

stanleyxu2005:
also exclude build script and build instruction from generated site


whjiang:
+1.

PR #1532, opened 2015-11-12
fix: cannot switch to a tab that contains space

stanleyxu2005:
The patch will keep spaces in tab caption but remove spaces in anchor.


whjiang:
+1.

PR #1533, opened 2015-11-12
allows cluster to setup port forwarding for testing e.g. kafka 

stanleyxu2005:
(2) print exec result in log (3) renamed packages


clockfly:
+1

PR #1534, opened 2015-11-12
will cleanup docker environment before cluster startup

stanleyxu2005:
I tried to remove existing containers, if a container with same name exists already. But Docker seems not to remove such container randomly. With a **remove all container** command, the issue will not happen. So I will cleanup alll existing docker containers before cluster getting started.


clockfly:
+1

PR #1540, opened 2015-11-13
fix #1536, support storm ack

### LINKED ISSUE #1536: storm over gearpump drops at-least-once support

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/gearpump/gearpump/1540)

### DIFF
--- a/experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormProducer.scala
+++ b/experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormProducer.scala
@@ -18,11 +18,20 @@
 
 package io.gearpump.experiments.storm.producer
 
+import java.util.concurrent.TimeUnit
+
+import akka.actor.Actor.Receive
 import io.gearpump.Message
 import io.gearpump.cluster.UserConfig
 import io.gearpump.experiments.storm.topology.GearpumpStormComponent.GearpumpSpout
 import io.gearpump.experiments.storm.util._
-import io.gearpump.streaming.task.{StartTime, Task, TaskContext}
+import io.gearpump.streaming.task._
+
+import scala.concurrent.duration.Duration
+
+object StormProducer {
+  private[storm] val TIMEOUT = Message("timeout")
+}
 
 /**
  * this is runtime container for Storm spout
@@ -30,19 +39,51 @@ import io.gearpump.streaming.task.{StartTime, Task, TaskContext}
 private[storm] class StormProducer(gearpumpSpout: GearpumpSpout,
                                    taskContext: TaskContext, conf: UserConfig)
   extends Task(taskContext, conf) {
+  import io.gearpump.experiments.storm.producer.StormProducer._
 
   def this(taskContext: TaskContext, conf: UserConfig) = {
     this(StormUtil.getGearpumpStormComponent(taskContext, conf)(taskContext.system)
       .asInstanceOf[GearpumpSpout], taskContext, conf)
   }
 
+  private val timeoutOpt = gearpumpSpout.getMessageTimeout
+
   override def onStart(startTime: StartTime): Unit = {
     gearpumpSpout.start(startTime)
+    if (gearpumpSpout.ackEnabled) {
+      getCheckpointClock
+    }
+    timeoutOpt.foreach(scheduleTimeout)
     self ! Message("start")
   }
 
   override def onNext(msg: Message): Unit = {
-    gearpumpSpout.next(msg)
+    msg match {
+      case TIMEOUT =>
+        timeoutOpt.foreach { timeout =>
+          gearpumpSpout.timeout(timeout)
+          scheduleTimeout(timeout)
+        }
+      case _ =>
+        gearpumpSpout.next(msg)
+    }
     self ! Message("continue")
+  }
+
+  override def receiveUnManagedMessage: Receive = {
+    case CheckpointClock(optClock) =>
+      optClock.foreach { clock =>
+        gearpumpSpout.checkpoint(clock)
+      }
+      getCheckpointClock
+  }
+
+  def getCheckpointClock: Unit = {
+    taskContext.scheduleOnce(Duration(StormConstants.CHECKPOINT_INTERVAL_SECS,
+      TimeUnit.SECONDS))(taskContext.appMaster ! GetCheckpointClock)
+  }
+
+  private def scheduleTimeout(timeout: Int): Unit = {
+    taskContext.scheduleOnce(Duration(timeout, TimeUnit.SECONDS)){ self ! TIMEOUT }
   }
 }
--- a/experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormSpoutOutputCollector.scala
+++ b/experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormSpoutOutputCollector.scala
@@ -20,23 +20,68 @@ package io.gearpump.experiments.storm.producer
 
 import java.util.{List => JList}
 
-import backtype.storm.spout.ISpoutOutputCollector
+import backtype.storm.spout.{ISpout, ISpoutOutputCollector}
+import io.gearpump.TimeStamp
 import io.gearpump.experiments.storm.util.StormOutputCollector
+import io.gearpump.util.LogUtil
+import org.slf4j.Logger
+
+case class PendingMessage(id: Object, messageTime: TimeStamp, startTime: TimeStamp)
 
 /**
  * this is used by Storm Spout to emit messages
  */
-private[storm] class StormSpoutOutputCollector(collector: StormOutputCollector) extends ISpoutOutputCollector {
+private[storm] class StormSpoutOutputCollector(
+    collector: StormOutputCollector, spout: ISpout, ackEnabled: Boolean) extends ISpoutOutputCollector {
+  private var pendingMessage: Option[PendingMessage] = None
 
-  override def emit(streamId: String, values: JList[AnyRef], messageId: scala.Any): JList[Integer] = {
-    collector.emit(streamId, values)
+  override def emit(streamId: String, values: JList[AnyRef], messageId: Object): JList[Integer] = {
+    val outTasks = collector.emit(streamId, values)
+    setPendingOrAck(messageId)
+    outTasks
   }
 
   override def reportError(throwable: Throwable): Unit = {
     throw throwable
   }
 
-  override def emitDirect(taskId: Int, streamId: String, values: JList[AnyRef], messageId: scala.Any): Unit = {
+  override def emitDirect(taskId: Int, streamId: String, values: JList[AnyRef], messageId: Object): Unit = {
     collector.emitDirect(taskId, streamId, values)
+    setPendingOrAck(messageId)
+  }
+
+
+  def ackPendingMessage(checkpointClock: TimeStamp): Unit = {
+    pendingMessage.foreach { case PendingMessage(id, messageTime, _) =>
+      if (messageTime <= checkpointClock) {
+        spout.ack(id)
+        reset
+      }
+    }
+  }
+
+  def failPendingMessage(timeout: Int): Unit = {
+    pendingMessage.foreach { case PendingMessage(id, _, startTime) =>
+      if (System.currentTimeMillis() - startTime >= timeout) {
+        spout.fail(id)
+      }
+      reset
+    }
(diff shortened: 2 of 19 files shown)


experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormProducer.scala
```diff
@@ -45,4 +53,17 @@ private[storm] class StormProducer(gearpumpSpout: GearpumpSpout,
...
+
+  override def receiveUnManagedMessage: Receive = {
+    case CheckpointClock(optClock) =>
+      optClock.foreach { clock =>
+        gearpumpSpout.onCheckpointClock(clock)
+      }
+      getCheckpointClock
+  }
+
+  def getCheckpointClock: Unit = {
```
stanleyxu2005: The name is very confusing. It implies me that the method will return a checkpoint clock, but actuall it checks checkpoints internally.

manuzhang: it asks appmaster for checkpoint clock

manuzhang:
this approach is like a state application, making using of CheckpointClock at ClockService.

Bolt tasks periodically report an acked timestamp to ClockService.
Spout tasks periodically poll the global CheckpointClock and ack the pending messages in a batch way.

experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormSpoutOutputCollector.scala
```diff
@@ -20,15 +20,20 @@ package io.gearpump.experiments.storm.producer
...
+import backtype.storm.spout.{ISpout, ISpoutOutputCollector}
+import io.gearpump.TimeStamp
 import io.gearpump.experiments.storm.util.StormOutputCollector
 
 /**
  * this is used by Storm Spout to emit messages
  */
-private[storm] class StormSpoutOutputCollector(collector: StormOutputCollector) extends ISpoutOutputCollector {
+private[storm] class StormSpoutOutputCollector(collector: StormOutputCollector, spout: ISpout) extends ISpoutOutputCollector {
+  private var pendingMessages = Vector.empty[(TimeStamp, Object)]
```
clockfly: It will OO quickly.

clockfly: It also will OO quickly if you do nothing about ack or fail.
Check https://github.com/nathanmarz/storm-contrib/blob/master/storm-kafka/src/jvm/storm/kafka/PartitionManager.java#L38

There is a TreeSet

clockfly:
I don' think this approach will ever work. 

How about this, we keep one message with min-timestamp pending, and ack success all rest messages. 

For example, for mesage sequence....
1, 2, 3 ,4, 5, 6, 7, 8, 9, 10, 11, 

We keep message 1 pending, (neither success nor fail)
and ack message 2, 3,4, 5,6,7,8,9,10,11 as success.

when we confirm all subsequent mssages are all processed, we ack 1 as success, and choose a new pending message.

In this way, there is also 1 pending message for each spout, solving the memory issue.

clockfly:
When there is message loss, we will recover from the offset of the pending message, which is older than message 2,3 ,4,5,6,7,....

So, all subsequent messages are replayed.

clockfly:
This only works for persistent spout with two characteristics:
1. replayable.   which means we are able to replay messages if there is any loss.
2. sequential, which means we are able to replay all messages after the lost messages.

clockfly:
So, we should propoably warn user during the DAG submission stage for cases like:
1. if we are sure at-least-once cannot be supported.
2. If we are not sure at-least-once can be supported.

clockfly:
The ack and fail must be called with meanfully message id, even if we don't support exacly-once.

Otherwise, there is possble memory leak inside implementation of Spout.

manuzhang:
which one do you mean ? The acked MessageId is passed from Spout.

clockfly:
How you handle failure? How you notify spout there are failures?

I don't know how at least once is working

experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormSpoutOutputCollector.scala
```diff
@@ -36,7 +39,26 @@ private[storm] class StormSpoutOutputCollector(collector: StormOutputCollector)
...
 
-  override def emitDirect(taskId: Int, streamId: String, values: JList[AnyRef], messageId: scala.Any): Unit = {
+  override def emitDirect(taskId: Int, streamId: String, values: JList[AnyRef], messageId: Object): Unit = {
+    setPendingOrAck(messageId)
     collector.emitDirect(taskId, streamId, values)
   }
+
+  def ack(checkpointClock: TimeStamp): Unit = {
+    if (pendingMessageId.nonEmpty && pendingMessageId.head._1 <= checkpointClock) {
+      spout.ack(pendingMessageId)
```
clockfly: Can you confirm will there be problems to do this in same thread?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormSpoutOutputCollector.scala
```diff
@@ -36,7 +39,26 @@ private[storm] class StormSpoutOutputCollector(collector: StormOutputCollector)
...
+      pendingMessageId = List.empty[(TimeStamp, Object)]
```
clockfly: Since there is only one element, do we need to use the List?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormProducer.scala
```diff
@@ -38,6 +43,9 @@ private[storm] class StormProducer(gearpumpSpout: GearpumpSpout,
 
   override def onStart(startTime: StartTime): Unit = {
     gearpumpSpout.start(startTime)
+    if (gearpumpSpout.ackEnabled) {
```
clockfly: If ack is not enabled, we still need to ack the spout in some way. Otherwise, there will always some pending message in spout memory.

clockfly: If ack is not enabled, we need to ack every message at the same time when message is emited.
If ack is enabled, we can leave one message unacked.

experiments/storm/src/main/scala/io/gearpump/experiments/storm/topology/GearpumpStormComponent.scala
```diff
@@ -113,14 +121,20 @@ object GearpumpStormComponent {
...
     override def next(message: Message): Unit = {
-      collector.setTimestamp(System.currentTimeMillis())
       spout.nextTuple()
     }
+
+    def onCheckpointClock(clock: TimeStamp): Unit = {
+      collector.ack(clock)
+    }
+
+    private def isSequentiallyReplayableSpout(className: String): Boolean = {
```
clockfly: Is the kafka the only supported type?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormSpoutOutputCollector.scala
```diff
@@ -36,7 +39,26 @@ private[storm] class StormSpoutOutputCollector(collector: StormOutputCollector)
...
+    }
+  }
+
+  def setPendingOrAck(messageId: Object): Unit = {
+    val timestamp = System.currentTimeMillis()
+    if (pendingMessageId.isEmpty) {
+      pendingMessageId :+= timestamp -> messageId
+    } else {
+      spout.ack(messageId)
```
clockfly: Can you confirm whether there be problems to do this in same thread?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/topology/GearpumpStormComponent.scala
manuzhang: kafka is the only external source module in storm-0.9.x version. There could be more when we upgrade to storm-0.10.

experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormSpoutOutputCollector.scala
manuzhang: this is consistent with the storm implementation.

manuzhang:
The at-least-once works like this now,

With ack enabled by user, StormProducer keeps one pending message for Spout (sequentially replayable) along with its message time and start time and acks immediately for other messages (as you proposed). 

* On `Bolt.ack`, each StormProcessor periodically reports the acked timestamp (it's guaranteed there will be no more messages before this timestamp to be acked) to ClockService. 
* StormProducer periodically polls the global minimum checkpoint clock and acks Spout if the pendng message time is `<=` checkpoint clock. 
* if timeout is enabled, StormProducer will fail Spout if the pending message has expired 
* On `Bolt.fail` and other exceptions, Gearpump will restart applications and the Spout will replay from the last committed message.

manuzhang:
The limitations are 

1. if there is data skew (like in wordcount), a StormProcessor may not receive any messages and won't report its checkpoint clock. Then, the global checkpoint clock can't make progress.
2. we are not able to warn at topology submission stage whether at-least-once is supported since spout is in serialized form and its `Class` is not on the client class path.

experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormSpoutOutputCollector.scala
```diff
@@ -36,7 +45,43 @@ private[storm] class StormSpoutOutputCollector(collector: StormOutputCollector)
...
+  override def emitDirect(taskId: Int, streamId: String, values: JList[AnyRef], messageId: Object): Unit = {
+    setPendingOrAck(messageId)
     collector.emitDirect(taskId, streamId, values)
   }
+
+
+  def ackPendingMessage(checkpointClock: TimeStamp): Unit = {
+    pendingMessage.foreach { case PendingMessage(id, messageTime, _) =>
+      if (messageTime <= checkpointClock) {
+        spout.ack(id)
```
clockfly: Are you sure calling spout.ack in same thread of emit is consistent with native storm?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormSpoutOutputCollector.scala
```diff
@@ -36,7 +45,43 @@ private[storm] class StormSpoutOutputCollector(collector: StormOutputCollector)
...
+  }
+
+  private def setPendingOrAck(messageId: Object): Unit = {
+    val curTime = System.currentTimeMillis()
+    if (ackEnabled && pendingMessage.isEmpty) {
+      val startTime = curTime
+      val messageTime = curTime
+      pendingMessage = Some(PendingMessage(messageId, messageTime, startTime))
+    } else {
+      spout.ack(messageId)
```
clockfly: Are you sure calling spout.ack in same thread of emit is consistent with native storm?

manuzhang: yes. https://github.com/apache/storm/blob/0.9.x-branch/storm-core/src/clj/backtype/storm/daemon/executor.clj#L485. but I got the order wrong. Updated now.

clockfly:
+1

Make sure you  have done some perfomance test on this.

manuzhang:
thanks, I can rest in piece now.

clockfly:
Can you check the build failure?

manuzhang:
looks irrelevant to this pr

```
[info] HadoopCheckpointStoreIntegrationSpec:
[info] - HadoopCheckpointStore should persist and recover checkpoints *** FAILED ***
[info]   InvalidEscapeException was thrown during property evaluation.
[info]     Message: invalid escape character at index 12 in "checkpoints-\"
[info]     Occurred when passed generated values (
[info]       arg0 = 100 // 4 shrinks
[info]     )
```

manuzhang:
opened #1582 for the spec failure

manuzhang:
Merge this such that the integration test won't be blocked. I'll open follow up PRs if finding any bugs or performance issues.

PR #1552, opened 2015-11-16
added a docker based KafkaCluster for testing. 

stanleyxu2005:
Must query its ip address and then access the broker via `actual_kafka_cluster_ip:9092`


whjiang:
+1.

PR #1565, opened 2015-11-19
moved out direct Docker command usage (storm and kafka) out of test s…

stanleyxu2005:
…pec (2) always kill all running applications after a test case is executed


huafengw:
+1

PR #1570, opened 2015-11-19
add some rest client api, 2 stability test and a storm test

huafengw:


manuzhang:
+1

integrationtest/src/main/scala/io/gearpump/integrationtest/minicluster/RestClient.scala
```diff
@@ -64,14 +77,45 @@ class RestClient(host: String, port: Int) {
...
+    val streamAppMasterSummary = queryStreamingAppDetail(appId)
+    if(streamAppMasterSummary != null) {
+      streamAppMasterSummary.executors
+    } else {
+      List.empty
     }
-    */
+  }
+
+  def killAppMaster(appId: Int): Boolean = {
```
stanleyxu2005: The PR was +1 and merged too early. `killAppMaster` and `killExecutor` need actor system, this is not REST approach. If we cannot kill corresponding appmaster/executor at process level, then we'd reconsider move these test cases to unit tests.

PR #1571, opened 2015-11-19
fix several time out issue

whjiang:


stanleyxu2005:
I'm +1 after you rebase 4 commits as 1

core/src/main/resources/reference.conf
```diff
@@ -229,7 +229,7 @@ master {
     }
     cluster {
       roles = ["master"]
-      auto-down-unreachable-after = 15s
+      auto-down-unreachable-after = 90s
```
stanleyxu2005: Better not change options in core. For testing we need to override the option value, if it is really necessary.

whjiang:
@stanleyxu2005 I think we shall change in core. As we need to ensure Gearpump can also run on edge device.

PR #1574, opened 2015-11-20
replaced unnecessary Thread.sleep with retryUntil 

stanleyxu2005:
(2) removed failing test
(3) polished test cases


huafengw:
+1

PR #1593, opened 2015-11-24
support to restart cluster after running a destructive test case (2) …

stanleyxu2005:
…create different container in a more easier way (3) shell execution will timeout instead of waiting infinitively (4) simplied kafka cluster and tests (5) fixed unstable test cases in RestServiceSpec


whjiang:
+1.

PR #1596, opened 2015-11-25
bugfix: tests were fail because docker file is out of date (2) restar…

stanleyxu2005:
…t cluster at after stage not before stage


huafengw:
+1

PR #1599, opened 2015-11-25
add restart test for REST service

skw1992:


integrationtest/src/it/scala/io/gearpump/integrationtest/checklist/RestServiceSpec.scala
```diff
@@ -294,7 +294,27 @@ class RestServiceSpec extends TestSpecBase {
 
   "application life-cycle" should {
     "newly started application should be configured same as the previous one, after restart" in {
+      // setup
+      val originAppId = restClient.submitApp(wordCountJar)
+      val originAppsCount = restClient.listRunningApps().length
+      val originAppDetail = restClient.queryStreamingAppDetail(originAppId)
+      val originProcessors = originAppDetail.processors
```
stanleyxu2005: You should keep `originAppDetail` only. There is no need to store originProcessor, originSplitNum, etc.

integrationtest/src/it/scala/io/gearpump/integrationtest/checklist/RestServiceSpec.scala
```diff
@@ -294,7 +294,27 @@ class RestServiceSpec extends TestSpecBase {
...
+      val originSplitNum = originProcessors.get(0).get.parallelism
+      val originSumNum = originProcessors.get(1).get.parallelism
 
+      // exercise
+      Util.retryUntil(restClient.restartApp(originAppId))
+      val runningApps = restClient.listRunningApps()
+      runningApps.length shouldEqual originAppsCount
+      val newAppId = runningApps.head.appId
+      val newAppDetail = restClient.queryStreamingAppDetail(newAppId)
```
stanleyxu2005: You should check the kill application is not running

integrationtest/src/it/scala/io/gearpump/integrationtest/checklist/RestServiceSpec.scala
```diff
@@ -294,7 +294,27 @@ class RestServiceSpec extends TestSpecBase {
 
   "application life-cycle" should {
     "newly started application should be configured same as the previous one, after restart" in {
+      // setup
+      val originAppId = restClient.submitApp(wordCountJar)
```
stanleyxu2005: Better start the application with specified split and sum numbers, e.g. split=4, sum=3. So that we can see, the application is restarted with the same arguments.

integrationtest/src/main/scala/io/gearpump/integrationtest/minicluster/RestClient.scala
```diff
@@ -193,7 +193,8 @@ class RestClient(host: String, port: Int) {
 
   def restartApp(appId: Int): Boolean = try {
     val resp = callApi(s"appmaster/$appId/restart", CRUD_POST)
-    upickle.default.read[Status](resp).success
+    upickle.default.read[Status](resp)
+    true
```
stanleyxu2005: As the actor system returns `{success: false, reason: "timeout..."}`, please open an issue to confirm, whether this is a reasonable response.

And leave a comment with the issue number in code, so that code reader will understand, why we do not return `status.success` but true. And we might fix it later.

skw1992:
@stanleyxu2005 updated, please review

stanleyxu2005:
+1

PR #1601, opened 2015-11-25
add readme and fixed a potential docker command issue

stanleyxu2005:
Accidentally I saw wired things in log. `docker ps -q -a --filter 'name=master0'` could return three container ids. But if I manually execute the same command in console, it returns only one container id. After removing the single quote marks, it also returns one container id.

### DIFF
--- a/integrationtest/src/main/scala/io/gearpump/integrationtest/Util.scala
+++ b/integrationtest/src/main/scala/io/gearpump/integrationtest/Util.scala
@@ -39,25 +39,24 @@ object Util {
     }
   }
 
-  def retryUntil(condition: => Boolean, attempts: Int = 30,
-                 interval: Duration = 2.seconds): Unit = {
-    var success = false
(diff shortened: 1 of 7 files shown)


integrationtest/src/main/scala/io/gearpump/integrationtest/Util.scala
```diff
@@ -51,7 +51,7 @@ object Util {
         assert(success)
       } catch {
         case ex if attemptsLeft > 0 =>
-          LOG.info(s"condition is not met. will test again in ${interval.toSeconds}s ($attemptsLeft attempts left)")
+          LOG.info(s"condition is not met. will test again in ${interval.toSeconds}s ($attemptsLeft attempts left)\n" + ex.getMessage)
```
huafengw: I think it will output too much log

stanleyxu2005:
@huafengw You are right. I've removed the extra exception message.

PR #1602, opened 2015-11-25
fix #1514, add an experimental module to support akka-stream 1.0

### LINKED ISSUE #1514: Big data streaming with Akka Stream API 1.0
The goal: Able to use Akka-stream DSL to express big data streaming

TODO:
Add more description and design doc.

clockfly:


conf/gear.conf
```diff
@@ -64,7 +64,7 @@ gearpump {
     enabled = true
 
     # We will take one metric out of ${sample.rate}
-    sample-rate = 10
+    sample-rate = 1
```
whjiang: why needs to change sample-rate?

core/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,103 @@
+#
```
whjiang: why we need a  new log4j.properties file in core?

core/src/main/resources/reference.conf
```diff
@@ -33,13 +33,13 @@ gearpump {
...
     # We will take one metric out of ${sample.rate}
-    sample-rate = 10
+    sample-rate = 1
 
     report-interval-ms = 15000
 
     # reporter = "graphite"
     # reporter = "akka"
-    reporter = "logfile"
+    reporter = "akka"
```
whjiang: what's meaning of this change?

daemon/src/main/scala/io/gearpump/cluster/local/LocalCluster.scala
```diff
@@ -0,0 +1,87 @@
+/*
```
whjiang: why we need a new LocalCluster here? What's the difference between this local cluster and the local mode?

conf/gear.conf
clockfly: 1. With old sample-rate setting 10, when there a a few messages, < 10, the UI will display nothing. 
2. We heard complain before that the UI metrics is in-accurate on how many messages are received, as we will round it by 10.

3. Performance impact of this change still need investigation.

core/src/main/resources/log4j.properties
clockfly: Here is the motive.

1. There is a log4j.proporties under conf/. But conf is not in default classpath when we do test in IDE. When you try to debug in IDE, you will find there is no log output.
2. In runtime, log4j will take the first loaded log4j.properties file. It wil still take conf/log4j.properties as higher priority. So there is no impact in production deployment.

core/src/main/resources/reference.conf
clockfly: Our default reporter is set to "akka" in gear.conf. This is to make sure reference.conf is consistent with gear.conf.

Setting reporter as "akka" means we will gather all metrics by sending it to a actor receiver.

daemon/src/main/scala/io/gearpump/cluster/local/LocalCluster.scala
clockfly: The shell command Local.sh is a shortcut tool to create a cluster, the cluster will only used one machine. It is still a standalone cluster. The user code is not colocated with that JVM process.

The LocalCluster.scala is a util to create a in-process cluster, it will share the same JVM as user code.

core/src/main/resources/log4j.properties
clockfly: Maybe I can move it to test/ folder

whjiang:
+1

PR #1608, opened 2015-11-26
fix #1607 fix NONE_SESSION when restart tasks

### LINKED ISSUE #1607: When recovery, the restarted Task may receive NONE_SESSIONID

huafengw:


clockfly:
What cause this problem?
How this patch works?

huafengw:
The bug is caused by ```tasks ++= newRestarted.mapValues((_, NONE_SESSION))```
Suppose Task A is restarted and a session ID is assigned, then Task B is restarted, Task A's session id will be cleaned.

clockfly:
The executor is designed to handle transactional update.

When there is a dynamic dag update, executor will try to start new tasks, but its state "tasks" will only be updated atomically when all tasks has been started successfully.

If there is a failure in between, the executor wil rollback the change and use old “tasks" state. 
It is important to keep "tasks" updated in atomic way. Otherwise, the tasks is in a uncertain state when dynaimc dag update fails.

huafengw:
I see, I'll try to fix it in this way.

huafengw:
@clockfly updated

clockfly:
How about obsolete version?

huafengw:
Not sure when to remove the obsolete version, should we only just store current dag version's taskargument when receiving StartAllTasks?

clockfly:
+1

PR #1610, opened 2015-11-26
fix #1609, fix storm message timestamp bugs

### LINKED ISSUE #1609: storm over gearpump drops first message

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/gearpump/gearpump/1610)

### DIFF
--- a/experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormProducer.scala
+++ b/experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormProducer.scala
@@ -46,21 +46,21 @@ private[storm] class StormProducer(gearpumpSpout: GearpumpSpout,
       .asInstanceOf[GearpumpSpout], taskContext, conf)
   }
 
-  private val timeoutOpt = gearpumpSpout.getMessageTimeout
+  private val timeoutMillis = gearpumpSpout.getMessageTimeout
 
   override def onStart(startTime: StartTime): Unit = {
     gearpumpSpout.start(startTime)
     if (gearpumpSpout.ackEnabled) {
       getCheckpointClock
     }
-    timeoutOpt.foreach(scheduleTimeout)
+    timeoutMillis.foreach(scheduleTimeout)
     self ! Message("start")
   }
 
   override def onNext(msg: Message): Unit = {
     msg match {
       case TIMEOUT =>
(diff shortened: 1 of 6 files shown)


manuzhang:
1. The bug is caused by that timestamp is set after the message is emitted. Hence, the first message gets and timestamp -1 and is filtered out. 
2. This PR also fixes that seconds and milliseconds are mixed up in several places

manuzhang:
please hold the merge. I've got to fix the tests

clockfly:
+1 on the patch.

One question about the checkpoint clock, why bolt need to report checkpoint clock?

Cannot the spout just query the AppMaster to get the latest clock? So that it know whether to ack or fail?

experiments/storm/src/main/scala/io/gearpump/experiments/storm/producer/StormProducer.scala
```diff
@@ -79,11 +79,11 @@ private[storm] class StormProducer(gearpumpSpout: GearpumpSpout,
...
   def getCheckpointClock: Unit = {
-    taskContext.scheduleOnce(Duration(StormConstants.CHECKPOINT_INTERVAL_SECS,
-      TimeUnit.SECONDS))(taskContext.appMaster ! GetCheckpointClock)
+    taskContext.scheduleOnce(Duration(StormConstants.CHECKPOINT_INTERVAL_MILLIS,
+      TimeUnit.MILLISECONDS))(taskContext.appMaster ! GetCheckpointClock)
   }
 
-  private def scheduleTimeout(timeout: Int): Unit = {
-    taskContext.scheduleOnce(Duration(timeout, TimeUnit.SECONDS)){ self ! TIMEOUT }
+  private def scheduleTimeout(timeout: Long): Unit = {
```
stanleyxu2005: You might use `duration` to avoid potential unit confusion

manuzhang: Long is more convenient as long as I use it consistently and make good names

manuzhang:
@clockfly AppMaster doesn't know whether Bolts have successfully executed tuples unless received report clock from bolts on `bolt.ack`

manuzhang:
@clockfly merge this in. We can continue the discussion on new issues or offline.

PR #1613, opened 2015-11-26
added a python wordcount test case for multilang support

stanleyxu2005:


integrationtest/docker/gearpump-launcher/Dockerfile
```diff
@@ -2,7 +2,10 @@
 # See accompanying LICENSE file.
 
 # The base image contains JRE8
-FROM stanleyxu2005/minimal-jre8
+FROM errordeveloper/oracle-jre
```
manuzhang: have you verified whether this image has the correct opkg repo url ?

stanleyxu2005: The base image was updated 2 days before. I successfully build my new image. It contains python 2.7.10. And I have verified the test case with the image.

manuzhang:
+1

PR #1646, opened 2015-12-03
fix #1645: parse appid in submission response better

### LINKED ISSUE #1645: Parse appid in submission response
This is a follow up improvement of the revision fd4528daae0f049bba127bd9c9187669563a0fef 
When application is submitted via command line, the client need to parse console output and extract app id. We need to parse app id correctly.

stanleyxu2005:


huafengw:
+1

PR #1647, opened 2015-12-03
fix #1424: when application is terminated, related status indicators …

### LINKED ISSUE #1424: When application is stopped, the UI still show the applicaiton is running
How to reproduce:

1. submit an application
2. switch to app detail page. Check the "green" status icon.
3. Kill the app in command line with gear kill ..
4. Check the UI in step 2, there is still green icon. There is no prompt that the applicaiton has stopped.

stanleyxu2005:
…will become gray


huafengw:
+1

PR #1657, opened 2015-12-04
#1660 remove Try block in io.gearpump.experiments.yarn.ContainerLaunchConte…

Renkai:
Since cluster on YARN will not run successfully anyway if getAMLocalResourcesMap does not perform well.I think the Try{}match{} block should be removed so that we can get some useful error info.

### DIFF
--- a/experiments/yarn/src/test/scala/io/gearpump/experiments/yarn/ContainerLaunchContextSpec.scala
+++ b/experiments/yarn/src/test/scala/io/gearpump/experiments/yarn/ContainerLaunchContextSpec.scala
@@ -18,6 +18,8 @@
 
 package io.gearpump.experiments.yarn
 
+import java.nio.file.{Files, Paths}
+
 import org.apache.hadoop.yarn.conf.YarnConfiguration
 import org.mockito.Mockito._
 import org.scalatest.FlatSpecLike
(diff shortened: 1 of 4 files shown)


manuzhang:
@Renkai Thanks for your contribution. Please open an issue id for your PR. You may follow the [CommitGuideline](https://github.com/gearpump/gearpump/blob/master/CommitGuideline.md)

kkasravi:
@Renkai I agree with providing useful error information since this exception is 'swallowed'. However rather than just exposing a raw exception stack trace, the Try should be moved up to the scope of Client and YarnApplicationMaster - eg
```
object ContainerLaunchContext {
...
  def apply(yc: YarnConfiguration, ac: AppConfig)(command: String): Try[ContainerLaunchContext]
}
```

Renkai:
@kkasravi Is this what you mean?I give some possible reasons for specific exception,we can append more if we found.

manuzhang:
@Renkai 
1. please make sure the UT passes
2. please squash your commits into one.

kkasravi:
Thanks @Renkai yes something along these lines.

Renkai:
@manuzhang I make the unit test passed but I'm not sure it's a proper way to change test code like this.

experiments/yarn/src/test/scala/io/gearpump/experiments/yarn/ContainerLaunchContextSpec.scala
```diff
@@ -29,10 +29,18 @@ class ContainerLaunchContextSpec extends FlatSpecLike with MockitoSugar {
...
     val command = "foo"
     val hdfsRoot = "hdfs"
+    val mockVersion = "mockVersion"
     val yarnConfig = mock[YarnConfiguration]
     val appConfig = mock[AppConfig]
     when(appConfig.getEnv(Constants.HDFS_ROOT)).thenReturn(hdfsRoot)
+    when(appConfig.getEnv("version")).thenReturn(mockVersion)
+    when(yarnConfig.get("fs.defaultFS", "file:///")).thenReturn("file:///")
+    import sys.process._
+    s"mkdir $hdfsRoot".!
```
manuzhang: You can use the `File` API and wrap it in a try-finally block for resource collection.

Renkai:
Some tests failed in travis-ci,but if I run these tests separately,they all passed.

manuzhang:
+1. Thanks for your contribution. Please keep on the good work

PR #1656, opened 2015-12-04
fix #1653 fix replacing a processor

huafengw:
It's a quick fix for the bug, not recommend to merge into master. Raised for discussion.

PR #1661, opened 2015-12-06
fix #1655: added a test case for replacing a processor

### LINKED ISSUE #1655: Create a test case for replacing a processor
Submit a SOL application and then change the split processor with Wordcount's split processor (task class and jar file) and expect the change can take effects.

stanleyxu2005:
The PR can review huafeng's #1656


huafengw:
+1

PR #1663, opened 2015-12-07
fix #1536, fix storm at least once support

### LINKED ISSUE #1536: storm over gearpump drops at-least-once support

manuzhang:
the previous discussion is at #1540


manuzhang:
The previous solution is flawed and the new idea is illustrated with the graph below. 

storm_gearpump_kafka

The previous solution only keeps one `pending message`, acks the following messages (to avoid spout OOM) and acks the `pending message` when the global minAckTime exceeds its timestamp. This is incorrect since KafkaSpout will advance the store offset to the latest message when the `pending message` is acked (remember we already ack the messages behind the `pending message`) but we are not sure messages after `pending message` have been processed by the system. 

The new idea is to add a `next pending message` when the `pending message`'s timestamp is exceeded by the global minAckTime and only ack the `pending message` when the `next pending message`'s timestamp is exceeded. Then KafkaSpout will advance the stored offset to that of `next pending message`. This is safe since we know that messages before `next pending message` have been processed by the system.

Then, `next pending message` becomes the new `pending message` and the whole procedure loops.

clockfly:
+1

PR #1669, opened 2015-12-08
fix #1666: will should application details (and load metrics afterwards)

### LINKED ISSUE #1666: Need to show application details even when its metrics are unavailable
This is a sub task of issue #1499 

When application is too busy to be handled (e.g. wordcount with 1000 splits and 1000 sum), metrics requests can be timed out. In this case, the dashboard should still be able to show basic information of application.

stanleyxu2005:
also enabled gzip compression for REST service overall.


clockfly:
+1

PR #1674, opened 2015-12-08
fix #1672: when worker is killed, its detail page should become inactive

### LINKED ISSUE #1672: When worker is killed, UI still show it is active
Similar to #1424

stanleyxu2005:
Also set a global rest query timeout to 30 seconds.


clockfly:
+1

PR #1676, opened 2015-12-09
Fixes #1640 YARN deployment - no easy way of obtaining master address…

### LINKED ISSUE #1640: YARN deployment - no easy way of obtaining master address when deployed without services
GearPump master on YARN is normally available in GearPump UI (`services`).

After deploying GearPump on YARN with  `SERVICES_ENABLED` set to false there is no easy way to get master address -- the address required by `services`.

kkasravi:
… when deployed without services

### DIFF
--- a/experiments/yarn/src/main/scala/io/gearpump/experiments/yarn/client/Client.scala
+++ b/experiments/yarn/src/main/scala/io/gearpump/experiments/yarn/client/Client.scala
@@ -19,10 +19,21 @@ package io.gearpump.experiments.yarn.client
 
 import java.io.{File,FileNotFoundException}
 
+import java.util.concurrent.TimeUnit
+
+import akka.actor.ActorSystem
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.client.RequestBuilding
+import akka.http.scaladsl.model.{HttpRequest, HttpResponse, StatusCodes}
+import akka.http.scaladsl.unmarshalling.Unmarshal
+import akka.stream.ActorMaterializer
+import akka.stream.scaladsl.{Flow, Sink, Source}
 import com.typesafe.config.ConfigFactory
+import io.gearpump.cluster.AppMasterToMaster.MasterData
+import io.gearpump.cluster.main.{ArgumentsParser, CLIOption, ParseResult}
 import io.gearpump.experiments.yarn
 import io.gearpump.experiments.yarn.{AppConfig, ContainerLaunchContext}
-import io.gearpump.cluster.main.{ArgumentsParser, CLIOption, ParseResult}
+import io.gearpump.services._
 import io.gearpump.util.{Constants, LogUtil}
 import org.apache.hadoop.fs.{FileSystem, Path}
 import org.apache.hadoop.yarn.api.ApplicationConstants
@@ -34,6 +45,7 @@ import org.apache.hadoop.yarn.util.{Apps, Records}
 import org.slf4j.Logger
 
 import scala.collection.JavaConversions._
+import scala.concurrent.duration.Duration
 import scala.util.{Failure, Success, Try}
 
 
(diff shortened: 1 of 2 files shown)


experiments/yarn/src/main/scala/io/gearpump/experiments/yarn/client/Client.scala
```diff
@@ -234,7 +274,7 @@ class Client(configuration: AppConfig, yarnConf: YarnConfiguration, yarnClient:
         case true =>
           submit() match {
             case Success(appId) =>
-              waitForTerminalState(appId)
+              logMasters(waitForTerminalState(appId))
```
stanleyxu2005: Dont capsule two calls in one line.
```
val report = waitForTerminalState(appId)
logMasters(report)
```

experiments/yarn/src/main/scala/io/gearpump/experiments/yarn/client/Client.scala
```diff
@@ -196,6 +204,38 @@ class Client(configuration: AppConfig, yarnConf: YarnConfiguration, yarnClient:
...
+    import upickle.default.read
+    implicit val system = ActorSystem("httpclient")
+    implicit val materializer = ActorMaterializer()
+    import concurrent.ExecutionContext.Implicits.global
+    LOG.info(s"trackingURL=${report.getTrackingUrl}")
+    val trackingURL = new java.net.URL(report.getTrackingUrl)
+    LOG.info(s"host=${trackingURL.getHost} port=${trackingURL.getPort} uri=/proxy/${report.getApplicationId}/api/v1.0/master")
+    lazy val trackingURLConnectionFlow: Flow[HttpRequest, HttpResponse, Any] =
+      Http().outgoingConnection(trackingURL.getHost, trackingURL.getPort)
+    val response = Source.single(RequestBuilding.Get(s"/proxy/${report.getApplicationId}/api/v1.0/master")).via(trackingURLConnectionFlow).runWith(Sink.head)
```
stanleyxu2005: Maybe define a protocol string `/api/v1.0`?

experiments/yarn/src/main/scala/io/gearpump/experiments/yarn/client/Client.scala
```diff
@@ -196,6 +204,38 @@ class Client(configuration: AppConfig, yarnConf: YarnConfiguration, yarnClient:
...
+            result match {
+              case Success(data) =>
+                val masterData = read[MasterData](data)
+                LOG.info(s"leader=${masterData.masterDescription.leader._1}:${masterData.masterDescription.leader._2}")
+              case Failure(throwable) =>
+                LOG.error("Failed to fetch masters", throwable)
+            }
+          })
+        case value =>
+          LOG.info(s"Bad status code=${value.intValue}")
```
stanleyxu2005: Should a bad status code be logged with an error (instead of info)?

stanleyxu2005:
+1 (maybe consider my two tiny comments)

experiments/yarn/src/main/scala/io/gearpump/experiments/yarn/client/Client.scala
```diff
@@ -196,6 +204,38 @@ class Client(configuration: AppConfig, yarnConf: YarnConfiguration, yarnClient:
       }
       Thread.sleep(1000)
     }
+    appReport
+  }
+
+  def logMasters(report: ApplicationReport): Unit = {
+    import upickle.default.read
+    implicit val system = ActorSystem("httpclient")
```
huafengw: Should terminate the system?

kkasravi:
thanks @stanleyxu2005 @huafengw - all issues should be resolved including rebase from master

clockfly:
It doesn't look right.

The goal is to create a service outside of YARN, why starting another services UI inside YARN?

kkasravi:
@clockfly Obtaining the master(s) addresses via the YARN API isn't possible. So we get the master addresses from the services UI. We could also terminate the services UI after getting this information which I'll add in a different issue.

clockfly:
I know what you are trying to do, but I don't feel this way is right.

kkasravi:
@karol-brejna-i needs something for now for gearpump broker. Let's discuss some alternatives and replace this if possible. I agree this seems to be a hack due to the limitation of yarn client API. Seems like other frameworks that have UI's would have similar issues.

kkasravi:
@clockfly BTW, look at the scheduleOnce where I schedule to query the services UI after 5 seconds to get the master info. This seems quite arbitrary and timing sensitive. Does akka-streams provide a Time Scheduler that I could put between Source and Sink.head that would wait this about of time or something similar?

clockfly:
I will merge this if I cannot find a better solution in one day.

kkasravi:
Ok, thanks @clockfly. @karol-brejna-i should be available on master soon.

kkasravi:
@clockfly I haven't looked at akka-streams 2.0-M1 but did look at TimerTransformer for introducing a time delay between Source and Sink.

karol-brejna-i:
Thanks @kkasravi!

clockfly:
+1

PR #1677, opened 2015-12-09
fix #1659 add a rest api to terminate service process

### LINKED ISSUE #1659: Services process should provide a REST API to exit
This is required for the TAP Gearpump broker integration

huafengw:
@kkasravi 
The api is http://host:8090/terminate and the response will be ```The requested resource could not be found but may be available again in the future.``` Is it OK for you?


clockfly:
+1

kkasravi:
Thanks @huafengw - perfect.

stanleyxu2005:
Guys. This is an administrative operation. Please keep in mind that we should at least move this api to a protected path `/admin` (when authorization is enabled)

PR #1684, opened 2015-12-09
fix #1683, Allow metrics to be aggregated before sending to UI

### LINKED ISSUE #1683: Allow metrics to be aggregated before sending to UI

clockfly:


stanleyxu2005:
Lgtm +1

huafengw:
HistoryMetricsServiceSpec failed

clockfly:
updated

core/src/main/scala/io/gearpump/metrics/MetricsAggregator.scala
```diff
@@ -0,0 +1,10 @@
+package io.gearpump.metrics
```
manuzhang: license header

core/src/main/scala/io/gearpump/util/HistoryMetricsService.scala
```diff
@@ -80,11 +82,26 @@ class HistoryMetricsService(name: String, config: HistoryMetricsConfig) extends
...
+             aggregators += aggregatorClazz -> aggregatorTry.get
+           } else {
+             LOG.error("Cannot apply aggregator " + aggregatorClazz, aggregatorTry.failed.get)
+             aggregators += aggregatorClazz -> new SkipAllAggregator
+           }
+         }
+         aggregators(aggregatorClazz)
+       }
+
+      sender ! HistoryMetrics(inputPath, aggregator.aggregate(fetchMetricsHistory(inputPath, readLatest)))
```
manuzhang: what if aggregate fails ?

clockfly: If exception happen, let it throw.

manuzhang: Then why should Aggregator instantiation failure should be treated differently ?

clockfly: one is expected, the other is not expected.

clockfly: Typically, I will just say "fail fast", most of the codebase is adopting this approach.
For most cases, the best way to handle the failure is to throw.

Unless for rare cases, like this:
aggregatorClazz is a parameter of a URL query, so we need to be defendent in this case.

manuzhang: yes. What I'm concerned is whether we are responsive to users. Does user know what fails and why ?

clockfly: If aggregatorClazz is wrong, what we do now is to log it with error log level. Admin can view the logs to find the errors. 

It is still "responsive", as we will use the SkipAllAggregator to return an empty set.

manuzhang:
+1

PR #1687, opened 2015-12-09
fix #1671, make default timeout shorter to be more UI responsive.

### LINKED ISSUE #1671: Timeout to retrieve worker list after killing a worker shortly
Steps to reprodure: 
1. Create 5 workers
2. Query workers `http://127.0.0.1:8090/api/v1.0/master/workerlist`. It will return 5 workers.
3. Kill 1 worker in console
4. Query workers again. It might timeout.
5. After seeing the timeout, query workers again, it will return 4 workers.

*Update 1*: My suggestion is to use a central repository to store cluster state.

clockfly:


huafengw:
+1

PR #1692, opened 2015-12-10
fix #1691: dashboard layout and textual changes and update dashing

### LINKED ISSUE #1691: Propose some dashboard layout and textual changes
1. Change "Receive Throughput Skew" to "Task Receive throughput", or some better
2. Still show the above graph if only one task is present.
3. Change "Task Selector" to "Pick one task to see its detail metrics..."or some better
4. Summary page, use (s): 2 processor -> processors , executor -> executors
5. Add a document link at the right corner. Along with the source code link.
6. More friendly JVM info
7. Update dashing library to fix table left stripe missing issue on Firefox and MSIE
8. Add JVM information on worker/application list page
9. Change question icon to information icon on overview header
10. Open configuration in new window
11. Change programmer preferred camel case application name to first letter capitalized representation

stanleyxu2005:
Also changed a REST call policy. Previously when querying worker status is timed out, we assume the worker is dead. No more query requests will be sent to backend. But if the backend is temporary stuck,  it might come back again shortly, so we should keep querying the backend.

### DIFF
--- a/services/dashboard/views/apps/streamingapp/executor.js
+++ b/services/dashboard/views/apps/streamingapp/executor.js
@@ -27,13 +27,15 @@ angular.module('dashboard')
     function($scope, $stateParams, $ptb, helper, restapi, conf, models, executor0) {
       'use strict';
 
+      $scope.executorName = executor0.id === -1 ?
+        'AppMaster' : ('Executor ' + executor0.id);
       $scope.overviewTable = [
-        $ptb.text('ID').done(),
+        $ptb.text('JVM Info').help('Format: PID@hostname').done(),
         $ptb.text('Actor Path').done(),
         $ptb.link('Worker').done(),
         $ptb.number('Task Count').done(),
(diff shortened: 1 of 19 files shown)


services/dashboard/views/apps/streamingapp/executor.js
```diff
@@ -27,13 +27,15 @@ angular.module('dashboard')
...
       'use strict';
 
+      $scope.executorName = executor0.id === -1 ?
+        'AppMaster' : ('Executor ' + executor0.id);
       $scope.overviewTable = [
-        $ptb.text('ID').done(),
+        $ptb.text('JVM Info').help('Format: PID@hostname').done(),
         $ptb.text('Actor Path').done(),
-        $ptb.link('Worker').done(),
+        $ptb.link('Working On').done(),
```
clockfly: Working On -> Worker

clockfly:
+1

PR #1696, opened 2015-12-10
fix #1682: allow specify a transition time when changing parallelism

### LINKED ISSUE #1682: Allow user to change the transition time when changing the parallelism

stanleyxu2005:
also make all form-control dialogs consistent


clockfly:
+1

PR #1699, opened 2015-12-11
fix #1698, refactor on gearpump-yarn experiment module

### LINKED ISSUE #1698: Refactor experiment-yarn module

clockfly:
still need to add some UT.

Some changes:
1. Refactor the yarn to make it simple
   a. add a glue layer to seperate yarn code from gearpump code
   b. Simplify the callbacks, RMClient...
2.  Will do dynamic allocation of ports to avoid port conflict errors.
3. Add a option to get active configration from yarn gearpump cluster, so that we can launch UI server on the gateway. 
4. Consolidate "gear.conf" with "yarn.conf"

It also fixed several issues of old impl:
1. In old code, different user will use the same hdfs path to store configuration, they will overwrite each other.
2. In old code, when user launch a DAG, the changes to gear.conf is not effective, as it is disregarded.
3. In old code, log4j config is not propograted to the runtime.

Follow ups:
- [x] . Should add more operation functions, cases like: 1. dynamic provision/remove a machine. 
- [ ] dyanamic add/remove a master
- [x] Should provide more command line tools to query the master status.
- [x] Currently, the XMX is not set for containers, There is risk that YARN may kill our daemon process directly. We need to add those XMX back, and make sure it is smaller than the container memory(as permgen still take some memory)


kkasravi:
@clockfly Thanks for refactoring. It looks like you're adding some good improvements.
1. Should we move yarn under gearpump/external? It is no longer an experiment and should be treated in the same way as kafka, hbase, etc. Given the scope of the changes, it may be easiest to do this now (mostly package renaming).
2. Should we have the yarn information in conf/gear.conf? It seems inconsistent with how we've eliminated the use of conf files in general (kafka, hbase). 
3. Should probably update docs/deployment-yarn.md to match what's in the experiments/yarn/README.md
4. Looks like no UT (I think you mention this).
5. Running this in oregon environment, -launch doesn't seem to override the gearpump.yarn.client.package-path property
```
$ output/target/pack/bin/yarnclient -launch /user/gearpump/gearpump-pack-2.11.5-0.7.1-SNAPSHOT.tar.gz -storeConfig /tmp/application.conf
File does not exist: /user/gearpump/gearpump.zip
	at org.apache.hadoop.hdfs.server.namenode.INodeFile.valueOf(INodeFile.java:66)
	at org.apache.hadoop.hdfs.server.namenode.INodeFile.valueOf(INodeFile.java:56)
	at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.getBlockLocationsUpdateTimes(FSNamesystem.java:1929)
	at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.getBlockLocationsInt(FSNamesystem.java:1870)
	at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.getBlockLocations(FSNamesystem.java:1850)
	at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.getBlockLocations(FSNamesystem.java:1822)
	at org.apache.hadoop.hdfs.server.namenode.NameNodeRpcServer.getBlockLocations(NameNodeRpcServer.java:558)
	at org.apache.hadoop.hdfs.server.namenode.AuthorizationProviderProxyClientProtocol.getBlockLocations(AuthorizationProviderProxyClientProtocol.java:87)
	at org.apache.hadoop.hdfs.protocolPB.ClientNamenodeProtocolServerSideTranslatorPB.getBlockLocations(ClientNamenodeProtocolServerSideTranslatorPB.java:363)
	at org.apache.hadoop.hdfs.protocol.proto.ClientNamenodeProtocolProtos$ClientNamenodeProtocol$2.callBlockingMethod(ClientNamenodeProtocolProtos.java)
	at org.apache.hadoop.ipc.ProtobufRpcEngine$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine.java:619)
	at org.apache.hadoop.ipc.RPC$Server.call(RPC.java:1060)
	at org.apache.hadoop.ipc.Server$Handler$1.run(Server.java:2044)
	at org.apache.hadoop.ipc.Server$Handler$1.run(Server.java:2040)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:422)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1671)
	at org.apache.hadoop.ipc.Server$Handler.run(Server.java:2038)

try to check if your gearpump version is right or HDFS was correctly configured.
```
I also tried replacing gearpump.zip with what existed in HDFS `      package-path = "/user/gearpump/gearpump-pack-2.11.5-0.7.1-SNAPSHOT.tar.gz"` and running `output/target/pack/bin/yarnclient -storeConfig /tmp/application.conf` and got a similar error.
```
$ output/target/pack/bin/yarnclient -storeConfig /tmp/application.conf
File does not exist: /user/gearpump/gearpump.zip
	at org.apache.hadoop.hdfs.server.namenode.INodeFile.valueOf(INodeFile.java:66)
	at org.apache.hadoop.hdfs.server.namenode.INodeFile.valueOf(INodeFile.java:56)
	at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.getBlockLocationsUpdateTimes(FSNamesystem.java:1929)
	at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.getBlockLocationsInt(FSNamesystem.java:1870)
	at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.getBlockLocations(FSNamesystem.java:1850)
	at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.getBlockLocations(FSNamesystem.java:1822)
	at org.apache.hadoop.hdfs.server.namenode.NameNodeRpcServer.getBlockLocations(NameNodeRpcServer.java:558)
	at org.apache.hadoop.hdfs.server.namenode.AuthorizationProviderProxyClientProtocol.getBlockLocations(AuthorizationProviderProxyClientProtocol.java:87)
	at org.apache.hadoop.hdfs.protocolPB.ClientNamenodeProtocolServerSideTranslatorPB.getBlockLocations(ClientNamenodeProtocolServerSideTranslatorPB.java:363)
	at org.apache.hadoop.hdfs.protocol.proto.ClientNamenodeProtocolProtos$ClientNamenodeProtocol$2.callBlockingMethod(ClientNamenodeProtocolProtos.java)
	at org.apache.hadoop.ipc.ProtobufRpcEngine$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine.java:619)
	at org.apache.hadoop.ipc.RPC$Server.call(RPC.java:1060)
	at org.apache.hadoop.ipc.Server$Handler$1.run(Server.java:2044)
	at org.apache.hadoop.ipc.Server$Handler$1.run(Server.java:2040)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:422)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1671)
	at org.apache.hadoop.ipc.Server$Handler.run(Server.java:2038)

try to check if your gearpump version is right or HDFS was correctly configured.
```
I also tried this:
```
$ JAVA_OPTS='-Dgearpump.yarn.client.package-path=/user/gearpump/gearpump-pack-2.11.5-0.7.1-SNAPSHOT.tar.gz' output/target/pack/bin/yarnclient -storeConfig /tmp/application.conf
Help: 
io.gearpump.experiments.yarn.client.Client -launch <Please specify the gearpump.zip package path on HDFS. If not specified, we will use default value /user/gearpump/gearpump.zip> -name <Application name showed in YARN> -verbose <print verbose log on console> -storeConfig <local path where the master config should be stored> 
-launch  (required:false, default:)
-name  (required:false, default:Gearpump)
-verbose  (required:false, default:false)
-storeConfig  (required:false, default:)
```
I haven't looked into why it's failing yet or looked at the code specifically. I'll provide some additional feedback if I can.

experiments/yarn/src/main/scala/io/gearpump/experiments/yarn/client/Client.scala
```diff
@@ -17,323 +17,171 @@
...
-      s" $mainClass" +
-      s" -version $version" +
-      " 1>" + logdir +"/" + ApplicationConstants.STDOUT +
-      " 2>" + logdir +"/" + ApplicationConstants.STDERR
-
-    LOG.info(s"command=$command")
-    command
-  }
+    // first step, check the version, to make sure local version matches remote version
+    val rootEntry = rootEntryPath(zip = gearPackagePath)
```
kkasravi: This used to work with package files of *.tar.gz. We should be able to handle either *.zip or *.tar.gz.

clockfly: Yes, zip and tar.gz both work.

clockfly:
There was some typo in command line argment. Please try again.

experiments/yarn/src/main/scala/io/gearpump/experiments/yarn/client/Client.scala
kkasravi: If we only want to do *.zip, we should check the suffix and provide better usage error

clockfly: tar.gz also work, it is just a varible name.

clockfly: Sorry, looks like only works for zip.

experiments/yarn/src/main/scala/io/gearpump/experiments/yarn/client/Client.scala
```diff
@@ -17,323 +17,171 @@
...
-  Try(apply()).recover{
-    case ex:FileNotFoundException =>
-      Console.err.println(s"${ex.getMessage}\n" +
-        s"try to check if your gearpump version is right " +
-        s"or HDFS was correctly configured.")
-    case ex => help; throw ex
+  private def storeConfig(url: String, targetFile: File): Unit = {
+    val client = new HttpClient()
+    val get = new GetMethod(url)
+    val status = client.executeMethod(get)
```
kkasravi: `status` is not used and should check for 200.

kkasravi: Currently it looks like the YARN applicationmaster is launched and runs but no /tmp/application.conf is stored when using the -storeConfig option. I believe this is because the services process is not running yet and returns a bad status code. Below I show the appid is returned by the yarnclient and the subdirectory is created with gear.conf, log4j.properties and yarn-site.xml.
```
$ output/target/pack/bin/yarnclient -launch /user/gearpump/gearpump-pack-2.11.5-0.7.1-SNAPSHOT.zip -storeConfig /tmp/application.conf
......================================================
===Application Id: application_1449749298938_5733
Help: 
io.gearpump.experiments.yarn.client.Client -launch <Please specify the gearpump.zip package path on HDFS. If not specified, we will use default value /user/gearpump/gearpump.zip> -name <Application name showed in YARN> -verbose <print verbose log on console> -storeConfig <local path where the master config should be stored> 
-launch  (required:false, default:)
-name  (required:false, default:Gearpump)
-verbose  (required:false, default:false)
-storeConfig  (required:false, default:)
```
```
$ yarn application -list
15/12/13 14:46:56 INFO client.RMProxy: Connecting to ResourceManager at cdh-master-0.node.demogotapaas.consul/10.10.10.146:8032
Total number of applications (application-types: [] and states: [SUBMITTED, ACCEPTED, RUNNING]):2
                Application-Id	    Application-Name	    Application-Type	      User	     Queue	             State	       Final-State	       Progress	                       Tracking-URL
application_1449749298938_5733	            Gearpump	                YARN	  ec2-user	root.ec2-user	           RUNNING	         UNDEFINED	            50%	http://cdh-worker-4.node.demogotapaas.consul:56663
```
```
$ hdfs dfs -ls /user/ec2-user
Found 2 items
drwx------   - ec2-user supergroup          0 2015-12-13 09:28 /user/ec2-user/.Trash
drwx------   - ec2-user supergroup          0 2015-12-13 14:46 /user/ec2-user/.gearpump_app5733
[ec2-user@cdh-master-0 gearpump]$ hdfs dfs -ls /user/ec2-user/.gearpump_app5733
Found 1 items
drwx------   - ec2-user supergroup          0 2015-12-13 14:46 /user/ec2-user/.gearpump_app5733/conf
[ec2-user@cdh-master-0 gearpump]$ hdfs dfs -ls /user/ec2-user/.gearpump_app5733/conf
Found 3 items
-rw-r--r--   3 ec2-user supergroup     166935 2015-12-13 14:46 /user/ec2-user/.gearpump_app5733/conf/gear.conf
-rw-r--r--   3 ec2-user supergroup       2062 2015-12-13 14:46 /user/ec2-user/.gearpump_app5733/conf/log4j.properties
-rw-r--r--   3 ec2-user supergroup      94058 2015-12-13 14:46 /user/ec2-user/.gearpump_app5733/conf/yarn-site.xml
```
```
$ ls /tmp/application.conf
ls: cannot access /tmp/application.conf: No such file or directory
```

experiments/yarn/src/main/scala/io/gearpump/experiments/yarn/client/Client.scala
```diff
@@ -17,323 +17,171 @@
...
-      appContext.setQueue(getEnv(YARNAPPMASTER_QUEUE))
+    writer.write(filterJvmReservedKeys.root().render())
+    writer.close()
+    out.close()
 
-      yarnClient.submitApplication(appContext)
-      appContext.getApplicationId
-    })
-  }
+    // save yarn-site.xml to remote
```
kkasravi: @huafengw @clockfly @TomaszT  I think this means that when launching the containers (masters, workers) this client YARN file is used rather than what exists on the target container node? Is the assumption that the Gearpump yarnclient will always be launched in the same YARN cluster (cdh version, etc) as where the master and workers will run? @karol-brejna-i is this a safe assumption? Will the Gearpump broker always launch the yarnclient in this environment? In any case it would be reasonable to have "yarn-site.xml" defined as a constant in io.gearpump.experiments.yarn.Constants even though it is not directly referenced by master or workers when they are launched (but i think the conf directory is). In the Constants.scala file we could also note where yarn-site.xml is loaded from and where it is used.

karol-brejna-i: @kkasravi, GearPump broker runs yarnclient on Cloud Foundry. I assumed yarn-site.xml to be present on machine/container where yarnclient is. I planned to supply yarn-site.xml with the broker app.

conf/gear.conf
```diff
@@ -181,4 +181,40 @@ gearpump {
...
+    client {
+      package-path = "/user/gearpump/gearpump.zip"
+    }
+
+    applicationmaster {
+      ## Memory of YarnAppMaster
+      command = "$JAVA_HOME/bin/java -Xmx512m"
+      memory = "512"
+      vcores = "1"
+      queue = "default"
```
huafengw: What is this ```queue``` configuration means? Need to remove it?

clockfly: No, it means the YARN application queue.

experiments/yarn/src/main/scala/io/gearpump/experiments/yarn/client/LaunchCluster.scala
```diff
@@ -0,0 +1,205 @@
...
+
+    if (!rootEntry.contains(version)) {
+      throw new IOException(s"Check version failed! Local gearpump binary version $version doesn't match with remote path $packagePath")
+    }
+
+    val resource = Resource.newInstance(memory, vcore)
+    val appId = yarnClient.createApplication
+
+    // will upload the configs to HDFS home directory of current user.
+    LOG.info("Uploading configuration files to remote HDFS(under /user/<id>/.gearpumpapp_)...")
```
huafengw: Not consistent with ```val confDir = s"${fs.getHomeDirectory}/.gearpump_app${appId.getId}/conf/"```

experiments/yarn/src/main/scala/io/gearpump/experiments/yarn/client/LaunchCluster.scala
```diff
@@ -0,0 +1,205 @@
...
+    var out = fs.create(remoteConfFile)
+    var writer = new OutputStreamWriter(out)
+
+    val cleanedConfig = ClusterConfig.filterOutDefaultConfig(akka)
+
+    writer.write(cleanedConfig.root().render())
+    writer.close()
+
+    // save yarn-site.xml to remote
+    val yarn_site_xml = s"$confDir/yarn-site.xml"
```
huafengw: You will also upload yarn-site.xml under conf, not mentioned in the doc.
Also, command ```yarnClient``` contains ```/etc/hadoop/conf``` in it's classpath, if user's yarn conf is not in that folder, will it failed to locate the right Hdfs address?

project/Pack.scala
```diff
@@ -6,8 +6,7 @@ import xerial.sbt.Pack._
 object Pack extends sbt.Build {
   val daemonClassPath = Seq(
     "${PROG_HOME}/conf",
-    "${PROG_HOME}/lib/daemon/*",
-    "${PROG_HOME}/lib/yarn/*"
+    "${PROG_HOME}/lib/daemon/*"
```
huafengw: DFSJarStoreService need yarn library

core/src/main/scala/io/gearpump/util/Constants.scala
```diff
@@ -119,9 +120,11 @@ object Constants {
   //Service related
   val GEARPUMP_SERVICE_HTTP = "gearpump.services.http"
   val GEARPUMP_SERVICE_HOST = "gearpump.services.host"
+  val GEARPUMP_SERVICE_SUPERVISOR_PATH = "gearpump.services.supervisor-actor-path"
 
   //The folders are under ${GEARPUMP_HOME}
-  val EXECUTOR_CLASSPATH_WHILTELIST = Array("conf", "lib")
+
+  val daemonPath = Array("lib" + File.separator + "daemon")
```
huafengw: Upper case?

huafengw: Executor's classpath should include services?

clockfly: service classpath is not included in Executor. 
Executor will inherit  classpath from worker
```
  val daemonClassPath = Seq(
    "${PROG_HOME}/conf",
    "${PROG_HOME}/lib/daemon/*"
  )
```

project/Pack.scala
clockfly: It will be added as environment classpath when you launch gearpump over yarn.

experiments/yarn/src/main/scala/io/gearpump/experiments/yarn/client/LaunchCluster.scala
clockfly: doc updated.

project/Pack.scala
huafengw: I mean the standalone mode, not yarn mode.

clockfly: updated.

clockfly:
Updated, please check again.

experiments/yarn/src/main/scala/io/gearpump/experiments/yarn/appmaster/YarnAppMaster.scala
```diff
@@ -0,0 +1,357 @@
...
+      i => Resource.newInstance(masterMemory, masterVCores)
+    ).toList
+    rmClient.requestContainers(containers)
+  }
+
+  private def launchMaster(container: Container): HostPort = {
+    LOG.info(s"Launch Master on container " + container.getNodeHttpAddress)
+    val host = container.getNodeId.getHost
+
+    val port = Util.findFreePort.get
```
huafengw: the free port is on the YarnAppMaster's host, not on the allocated container's, right?

clockfly: You are right! the return container port actually points to the node manager.

huafengw:
+1

PR #1702, opened 2015-12-14
fix #1701: wrong dirty check of modify processor

### LINKED ISSUE #1701: modify processor is not enabled at dashboard

stanleyxu2005:
also includes:
* polished modal dialog by adding back a horizontal split line after title
* updated dashing library (datepicker, style fixes, etc.) and other major libraries
* when application has stalling tasks, click on the yellow warning icon around application clock will switch to dag tab to check issue


huafengw:
+1

PR #1704, opened 2015-12-14
fix #1703: task metrics in bar chart were empty

### LINKED ISSUE #1703: Refresh processor detail page will lose metrics values in bar chart
All values will become 0 after a page refresh. And click the tab "Processor" will navigate to default page -> cluster/master.

stanleyxu2005:


huafengw:
+1

PR #1708, opened 2015-12-15
fix #1679 dynamic dag will fall back when replacing a processor failed

### LINKED ISSUE #1679: if replacing a processor failed, the application should rollback to last dag

huafengw:


clockfly:
+1

PR #1712, opened 2015-12-15
fix #1711: layout is more responsive on small devices

### LINKED ISSUE #1711: Propose dashboard to be more responsive on small devices

stanleyxu2005:
1. alternatively show menu items as dropdown menu for small screen size
2. selective hide unimportant table columns for small screen size
3. more reasonable column size for small screen size
4. blue loading bar (at top) will shown when a request takes longer than 1s (previously was 200ms)
5. fixed a sorting table flicker issue


clockfly:
+1

PR #1717, opened 2015-12-16
fix #1600 restart application return false

### LINKED ISSUE #1600: The restart application REST api returns false
The restart application api returns `{"success":false,"reason":"Ask timed out on [Actor..."}`, but the application can be restarted.

huafengw:


stanleyxu2005:
+1

PR #1748, opened 2015-12-21
fix #1747: updated the way to parse application id in command-line (2…

### LINKED ISSUE #1747: Integration Test Fail due to submit application response change
When submitting an application via command line, now the output has been changed. The previous test code could not parse application id as expected.

stanleyxu2005:
…) included DynamicDagSpec


huafengw:
+1

PR #1705, opened 2015-12-14
fix #1318, fix MinClock not updated fast enough for slow stream

manuzhang:
[Review on Reviewable](https://reviewable.io/reviews/gearpump/gearpump/1705)

PR #1749, opened 2015-12-21
fix #1318, fix MinClock not updated fast enough for slow stream

### LINKED ISSUE #1318: MinClock is not updated fast enough for slow stream
For slow stream, like 1 message in 1 second, the MinClock is not udpated fast enough.

manuzhang:
previously on #1705


streaming/src/main/scala/io/gearpump/streaming/task/Subscription.scala
```diff
@@ -182,7 +184,28 @@ class Subscription(
   }
 
   def minClock: TimeStamp = {
-    minClockValue.min
+    if (minClockValue.isEmpty) {
```
clockfly: It doesn't look right to do this in minClock
as minClock is supposed to be function with no side effect.

PR #1750, opened 2015-12-21
fix #1747: storm spec failure due to parse application id

### LINKED ISSUE #1747: Integration Test Fail due to submit application response change

stanleyxu2005:


huafengw:
+1

PR #1764, opened 2015-12-23
fix #1763, change all GearPump to Gearpump

### LINKED ISSUE #1763: "Gearpump" or "GearPump"
We use both of the terms now. "Gearpump" has around 300 occurrences and "GearPump" 40 occurrences. I think we should reach a consensus on which to use and use it from now on.

manuzhang:


whjiang:
+1.

PR #1771, opened 2015-12-24
fix #1770, processor metrics is aggregated in wrong way

### LINKED ISSUE #1770: processor metrics is aggregated wrong
This bug is reported by @stanleyxu2005 

Description:
Two immediate query for aggregated metrics may return different result, 
for example, the first query may return (time=1, value=100) (time=2:value=150),(time=3:value=200),
the second query may return (time=1,value=250),(time=3,value=200)

clockfly:


stanleyxu2005:
+1

PR #1773, opened 2015-12-24
fix #1772, When upload application in UI, custom config file is not effective

### LINKED ISSUE #1772: When upload application in UI, custom config file is not effective

clockfly:
Looks like the Typesafe config will check the file suffix.


huafengw:
+1

PR #1776, opened 2015-12-24
fix #1772: better way to parse app id from console output

### LINKED ISSUE #1772: When upload application in UI, custom config file is not effective

stanleyxu2005:


core/src/main/scala/io/gearpump/cluster/client/MasterClient.scala
```diff
@@ -67,4 +67,15 @@ class MasterClient(master : ActorRef) {
     val result = Await.result((master ? AppMastersDataRequest).asInstanceOf[Future[AppMastersData]], Duration.Inf)
     result
   }
+}
+
+object MasterClient {
+
+  val SUBMIT_APPLICATION_SUCCESS_AND_APPID_IS = "Submit application succeed. The application id is "
+
+  def extractAppIdFromConsoleOutput(s: String): Int = {
```
clockfly: I don't like the hack-around here.

clockfly: And I don't see much different between these two version.

stanleyxu2005:
This is not a hack. The command line output is mixed with logs and real output. Without a extract function, how can you extract the app id in a good way?

clockfly:
OK, see your motive. But we need to do it more cleanly.

1. move the constant of function out of MasterClient. It doesn't belong to MasterClient.
2. Use a better name other than SUBMIT_APPLICATION_SUCCESS_AND_APPID_IS, and the pattern should have some uniqueness likely with some special seperator chars. 
3. The constants should be defined in Constant.h.
4. Remove extractAppIdFromConsoleOutput or make it private at where it is used, instead of a public function.

stanleyxu2005:
@clockfly I've revised my patch. Now the extraction will be done in `AppSubmitter`, which is not the part of core.

clockfly:
I don't think it is a good fix, I am +0 on this.

stanleyxu2005:
Yeah. After discussion, it might be better not to search application id in output in general. Close the issue.

PR #1780, opened 2015-12-25
fix #1778, AppSubmitter "-config" option is not used

### LINKED ISSUE #1778: AppSubmitter "-config" option is not used
We already have a "-conf" option on Gear which allows users to set a custom config file.

clockfly:


manuzhang:
Could you add "-conf" options to doc(https://github.com/gearpump/gearpump/blob/master/docs/commandline.md) as well ?

clockfly:
@manuzhang 
 updated.

docs/commandline.md
```diff
@@ -80,5 +80,5 @@ Please check [Deployment for Standalone mode](deployment-standalone.html) for mo
 To start UI server, you can:
 
 ```bash
-services
+services  [-conf <custom gearpump config file>]
```
manuzhang: services has `config` option. Please check https://github.com/gearpump/gearpump/blob/master/services/jvm/src/main/scala/io/gearpump/services/main/Services.scala#L35

clockfly:
Ok, done.

Some default options are not listed, since I think we don't need to document everything.

manuzhang:
+1

PR #1796, opened 2015-12-26
fix #1411: use new api to query task level metrics and show more metr…

### LINKED ISSUE #1411: It take too long for UI metrics chart to display
How to reproduce:
1. submit the application jar by UI
2. Switch to application view
3. Check the charts, it take a long time to display the metrics. (Maybe > 10 seconds?)

stanleyxu2005:
…ics details in processor page (radical performance improvement)


stanleyxu2005:
Screenshots are posted at #1411

clockfly:
Why no bar chart(like task receive throughput distribution)?

stanleyxu2005:
The task metrics table contains more details (total messages, mean rate, ma 1m) than the bar chart. It is another representation form of the data. You can use the sort function to see data skew. 

I admit bar chart will be intuitive to represent the throughput distribution in some case. But if there are more than 200 tasks there, it will be more meaningful to query some metric values by top 10.  

As the page becomes crowd after the change, I don't think it is a good idea to put a bar chart here in the current layout design. I'd rather suggest make a layout change in future release.

manuzhang:
@stanleyxu2005 How about a tab to switch between table and chart ?

clockfly:
+1

PR #1797, opened 2015-12-28
fix #1018 transfer Any in Messages

### LINKED ISSUE #1018: transfer "Any" in messages
Continuing the discussion in #836, `Message` now allows `AnyRef`to be transferred, since kryo needs to know the class name to serialize the Message.

```scala
case class Message(msg: AnyRef, timestamp: TimeStamp = Message.noTimeStamp)
```

After some experiment, I found Kryo is able to serialize  `AnyVal`s that can be unboxed into Java primitives with `DefaultSerializer`. One exception I'm aware of is `BoxedUnit``()`, which needs user to configure like other `AnyRef`s which haven't been registered. 

Hence, I propose we allow transferring `Any` in messages.

huafengw:


clockfly:
@huafengw 

Can you document in class Message that 
"Unit and Null are not valid message type"

clockfly:
+1

PR #1809, opened 2015-12-29
fix #1808, configurable timeout setting for starting a new executor system

### LINKED ISSUE #1808: allow to config start actor system timeout
On a extreme slow machine, like aws nano instance(500MB memory, 1 core), it may take a long period of time to start a new JVM, sometimes more than 20 seconds. 

We should allow the timeout setting to be configurable.

clockfly:
On a extreme slow machine, like aws nano instance(500MB memory, 1 core), it may take a long period of time to start a new JVM, sometimes more than 20 seconds.

We should allow the timeout setting to start a new executor system JVM configurable.

### DIFF
--- a/conf/gear.conf
+++ b/conf/gear.conf
@@ -148,6 +148,10 @@ gearpump {
     http = 8090
   }
 
+  ## Time out setting to start a new executor system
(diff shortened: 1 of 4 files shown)


huafengw:
+1

PR #1811, opened 2015-12-29
fix #1810, improve processing time precision

### LINKED ISSUE #1810: average processing time is almost 0 for simple processing logic like SOL

manuzhang:


streaming/src/main/scala/io/gearpump/streaming/task/TaskActor.scala
```diff
@@ -151,7 +151,7 @@ class TaskActor(
     var done = false
 
     var count = 0
-    val start = System.currentTimeMillis()
+    val start = System.nanoTime()
```
stanleyxu2005: The variable name doesn't contain unit info. Usually gearpump uses millis as *default* unit, but here it is nanos. Maybe add a suffix?

manuzhang: added comments. The nano time is local and not exposed. I'd like to keep it simple.

clockfly:
Why is necessary?
http://stackoverflow.com/questions/19052316/why-is-system-nanotime-way-slower-in-performance-than-system-currenttimemill

PR #1823, opened 2016-01-04
fix #1816 replace akka-actor jar file

### LINKED ISSUE #1816: abnormal gc behavior of Executor
Recently I was tuning Gearpump's performance and I found a randomly abnormal gc behavior of Excutor's jvm.
Here is the snapshot of a normal Executor Jvm's gc behaviour:
normal
The most of gc activity is minor gc and the used heap is growing very slowly.
While here is the abnormal one:
unnormal
You can see the full gc acticity is very often and it eventually will hurt the performance of Task on that Executor.
After continue digging, I found the abnormal jvm will generate a lot of ```akka.dispatch.AbstractNodeQueue$Node``` object:
node
Here it takes about 243MB of memory while the normal Executor jvm does not take that much.

Then I found there is already a similar issue raised in Akka:  https://github.com/akka/akka/issues/17547

huafengw:


clockfly:
+1

Please add more notes about the reason.

Suggest to rename akka_actor as unmanagedlibs
Suggest to add a README.md in unmanagedlibs for the motive, and changes, and link back to #1816 
Suggest to rename akka-actor_2.10-2.3.12-new.jar to somename like  akka-actor_2.10-2.3.12-fix-1816.jar
Suggest to comment in Pack.scala to indicate the chanage is a workaround to #1816, and should be removed in future.

PR #1828, opened 2016-01-04
fix #1779: non-source processors will not calculate receive latency 

### LINKED ISSUE #1779: overall Average Message Receive Latency should not be averaged over Source Processors
since Average Message Receive Latency of a Source Processor is always 0

stanleyxu2005:
the PR also includes some UI changes:

1. Chart's tooltip will disappear when mouse is out of the chart control area
2. Empty table (e.g. empty application list table) now has a better description panel
3. Click on table column's sort arrow will have animation effect 
4. Pagination control would not update when the table is showing some search results.
5. The first hour application will not receive any historical metrics, so the chart for "Past 72 Hours" will show "No Graph Data Available".


stanleyxu2005:
Close the PR, as we want to implement an end-to-end latency for application overview.

PR #1832, opened 2016-01-05
fix #1831, Services exits on bind failure

### LINKED ISSUE #1381: test on travis out of memory

### LINKED ISSUE #1831: Services daemon swallows bind failure
so that a Services process is launched even if the address is already bound

manuzhang:


docs/faq.md
```diff
@@ -33,5 +33,9 @@ Please check whether you are doing blocking jobs (e.g. sleep, IO) in your task.
 
 Generally, we recommend use the default `share-thread-pool-dispatcher` which has better performance and only turn to the `single-thread-dispatcher` when you have to.
 
+##### Why can't I open Dashboard even if the Services process has been launched successfully ?
+
+By default, our Services process binds to a local **IPv6 port**. It's possible that another process on your system has already taken up the same **IPv4 port**. You may check by `lsof -i -P | grep -i "Listen"` if your system is Unix/Linux.
```
clockfly: Can we disallow binding to IPV6?

services/jvm/src/main/scala/io/gearpump/services/RestServices.scala
```diff
@@ -108,7 +105,11 @@ object RestServices {
 
     implicit val materializer = ActorMaterializer()
 
-    Http().bindAndHandle(Route.handlerFlow(services.routes), host, port)
+    Http().bindAndHandle(Route.handlerFlow(services.routes), host, port) onFailure {
+      case t: Throwable =>
+        System.err.println(t)
+        System.exit(-1)
```
clockfly: System.exit is too disruptive. It is dangerous to use it in YARN. Can you throw an exception instead?

manuzhang: exception cannot stop the process from running

docs/faq.md
manuzhang: well, it's possible another process takes the same IPv6 port.

services/jvm/src/main/scala/io/gearpump/services/RestServices.scala
manuzhang: the exception is thrown in future and doesn't stop the process from running

clockfly:
System.exit(-1) is not acceptable. 

It will stop the UT, and also break the YARN module.

docs/faq.md
clockfly: You mean 
When:
1. When another process take the ipv6 port 8090.
2. UI successfully bind to ipv4 port 8090.

the user cannot open dashboard of the ipv4:8090?

manuzhang: yes

clockfly:
@manuzhang 


Are you going to revise the patch?

manuzhang:
oh, yes

manuzhang:
@clockfly updated

clockfly:
+1

PR #1834, opened 2016-01-05
fix #1713 regression on HBaseSink

### LINKED ISSUE #1713: HBaseSink regression from 0.7
We used to be able to add a Configuration parameter to HBaseSink which would a HBaseProperties.ZOOKEEPER_QUORUM to be passed to the constructor. Now this is no longer possible and breaks an example. See below 

```
public abstract class BaseRepository {

    protected final String tableName;
    protected final Configuration hbaseConf;

    public BaseRepository(UserConfig userConfig) {
        String tablePrefix = userConfig.getString(HbaseProperties.TABLE_PREFIX).get();
        tableName = tablePrefix + HbaseValues.TABLE_NAME;

        String zkQuorum = userConfig.getString(HbaseProperties.ZOOKEEPER_QUORUM).get();

        hbaseConf = new Configuration();
        hbaseConf.set(HbaseProperties.ZOOKEEPER_QUORUM, zkQuorum);
    }


    public void createTable() throws IOException {
        HBaseSink sink = getHbaseSink();
        try (Connection conn = sink.connection(); Admin admin = conn.getAdmin()) {
            if (!admin.tableExists(getTableName())) {
                HTableDescriptor table = new HTableDescriptor(getTableName());
                for (String family: HbaseValues.COLUMNS_FAMILIES) {
                    table.addFamily(new HColumnDescriptor(family));
                }
                admin.createTable(table);
            }
        }
    }

    public abstract TableName getTableName();

    protected HBaseSink getHbaseSink() {
        return new HBaseSink(UserConfig.empty(), tableName, hbaseConf);
    }
}
```

huafengw:


clockfly:
You missed here
http://www.gearpump.io/releases/latest/deployment-security.html

huafengw:
The deployment-security doc matches the code.

clockfly:
Why HBaseSecurityUtil need to depends on a UserConfig? It looks stranage if all you need is a key tab file...

huafengw:
It contains the principal and key tab file. 
```
val keytabContent = Files.toByteArray(new File("path_to_keytab_file))
val appConfig = UserConfig.empty
      .withString("gearpump.kerberos.principal", principal)
      .withBytes("gearpump.keytab.file", keytabContent)
```

clockfly:
Maybe this time we can take some time to make sure it is secure.

huafengw:
You mean the current way is not secure?

clockfly:
About HBaseSecurityUtil 

The problem is that when you declare interface like this, it is actually impossible for other to reuse it.
As it accept a UserConfig which is actually a HashMap.

So I suggest either we extract the connect method from HBaseSecurityUtil and make it private to HBaseSink, or change  HBaseSecurityUtil API to be more usable and explicit.

clockfly:
+1

PR #1836, opened 2016-01-05
fix #1779: show critical path latency for application

### LINKED ISSUE #1779: overall Average Message Receive Latency should not be averaged over Source Processors

stanleyxu2005:
the PR also includes some UI changes:

Chart's tooltip will disappear when mouse is out of the chart control area
Empty table (e.g. empty application list table) now has a better description panel
Click on table column's sort arrow will have animation effect
Pagination control would not update when the table is showing some search results.
The first hour application will not receive any historical metrics, so the chart for "Past 72 Hours" will show "No Graph Data Available".


clockfly:
@whjiang 

 Can you take a look at the "critical path" logc?

PR #1858, opened 2016-01-13
fix #1855: when service is unavailable, do not query models without w…

### LINKED ISSUE #1855: UI will try to reconnect to backend in a dead loop with no interval
When the backend UI server is down, the frontend will aggressively query the backend service, even though there is "failure"

In my machine, there are 67 backend calls in 30 seconds. 
Looks like it is a dead loop without setting retry interval. 

I think the same interval setting should applies (3 seconds) no matter it can or cannot access to backend UI.

stanleyxu2005:
…ait interval


clockfly:
+1

PR #1867, opened 2016-01-14
fix #1866, use geardefault.conf to replace reference.conf to avoid conflict

### LINKED ISSUE #1866: Define geardefault.conf to avoid conflict with akka reference.conf
## User story:

We need some default configuration for gearpump. The superseding order is like this:
application.conf > gear.conf > gearpump default configs > reference.conf

Current code use "reference.conf" to store the gearpump default configs. The problem is that for same key, the reference.conf loaded first by classloader will supercede the  reference.conf loaded later. 

That is not desired for us, as we always want "gearpump default configs" to have higher priority.

One solution is to introduce file "geardefault.conf“ which will always override reference.conf.

clockfly:
User story:

We need some default configuration for gearpump. The superseding order is like this:
application.conf > gear.conf > gearpump default configs > reference.conf

Current code use "reference.conf" to store the gearpump default configs. The problem is that for same key, the reference.conf loaded first by classloader will supercede the reference.conf loaded later.

That is not desired for us, as we always want "gearpump default configs" to have higher priority.

One solution is to introduce file "geardefault.conf“ which will always override reference.conf.


manuzhang:
would  `default.conf` be a more concise name ?

clockfly:
the default config file is in classpath, so we have to make it unique in some way. geardefault.conf is more unique than default.conf.

The old impl is wong in some special cases:
For example if set `akka.loglevel = "DEBUG"` in application.conf, it will not be effective. As the reference.conf have set `config base.akka.loglevel="INFO"`. This is wrong.

stanleyxu2005:
Lgtm +1

PR #1869, opened 2016-01-14
fix #1868: show correct data points when switch current and past view

### LINKED ISSUE #1868: Wrong chart data when there are few data points in chart
Shortly after application is launched, the recent metrics data points are few. At this time, switch to "Past 72 hours" and then switch back to current, there will draw duplicated data points.

By the way, when there is only one data point in "Past 72 hours", I don't think we should show the orphan data point at all. I'd prefer show data points when the application runs for over an hour, so that there are at least two data points.

stanleyxu2005:
The RP also includes:

1. Overhauled the chart implementation, so that we can show different set of data points (of past view and current view) in single chart instance.
2. Show past data points in chart, when there are at least 2 data points
3. Renamed `jvm-metrics` to `jvm-metrics-view`, which is for better code understanding
4. Changed `_.pluck` to `_.map` regarding lodash's suggestion
5. Created a constant `DAG_DEATH_UNSPEICIFED` instead of using a magic number
6. Health check dialog might be invisible in some case.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -290,7 +290,7 @@ object Build extends sbt.Build {
       "org.webjars.bower" % "vis" % "4.7.0",
       "org.webjars.bower" % "clipboard.js" % "0.1.1",
       "org.webjars.npm" % "dashing-deps" % "0.1.2",
-      "org.webjars.npm" % "dashing" % "0.4.0"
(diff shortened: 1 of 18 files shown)


clockfly:
+1

PR #1873, opened 2016-01-15
fix #1830: removed cluster side menu and tuned main navbar

### LINKED ISSUE #1830: In mobile phone, cannot switch between master tab and worker tab.
Problem:
In mobile phone, cannot switch between master tab and worker tab.
Looks like the tab switcher is not rendered on mobile phone web browser.

stanleyxu2005:


clockfly:
+1

PR #1888, opened 2016-01-19
fix #1723 add cpu resource isolation

### LINKED ISSUE #1723: Resource isolation per application with linux container
If CPU/memory resource could be limited by application,like Spark on YARN,it would be really great.Do you have some plan to do this?

Thanks.

huafengw:


conf/gear.conf
```diff
@@ -170,6 +184,10 @@ gearpump {
   ### Executor JVM can contains multiple tasks
   ###################
   executor {
+    ## This config only works when cgroup is enabled.
+    ## The value means the number of CPU cores per executor can use, -1 means no limitation.
+    cpu.core.upper.limit = 1
```
clockfly: cpu-core-upper-limit  is better.

clockfly:
Should we put it as an experiment module first? 

As 
1. There is no UT yet, it will impact the UT coverage report. 
2. We need more time to observe whether this is good.

core/src/main/scala/io/gearpump/util/Util.scala
```diff
@@ -63,9 +63,9 @@ object Util {
   }
 
   def startProcess(options : Array[String], classPath : Array[String], mainClass : String,
```
clockfly: I suggest define a new util function instead of modifying current startProcess.

clockfly: For example, UI don't need cgroup. It is confusing to see commandPrefix.

daemon/src/main/java/io/gearpump/cluster/container/CgroupCenter.java
```diff
@@ -0,0 +1,215 @@
+/**
```
clockfly: How about using package name cgroup instead of container?

As all classes under this dir are cgroup related.g

clockfly: And name "container" is not that specific.

docs/deployment-resource-isolation.md
```diff
@@ -0,0 +1,54 @@
+---
+layout: global
+displayTitle: Gearpump Resource Isolation
```
clockfly: Tile Resource Isolation for different streaming applications.

docs/deployment-resource-isolation.md
```diff
@@ -0,0 +1,54 @@
...
+title: Gearpump Resource Isolation
+description: Gearpump Resource Isolation
+---
+
+cgroups (abbreviated from control groups) is a Linux kernel feature to limit, account, and isolate resource usage (CPU, memory, disk I/O, etc.) of process groups.
+
+In Gearpump, we use cgroup to manage CPU resources. You need to check the following items and configuration before use.
+
+*  Check current user's uid and gid in /etc/passwd file. For example if user is gear, then check gear's uid and gid in /etc/passwd file
```
clockfly: What "current user" point to? THe system admin? The application submitter?

docs/deployment-resource-isolation.md
```diff
@@ -0,0 +1,54 @@
...
+*  Cgroups feature is supported or not by current system's kernel
```
clockfly: How to check? Any pointers?

docs/deployment-resource-isolation.md
```diff
@@ -0,0 +1,54 @@
...
+
+   Check /etc/cgconfig.conf exist or not. If not exists, please "yum install libcgroup". If exists, set CPU sub system's mount directory location, and change uid/gid to Gearpump user's uid/gid in that config file, let's take 500 for example, pay attention to the uid/gid mentioned before.
```
clockfly: How to set "set CPU sub system's mount directory location"?

clockfly: "let's take 500 for example, pay attention to the uid/gid mentioned before."

What is 500?

docs/deployment-resource-isolation.md
```diff
@@ -0,0 +1,54 @@
...
+  
+```
+  mount {
```
clockfly: Where this should be added to?

docs/deployment-resource-isolation.md
```diff
@@ -0,0 +1,54 @@
...
+*  Then start cgroup service
+
+```
+service cgconfig restart 
+chkconfig --level 23456 cgconfig on
+```
+Note: cgconfig.conf can only be changed in root mode.
+
+## Enable Cgroups in Gearpump 
+Enable cgroup in Gearpump config file gear.conf under folder conf/
```
clockfly: How to enable?

docs/deployment-resource-isolation.md
```diff
@@ -0,0 +1,54 @@
...
+
+```
+   worker.enable.cgroup: true
```
clockfly: missed the gearpump. prefix?

docs/deployment-resource-isolation.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+## Launch Application
+When submitting an application to Gearpump, user is supposed to modify the configuration ```executor.cpu.core.upper.limit``` in gear.conf, it's the number of CPU cores per executor can use and -1 means no limitation.
```
clockfly: Is this user need to be the same user mentioned before (uid: 500)?

docs/deployment-resource-isolation.md
```diff
@@ -0,0 +1,54 @@
+---
```
clockfly: Suggest to re-organize this document to three sections.

1. From admin's perspective. How to config system, gearpump to enable cgroup.
2. From app submitter's perspective, how to run a app in a cluster which have cgroup enabled. 
3. What is the known limitations? How this behave on Windows? What happens when a user submit app without configuring cgroup? Is the jar submission by UI still working?

daemon/src/main/scala/io/gearpump/cluster/worker/Worker.scala
```diff
@@ -349,6 +351,8 @@ private[cluster] object Worker {
           file.getFile
         }
 
+        val cgroupCommand = cgroupManager.map(_.startNewExecutor(config, launch.resource.slots, launch.appId, launch.executorId)).getOrElse(List.empty)
```
clockfly: What this means? What if cgroup command don't exist?

conf/gear.conf
clockfly: Suggest to group al cgroup stuffs under entry " cgroup"

clockfly: Like cgroup.enabled = true
cgroup.executor-cpu-core-limit = 1
cgroup.rootdir = ''

daemon/src/main/java/io/gearpump/cluster/container/CgroupCenter.java
```diff
@@ -0,0 +1,215 @@
...
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.*;
+
+public class CgroupCenter implements CgroupOperation {
+
+    public static Logger LOG = LoggerFactory.getLogger(CgroupCenter.class);
```
manuzhang: please fix the indentation

daemon/src/main/scala/io/gearpump/cluster/worker/CGroupManager.scala
```diff
@@ -0,0 +1,156 @@
...
+    config.getInt(Executor_CPU_CORE_UPPER_LIMIT)
+  }
+
+  def isEnabledCgroup(config: Config): Boolean = {
+    config.getBoolean(WORKER_ENABLE_CGROUP)
+  }
+
+  def getInstance(config: Config): Option[CGroupManager] = {
+    if (isEnabledCgroup(config)) {
+      if (SystemUtils.IS_OS_WINDOWS) {
```
manuzhang: please cover OS X as well

clockfly:
Too much change for core or daemon.

I expect only a few line of changes.

clockfly:
+1

PR #1905, opened 2016-01-20
fix #1904, fixed "DAG canvas context menu cannot be displayed."

### LINKED ISSUE #1904: [Mobile] DAG canvas context menu cannot be rendered.
Root cause:
On mobie, we don't have right click, so we use "touch and hold" to simulate the right click. 
We need to handle "hold" event in visjs network canvas.

clockfly:


stanleyxu2005:
+1

clockfly:
Thanks!

PR #1907, opened 2016-01-20
fix #1904, disable the default context menu for canvas.

### LINKED ISSUE #1904: [Mobile] DAG canvas context menu cannot be rendered.

clockfly:
Default context menu shows "Save Image, Copy Image..."


stanleyxu2005:
+1 indeed

Sorry, I did not put a comment about that before, so that when you removed this line from previous commit, but I just thought it is okay.

PR #1914, opened 2016-01-21
fix #1913, add warning for detected loop in DAG.

### LINKED ISSUE #1913: Add warning for detected loop in DAG.
We must give warning before in ``object StreamApplication``
```
val indices = dag.topologicalOrderIterator.toList.zipWithIndex.toMap
``` 

Otherwise, it will stuck in infinite loop (because topological order operation cannot be applied in graph with loop) and give no warnings about loop in DAG.

pangolulu:


huafengw:
We just give a warning? Samoa is still can't be correctly launched in Gearpump

streaming/src/main/scala/io/gearpump/streaming/StreamApplication.scala
```diff
@@ -144,6 +139,11 @@ object StreamApplication {
   def apply[T <: Processor[Task], P <: Partitioner] (name : String, dag: Graph[T, P], userConfig: UserConfig): StreamApplication = {
     import Processor._
 
+    val LOG = LogUtil.getLogger(this.getClass)
```
huafengw: move the logger out of apply function

pangolulu: OK. Because if there is loop in graph,  ``val indices = dag.topologicalOrderIterator.toList.zipWithIndex.toMap`` will be in infinite loop. And SAMOA task  get stucked.

huafengw: I know, do we want to have a fix there?

pangolulu: If we force it available to contain loops, does gearpump fails in the runtime?

clockfly:
+1

PR #1915, opened 2016-01-21
fix #1896, move storm examples under integration test

### LINKED ISSUE #1896: move Storm examples under integrationtest
As discussed in https://github.com/gearpump/gearpump/pull/1863#issuecomment-171846356, Storm examples are only used for integration test and we'd better move them there to avoid confusion for users

manuzhang:
The motivation is to move examples for integration test away from users.

changes:

1. move `integrationtest/src` to `integrationtest/core/src`
2. move `examples/streaming/storm09` and `examples/streaming/storm010` to `integrationtest/storm09` and `integrationtest/storm010`
3. move integration test build scripts to `BuildIntegrationTest.scala`
4. pack `integrationtest/target/*` to `pack/integrationtest`, which is excluded from `packArchive`.


stanleyxu2005:
+1 in general

If I remember correctly, we don't involve integration test in build file, which speed up the build process. I'd continue to see this part of code out of the build.

manuzhang:
integration test is [in our build file]( https://github.com/gearpump/gearpump/blob/master/project/Build.scala#L220). This PR doesn't change this behavior.

PR #1925, opened 2016-01-25
fix #1924: fix storm test spec failure

### LINKED ISSUE #1924: Storm Integration test spec cannot be started in case it starts after stability tests
I noticed the whole test suite always ignores `StormCompatibilityTestSpec`. If I run this spec individually, it works. There is some conditional code impacts the result.

stanleyxu2005:


huafengw:
+1

PR #1927, opened 2016-01-25
fix #1926 fix client side configuration does not work in CGroup

### LINKED ISSUE #1926: [CGroup] client side configuration does not work

huafengw:


daemon/src/test/scala/io/gearpump/cluster/MiniCluster.scala
```diff
@@ -63,7 +63,7 @@ class MiniCluster {
     val workerListFuture = (master ? GetAllWorkers).asInstanceOf[Future[WorkerList]]
 
     // wait until the worker is registered.
-    val workers = Await.result[WorkerList](workerListFuture, 5 seconds)
+    val workers = Await.result[WorkerList](workerListFuture, 15 seconds)
```
stanleyxu2005: Can we externalize the const?

huafengw: Maybe in we can do it in another pr

stanleyxu2005:
+1

PR #1940, opened 2016-01-29
fix #1932, client config overwrites critical system config

### LINKED ISSUE #1932: client config overwrites critical system config
1. Critical system configs (e.g. `gearpump.home`) should not be allowed to be overwritten by client submission config
2. application config shown on UI is not the final merged configs

clockfly:


daemon/src/main/scala/io/gearpump/cluster/worker/Worker.scala
```diff
@@ -303,14 +304,29 @@ private[cluster] object Worker {
...
       }.getOrElse(ConfigFactory.empty())
-      userConfig.withFallback(systemConfig)
+
+      resolveExecutorConfig(workerConfig, submissionConfig)
+    }
+
+    // For some config, worker has priority, for others, user Application submission config
+    // have priorities.
+    private def resolveExecutorConfig(workerConfig: Config, submissionConfig: Config): Config = {
+      val config = submissionConfig.withoutPath(Constants.GEARPUMP_HOSTNAME)
```
stanleyxu2005: The code is good. But from user perspective, better to print a warning about what config items have been ignored.

manuzhang:
+1

PR #1941, opened 2016-01-29
fix #1935, add topological sort algorithm for graph with circles.

### LINKED ISSUE #1935: Add topological sort algorithm for graph with circles
The algorithm can give the best topological order for graph with circles.
A reference of this algorithm is http://www.drdobbs.com/database/topological-sorting/184410262

pangolulu:


clockfly:
Thanks! @pangolulu

clockfly:
Can you add the algorithm link you use in the source code.
Can you add some unit test for it? The test is in file GraphSpec.

pangolulu:
The algorithm is following [http://www.drdobbs.com/database/topological-sorting/184410262](). @clockfly

core/src/main/scala/io/gearpump/util/Graph.scala
```diff
@@ -254,6 +254,90 @@ class Graph[N, E](vertexList: List[N], edgeList: List[(N, E, N)]) extends Serial
   }
 
   /**
+   * Return all circles in graph.
+   */
+  def findCircles: mutable.Set[mutable.Set[N]] = {
+    val inStack = mutable.Map[N, Boolean]()
+    val stack = mutable.Stack[N]()
+    val DFN = mutable.Map[N, Int]()
```
clockfly: DFN, LOW is not good varible name. Maybe index and lowlink better? 
https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm

core/src/main/scala/io/gearpump/util/Graph.scala
```diff
@@ -254,6 +254,90 @@ class Graph[N, E](vertexList: List[N], edgeList: List[(N, E, N)]) extends Serial
...
+    }
+
+    circles
+  }
+
+  /**
+   * Return an iterator of vertex in topological order of graph with circles
+   * The node returned by Iterator is stable sorted.
+   */
+  def topologicalOrderWithCirclesIterator: Iterator[N] = {
```
clockfly: Are you sure it is "stable sorted"?

Since you are using Set, and Set is not stable when expanded.

core/src/test/scala/io/gearpump/util/GraphSpec.scala
```diff
@@ -178,6 +178,30 @@ class GraphSpec extends PropSpec with PropertyChecks with Matchers {
...
+    graph.addEdge("B", defaultEdge, "C")
+
+    val topoNoCircles = graph.topologicalOrderIterator
+    val topoWithCircles = graph.topologicalOrderWithCirclesIterator
+
+    assert(topoNoCircles.zip(topoWithCircles).forall(x => x._1 == x._2))
+  }
+
+  property("Topological sort of graph with circles should work properly") {
+    val graph = Graph(1 ~> 3 ~> 4 ~> 6 ~> 5, 4 ~> 1, 1 ~> 2 ~> 4)
```
clockfly: Can you add a test case which have two strong connected components, and one component ranks over the other by topology order. 

So that we can examine the behavior for multiple cycles.

core/src/main/scala/io/gearpump/util/Graph.scala
```diff
@@ -254,6 +254,90 @@ class Graph[N, E](vertexList: List[N], edgeList: List[(N, E, N)]) extends Serial
   }
 
   /**
+   * Return all circles in graph.
+   */
+  def findCircles: mutable.Set[mutable.Set[N]] = {
```
clockfly: Is the findCircles used elsewhere? if not, probably make it private

clockfly: Please add the algorithm link as comments in source code.

clockfly:
How about making this change, change the reference of topologicalOrderIterator to use topologicalOrderWithCirclesIterator instead, and see whether the SAMOA module works?

pangolulu:
@clockfly 
I have changed the variables name, and make the sort stable using ``MutableList``.

I changed the GraphSpec to test whether ``topologicalOrderWithCirclesIterator`` equals ``topologicalOrderIterator`` in the graph with no circle but contains more than one sort to see if the actual sort is stable.

And test ``topologicalOrderIterator`` in the complex graph with more than one circles.

pangolulu:
@clockfly 
Yes, I will try ``topologicalOrderWithCirclesIterator `` in SAMOA.

clockfly:
There are several reference of topologicalOrderIterator in Gearpump, you need to replace them all.

pangolulu:
@clockfly Yes, ok.

clockfly:
@pangolulu 

How is the result?

pangolulu:
@clockfly 

I meet an error, it also happens before. I am debugging it. Maybe it is the problem of gearpump version used in SAMOA.

error

darionyaphet:
@pangolulu 

`http://www.drdobbs.com/database/topological-sorting/184410262`  seems wrong  ...

pangolulu:
@darionyaphet 

what's the problem?

darionyaphet:
@pangolulu the links couldn't point to the wanted page.

pangolulu:
@clockfly 

It seems that when I use gearpump ``<version>0.7.5</version>`` in SAMOA ``pom.xml``, it will go wrong. And I change it to ``<version>0.7.4</version>``, it's right now.

I test ``VerticalHoeffdingTree``  with ``topologicalOrderWithCirclesIterator``, the app can be submitted now. But it meets another problem,  ``VerticalHoeffdingTree`` has duplicated edges, so gearpump will raise an exception for duplicated edges (``attribute`` and ``control`` edge).

vht

duplited edges

Next, I will comment ``require(!dag.hasDuplicatedEdge(), "Graph should not have duplicated edges")``, and see if it will go correct.

clockfly:
@huafengw 

About the duplicate edges in SAMOA, I think we took care of it before? By re-organizing the message format?

clockfly:
@pangolulu Can you investigate what is wrong on 0.7.5?

huafengw:
I check the samoa-gearpump code and we did not combine duplicated edges into one.

pangolulu:
@huafengw 

Can we deal with the duplicated edges by a unique edge with different message type?

clockfly:
+1

PR #1946, opened 2016-02-02
fix 945 fix stability integration test

huafengw:


integrationtest/core/src/it/scala/io/gearpump/integrationtest/checklist/StabilitySpec.scala
```diff
@@ -70,7 +70,6 @@ class StabilitySpec extends TestSpecBase {
   "kill worker" should {
     "worker will not recover but all its executors will be migrated to other workers" in {
       // setup
-      restartClusterRequired = true
```
stanleyxu2005: Without this line, the next test case will have one worker less. Consider all test cases are independent or executed randomly, what is the sequence, if cluster has one worker less?

stanleyxu2005:
Lgtm +1

PR #1948, opened 2016-02-02
fix #1895, make GearpumpNimbus a standalone service

### LINKED ISSUE #1895: make GearpumpNimbus a stand-alone service
The current solution for Storm compatibility work in this way.

1. a user submits a Storm topology with a uber jar through the `storm -jar` command
2. a local `GearpumpNimbus` is launched
3.  user topology is executed with the address of `GearpumpNimbus`
4. topology is submitted to the `GearpumpNimbus` and translated to Gearpump DAG
5. Gearpump DAG is submitted to Gearpump master

Users are limited to what `storm -jar` command provides and the local `GearpumpNimbus` is launched and used only for once. 

There is another use case as in [StreamCQL](https://github.com/HuaweiBigData/StreamCQL). A `CQLClient` is launched where users could continuously execute CQL sentences which are translated to Storm topologies and submitted to `Nimbus`. 

I'm thinking about making the `GearpumpNimbus` a stand-alone service like Storm's `Nimbus` such that there will be no limitation on usage for Storm applications.

manuzhang:
the storm integration test is not yet working so I comment it out. I've manually verified the storm over gearpump functionality. I'd like to fix the integration test as follow-up PR.


clockfly:
+1

PR #1949, opened 2016-02-02
fix #1910 enlarge timeout in MasterClient

### LINKED ISSUE #1910: enlarge timeout in MasterClient
Currently, the timeout for client to submit/resolve/shutdown/list application is 15s. It is too small for resource limited environment. 

Suggestion:
1. Enable this timeout to 90s by default. I don't mean changing the overall timeout setting. Instead, ONLY the timeout setting for MasterClient shall be changed.
2. Make this timeout setting a parameter. So that, we can allow user to change its value for their need. Actually, this parameter is a pure client setting. User know better than us on its value.

whjiang:
enlarge timeout to 90s and make it configurable.

### DIFF
--- a/core/src/main/resources/geardefault.conf
+++ b/core/src/main/resources/geardefault.conf
@@ -206,6 +206,9 @@ gearpump {
   ### the appmaster will shutdown itself.
   resource-allocation-timeout-seconds = 10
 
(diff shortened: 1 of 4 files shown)


clockfly:
+1

PR #1951, opened 2016-02-03
fix #1943 allow user to config how many executors to use in an applic…

### LINKED ISSUE #1943: Allow user to config how many executors to use in an application

huafengw:
This pr is the basic version, not include limiting the executor number when dynamic dag.

### DIFF
--- a/conf/gear.conf
+++ b/conf/gear.conf
@@ -62,10 +62,13 @@ gearpump {
   ### When the resource cannot be allocated in the timeout, then
   ### the appmaster will shutdown itself.
   resource-allocation-timeout-seconds = 120
-  
+
   ##
   ## Executor share same process of worker
   worker.executor-share-same-jvm-as-worker = false
+
+  ## Number of executors to launch when starting an application
+  application.executor-num = 1
 
   ###########################
   ### Change the dispather for tasks
--- a/core/src/main/scala/io/gearpump/util/Constants.scala
+++ b/core/src/main/scala/io/gearpump/util/Constants.scala
@@ -162,5 +162,5 @@ object Constants {
 
   val PREFER_IPV4 = "java.net.preferIPv4Stack"
 
-
+  val APPLICATION_EXECUTOR_NUMBER = "gearpump.application.executor-num"
 }
--- a/daemon/src/main/scala/io/gearpump/cluster/main/AppSubmitter.scala
+++ b/daemon/src/main/scala/io/gearpump/cluster/main/AppSubmitter.scala
@@ -37,6 +37,7 @@ object AppSubmitter extends AkkaApp with ArgumentsParser {
   override val options: Array[(String, CLIOption[Any])] = Array(
     "namePrefix" -> CLIOption[String]("<application name prefix>", required = false, defaultValue = Some("")),
     "jar" -> CLIOption("<application>.jar", required = true),
+    "executors" -> CLIOption[Int]("number of executor to launch", required = false, defaultValue = Some(1)),
     "verbose" -> CLIOption("<print verbose log on console>", required = false, defaultValue = Some(false)),
     // For document purpose only, OPTION_CONFIG option is not used here.
     // OPTION_CONFIG is parsed by parent shell command "Gear" transparently.
@@ -58,6 +59,7 @@ object AppSubmitter extends AkkaApp with ArgumentsParser {
 
     // Set jar path to be submitted to cluster
     System.setProperty(Constants.GEARPUMP_APP_JAR, jar)
+    System.setProperty(Constants.APPLICATION_EXECUTOR_NUMBER, config.getInt("executors").toString)
 
     val namePrefix = config.getString("namePrefix")
     if (namePrefix.nonEmpty) {
--- a/daemon/src/main/scala/io/gearpump/cluster/scheduler/PriorityScheduler.scala
+++ b/daemon/src/main/scala/io/gearpump/cluster/scheduler/PriorityScheduler.scala
@@ -26,14 +26,13 @@ import io.gearpump.cluster.scheduler.Scheduler.PendingRequest
 
 import scala.collection.mutable
 
-class PriorityScheduler extends Scheduler{
-
+class PriorityScheduler extends Scheduler {
   private var resourceRequests = new mutable.PriorityQueue[PendingRequest]()(requestOrdering)
 
   def requestOrdering = new Ordering[PendingRequest] {
     override def compare(x: PendingRequest, y: PendingRequest) = {
       var res = x.request.priority.id - y.request.priority.id
-      if(res == 0)
+      if (res == 0)
         res = y.timeStamp.compareTo(x.timeStamp)
       res
     }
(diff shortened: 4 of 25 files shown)


conf/gear.conf
```diff
@@ -76,6 +75,10 @@ gearpump {
   ### -XX:+PrintGCApplicationStoppedTime
   verbose-gc = false
 
+  ##The number of Executors to launch when first starting the streaming application
+  ##The default value -1 means no limitation, each Worker will spawn only one Executor.
+  application.executor-num = -1
```
manuzhang: why not 1 ? `no limitation` looks contrary to `only one Executor`

huafengw: each Worker will spawn only one Executor, it's not 1

huafengw:
Now when user submit an application from command line, user can specify the executor number like
```
bin/gear app -jar XX.jar -executors 4 MAIN_CLASS
```

manuzhang:
would you mind updating the documentation ?

services/jvm/src/main/scala/io/gearpump/services/MasterService.scala
```diff
@@ -201,10 +201,10 @@ object MasterService {
   /**
    * Submit Native Application.
    */
-  def submitGearApp(jar: Option[File], args: String, systemConfig: Config, userConfigFile: Option[File]): Boolean = {
+  def submitGearApp(jar: Option[File], executorNum: String, args: String, systemConfig: Config, userConfigFile: Option[File]): Boolean = {
```
clockfly: executorNum should be type Int?

clockfly: It is not safe to use String, as the executorNum is user provided value from http request, we need to sanitize it.

services/dashboard/views/apps/submit/submit.js
```diff
@@ -10,6 +10,7 @@ angular.module('dashboard')
 
       $scope.dialogTitle = 'Submit Gearpump Application';
       $scope.confFileSuffix = '.conf';
+      $scope.executorNum = 1;
```
clockfly: It means user cannot customize the executorNum in UI?

docs/commandline.md
```diff
@@ -18,9 +18,11 @@ If you use Maven you can have a look [here](https://maven.apache.org/plugins/mav
 You can use the command `gear` under the bin directory to submit, query and terminate an application:
 
 ```bash
-gear app [-namePrefix <application name prefix>] [-conf <custom gearpump config file>] -jar xx.jar MainClass <arg1> <arg2> ...
+gear app [-namePrefix <application name prefix>] [-executors <number of executors to launch>] [-conf <custom gearpump config file>] -jar xx.jar MainClass <arg1> <arg2> ...
 ```
 
+If you not specify the option ```executors```, the number of executors to be launched will be the same as living number of Workers.
```
clockfly: Maybe we should just use 1 as the default value. Maybe it is safer compared with launching as mang executors as workers.

clockfly: Then we don't need to specially handling the case when executor Number is "-1"

daemon/src/test/scala/io/gearpump/cluster/scheduler/PrioritySchedulerSpec.scala
```diff
@@ -136,4 +135,42 @@ class PrioritySchedulerSpec(_system: ActorSystem) extends TestKit(_system) with
       mockAppMaster.expectMsg(5 seconds, ResourceAllocated(Array(ResourceAllocation(Resource(60), mockWorker1.ref, workerId1))))
     }
   }
+
+  "The PriorityScheduler" should {
+    "handle the resource request with different executor number" in {
```
clockfly: This is a long test, 36 lines. Please document some key steps so we know how the test works.

daemon/src/main/scala/io/gearpump/cluster/main/AppSubmitter.scala
```diff
@@ -58,6 +59,7 @@ object AppSubmitter extends AkkaApp with ArgumentsParser {
 
     // Set jar path to be submitted to cluster
     System.setProperty(Constants.GEARPUMP_APP_JAR, jar)
+    System.setProperty(Constants.APPLICATION_EXECUTOR_NUMBER, config.getString("executors"))
```
clockfly: Can we avoid changing the system property? Since it is a global env.

daemon/src/main/scala/io/gearpump/cluster/main/AppSubmitter.scala
```diff
@@ -37,6 +37,7 @@ object AppSubmitter extends AkkaApp with ArgumentsParser {
   override val options: Array[(String, CLIOption[Any])] = Array(
     "namePrefix" -> CLIOption[String]("<application name prefix>", required = false, defaultValue = Some("")),
     "jar" -> CLIOption("<application>.jar", required = true),
+    "executors" -> CLIOption[String]("number of executor to launch", required = false, defaultValue = Some("-1")),
```
clockfly: Type should be Int?

core/src/main/scala/io/gearpump/util/Constants.scala
```diff
@@ -151,4 +151,6 @@ object Constants {
   val GEARPUMP_UI_AUTHENTICATOR_CLASS = "gearpump.ui-security.authenticator"
 
   val PREFER_IPV4 = "java.net.preferIPv4Stack"
+
+  val APPLICATION_EXECUTOR_NUMBER = "gearpump.application.executor-num"
```
clockfly: If this means a new config, then we should also add this into geardefault.conf and gear.conf.

services/dashboard/views/apps/submit/submit.js
huafengw: That's actually the default value

daemon/src/main/scala/io/gearpump/cluster/scheduler/PriorityScheduler.scala
```diff
@@ -70,56 +69,78 @@ class PriorityScheduler extends Scheduler{
...
+            allocated = allocated + request.resource
+            resourcesSnapShot.update(request.workerId, (worker, availableResource - request.resource))
           } else {
             scheduleLater = scheduleLater :+ PendingRequest(appId, appMaster, request, timeStamp)
           }
       }
     }
-    for(request <- scheduleLater)
-      resourceRequests.enqueue(request)
+    scheduleLater.foreach(resourceRequests.enqueue(_))
```
clockfly: This doesn't make much sense. The left side is more readble.

services/dashboard/views/apps/submit/submit.html
```diff
@@ -41,6 +41,10 @@ <h4 ng-bind="dialogTitle"></h4>
               ng-model="conf" accept-pattern="{{confFileSuffix}}"></form-control>
             <!-- input 3 -->
             <form-control
+              type="integer" min="1" label="Executor Number" ng-hide="isStormApp"
```
clockfly: Why hiding this from storm?

daemon/src/main/scala/io/gearpump/cluster/scheduler/PriorityScheduler.scala
```diff
@@ -70,56 +69,78 @@ class PriorityScheduler extends Scheduler{
...
-          if (resourcesSnapShot.contains(request.workerId)) {
-            val (worker, availableResource) = resourcesSnapShot.get(request.workerId).get
-            if (availableResource > request.resource) {
-              appMaster ! ResourceAllocated(Array(ResourceAllocation(request.resource, worker, request.workerId)))
-              allocated = allocated  + request.resource
-              resourcesSnapShot.update(request.workerId, (worker, availableResource - request.resource))
-            }
+          val workerAndResource = resourcesSnapShot.get(request.workerId)
+          if (workerAndResource.nonEmpty && workerAndResource.get._2 > request.resource) {
+            val (worker, availableResource) = workerAndResource.get
```
clockfly: Safe to skip the check "availableResource > request.resource"?

daemon/src/main/scala/io/gearpump/cluster/scheduler/PriorityScheduler.scala
```diff
@@ -70,56 +69,78 @@ class PriorityScheduler extends Scheduler{
...
           }
         case SPECIFICWORKER =>
-          if (resourcesSnapShot.contains(request.workerId)) {
-            val (worker, availableResource) = resourcesSnapShot.get(request.workerId).get
-            if (availableResource > request.resource) {
-              appMaster ! ResourceAllocated(Array(ResourceAllocation(request.resource, worker, request.workerId)))
-              allocated = allocated  + request.resource
-              resourcesSnapShot.update(request.workerId, (worker, availableResource - request.resource))
-            }
+          val workerAndResource = resourcesSnapShot.get(request.workerId)
```
clockfly: What you have changed??? I don't see different with code before change.

huafengw: There was a bug before the change. If worker's remaining resource is not larger than the required resource, the request is not put back.

daemon/src/main/scala/io/gearpump/cluster/scheduler/PriorityScheduler.scala
huafengw: I think so because it requests resources on a specific worker

clockfly:
+1

PR #1955, opened 2016-02-04
fix #1952, document how to use StreamCQL over Gearpump

### LINKED ISSUE #1952: Document how to use StreamCQL over Gearpump
I know there are document in project StrealCQL. We should also add some basic documents in Gearpump, so that users know Gearpump can be used in this way.

manuzhang:


clockfly:
+1

PR #1959, opened 2016-02-05
fix #1958: better test report organizing

### LINKED ISSUE #1958: Better test report organizing
1. The integration test is running in a Docker container as root user. Any produced files that on a mounted path will be changed to root. This might prevent Gearpump running locally. Test logs and jarstore files should be written to some temporary location of docker container, which will be mounted individually
2. Cluster might be restarted several times during the test. It will be better to group test log files into different folders.

stanleyxu2005:


integrationtest/core/src/main/scala/io/gearpump/integrationtest/minicluster/RestClient.scala
```diff
@@ -53,13 +53,21 @@ class RestClient(host: String, port: Int) {
...
   }
 
+  private def decodeAs[T: upickle.default.Reader](expr: String): T = try {
+    read[T](expr)
+  } catch {
+    case ex: Throwable =>
+      LOG.error(ex)
+      throw ex
+  }
+
```
kkasravi: I didn't think read[T] works - that it needs a real type.

kkasravi:
+1 as long as read[T] works as noted above.

PR #1960, opened 2016-02-09
A few documentation points

rvvincelli:
Hi, I've added a few clarifications to the documentation after I tested a small Gearpump Kafka-HBase wordcount application - thank you everything worked fine, Kerberos configuration included!

There would be one more point but it is discussed [here](https://github.com/gearpump/gearpump/pull/1077) already, please have a look if you have time.

Cheers!


kkasravi:
+1

PR #1965, opened 2016-02-14
fix #1892: added instruction text to operate network graph 

### LINKED ISSUE #1892: Improve the UI experience
These important functions in UI can be easily overlooked, due to the UI design:
1. Compose the DAG. User can compose the application by drag and drop in UI.
2. Change parallelism in DAG canvas. User can change the parallelism in runtime without message loss.
3. Replace DAG in DAG canvas, user can replace partial nodes of the DAG, without impacting upstream and downstream nodes. 

We should do some improvement to change it.

ui_improve

stanleyxu2005:
the PR also includes minor tweaks of context menu


clockfly:
+1

PR #1968, opened 2016-02-16
fix #1966 make Partitioner API Java compatible

### LINKED ISSUE #1966: MulticastPartitioner API is hard to extend in Java
It's easier to use Java collections in Scala than the other way around,but the function ```getPartitions``` returns ```scala.collection.immutable.List```, so it is not user-friendly if user want to implement his own MulticastPartitioner.

huafengw:
Also include DataSource API change

### DIFF
--- a/core/src/main/scala/io/gearpump/partitioner/BroadcastPartitioner.scala
+++ b/core/src/main/scala/io/gearpump/partitioner/BroadcastPartitioner.scala
@@ -21,7 +21,14 @@ package io.gearpump.partitioner
 import io.gearpump.Message
 
 class BroadcastPartitioner extends MulticastPartitioner {
-  override def getPartitions(msg: Message, partitionNum: Int, currentPartitionId: Int): List[Int] = {
-    (0 until partitionNum).toList
+  private var lastPartitionNum = -1
+  private var partitions = Array.empty[Int]
+
+  override def getPartitions(msg: Message, partitionNum: Int, currentPartitionId: Int): Array[Int] = {
+    if (partitionNum != lastPartitionNum) {
+      partitions = (0 until partitionNum).toArray
+      lastPartitionNum = partitionNum
+    }
+    partitions
   }
 }
--- a/core/src/main/scala/io/gearpump/partitioner/Partitioner.scala
+++ b/core/src/main/scala/io/gearpump/partitioner/Partitioner.scala
@@ -41,9 +41,9 @@ trait UnicastPartitioner extends Partitioner {
 }
 
 trait MulticastPartitioner extends Partitioner {
-  def getPartitions(msg: Message, partitionNum: Int, currentPartitionId: Int): List[Int]
+  def getPartitions(msg: Message, partitionNum: Int, currentPartitionId: Int): Array[Int]
 
-  def getPartitions(msg: Message, partitionNum: Int): List[Int] = {
+  def getPartitions(msg: Message, partitionNum: Int): Array[Int] = {
     getPartitions(msg, partitionNum, Partitioner.UNKNOWN_PARTITION_ID)
   }
 }
--- a/examples/streaming/stockcrawler/src/main/scala/io/gearpump/streaming/examples/stock/QueryServer.scala
+++ b/examples/streaming/stockcrawler/src/main/scala/io/gearpump/streaming/examples/stock/QueryServer.scala
@@ -21,29 +21,29 @@ package io.gearpump.streaming.examples.stock
 
 import java.util.concurrent.TimeUnit
 
-import akka.actor.{Props, Actor}
-import akka.io.IO
-import io.gearpump.streaming.ProcessorId
-import io.gearpump.streaming.appmaster.{ProcessorSummary, StreamAppMasterSummary, AppMaster}
-import AppMaster.{TaskActorRef, LookupTaskActorRef}
 import akka.actor.Actor._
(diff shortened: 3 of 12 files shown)


manuzhang:
could you add a Java UT to make sure this works ?

streaming/src/test/scala/io/gearpump/streaming/task/SubscriptionSpec.scala
```diff
@@ -22,7 +22,7 @@ import java.util.Random
 
 import io.gearpump.Message
 import io.gearpump.cluster.UserConfig
-import io.gearpump.partitioner.{Partitioner, HashPartitioner}
+import io.gearpump.streaming.partitioner.{HashPartitioner, Partitioner}
```
clockfly: Chaning package name will introduce compatibility issue.

external/kafka/src/main/scala/io/gearpump/streaming/kafka/KafkaSource.scala
```diff
@@ -172,7 +172,7 @@ class KafkaSource(
           count += 1
         }
     }
-    messageBuffer.toList
+    messageBuffer.toArray
```
manuzhang: Do you think it's better to use Array here to avoid memory copy ?

streaming/src/test/scala/io/gearpump/streaming/task/SubscriptionSpec.scala
manuzhang: Do we already break it with the Partitioner API change ?

huafengw: So not change the package?

clockfly: @manuzhang 

As far as I know, Partitioner and HashPartitioner are not changed. 
MulticastPartitioner is not directly used by end user.

external/kafka/src/main/scala/io/gearpump/streaming/kafka/KafkaSource.scala
huafengw: But I can not know the size of the result array in advance.

manuzhang: we have the `batchSize`

huafengw:
@manuzhang I try to optimize the KafkaSource read function, please review the change.

clockfly:
+1

external/kafka/src/main/scala/io/gearpump/streaming/kafka/KafkaSource.scala
```diff
@@ -161,18 +161,20 @@ class KafkaSource(
...
-        var count = 0
-        while (count < batchSize) {
-          fetch.poll.flatMap(filterMessage).foreach(messageBuffer += _)
+  override def read(batchSize: Int): Array[Message] = {
+    var count = 0
+    val messages = new Array[Message](batchSize)
+    fetchThread.foreach { fetch =>
+      (0 until batchSize).foreach { _ =>
+        fetch.poll.flatMap(filterMessage).foreach { message =>
+          messages.update(count, message)
```
manuzhang: what does `update` do ? Why not `messages(count) = message` ?

external/kafka/src/main/scala/io/gearpump/streaming/kafka/KafkaSource.scala
```diff
@@ -161,18 +161,20 @@ class KafkaSource(
...
           count += 1
         }
+      }
     }
-    messageBuffer.toList
+    if (count < batchSize - 1) {
```
manuzhang: `count < batchSize` ?

huafengw: count starts with zero

external/kafka/src/main/scala/io/gearpump/streaming/kafka/KafkaSource.scala
```diff
@@ -161,18 +161,20 @@ class KafkaSource(
...
+        fetch.poll.flatMap(filterMessage).foreach { message =>
+          messages(count) = message
           count += 1
         }
+      }
     }
-    messageBuffer.toList
+    if (count < batchSize) {
+      messages.take(count)
+    } else messages
```
manuzhang: I'm a bit worried about this. 

What if 

* a user takes an array but reads an index that doesn't exist ?
* a user `read` twice ? It seems the behavior is undetermined. It could be the original array. It could be a shrunk copy.

huafengw: The original implementation has the same problem

core/src/main/scala/io/gearpump/partitioner/BroadcastPartitioner.scala
```diff
@@ -21,7 +21,14 @@ package io.gearpump.partitioner
 import io.gearpump.Message
 
 class BroadcastPartitioner extends MulticastPartitioner {
-  override def getPartitions(msg: Message, partitionNum: Int, currentPartitionId: Int): List[Int] = {
-    (0 until partitionNum).toList
+  var lastPartitionNum = -1
```
manuzhang: please make it private if not accessed elsewhere.

manuzhang:
+1

PR #1970, opened 2016-02-17
fix #1938, add DataSource Java API

### LINKED ISSUE #1938: Scala private[package] method is public in Java
As per [this thread](https://groups.google.com/forum/#!searchin/scala-user/package$20private$20public$20in$20java/scala-user/nW76vx900XQ/ls6UdilRNdcJ), Scala private[package] is public in Java (JVM bytecode). That means we should avoid private[package] in public API or add Java wrappers.

manuzhang:
1. DataSource.read returns `Iterator[Message]` instead of `List[Message]`
2. add `DataSource.java` and `KafkaSource.java`


manuzhang:
Need more check for compatibility and performance regression.

clockfly:
Why cannot keep an interface that works for both scala and java?

PR #1973, opened 2016-02-19
fix #1972, backoff retry kafka consuming on exception

### LINKED ISSUE #1972: Backoff retry consuming from Kafka on broker changes
The current way stops consuming immediately if kafka broker is down temporarily. We'd better apply a backoff strategy to retry connecting to broker and restart consuming.

manuzhang:


clockfly:
I think the impl is over-complicated. Probably we can flatten the code to more easy to read and understand.

manuzhang:
@clockfly updated

clockfly:
+1

PR #1979, opened 2016-02-25
fix #1928, fix verbose logging

### LINKED ISSUE #1928: command option -verbose does not have effect on output
Whether it's true or not, the outputs of the command have no difference.

manuzhang:


clockfly:
Maybe we should turn on log level "DEBUG" for -verbose flag.

Set the default level to INFO is still essential. For example, if I want to log the application Id and show this on screen, I cannot use "ERROR", as there is no error happening.

manuzhang:
1. If only for application id, we can print it. 
2. If info is really essential, is it too "verbose" ? We may need change some info logs to debug

manuzhang:
typical INFO output at client side

```
[ClusterConfig$] loading config file application.conf...
[Slf4jLogger] Slf4jLogger started
[Remoting] Starting remoting
[Remoting] Remoting started; listening on addresses :[akka.tcp://client1931912946@127.0.0.1:45102]
[Metrics$] Metrics is enabled...,  true
[ClientContext] Starting system client1931912946
[MasterProxy@masterproxy829808872] Contacts point URL: akka.tcp://master@127.0.0.1:3000/user/master
[MasterProxy@masterproxy829808872] sending identity to ActorSelection[Anchor(akka.tcp://master@127.0.0.1:3000/), Path(/user/master)]
[MasterProxy@masterproxy829808872] Master Proxy is started...
[MasterProxy@masterproxy829808872] Connected to [akka.tcp://master@127.0.0.1:3000/user/master]
[MasterProxy@masterproxyclient1931912946] Contacts point URL: akka.tcp://master@127.0.0.1:3000/user/master
[ClientContext] Creating master proxy Actor[akka://client1931912946/user/masterproxyclient1931912946#603111233] for master list: List(HostPort(127.0.0.1,3000))
[MasterProxy@masterproxyclient1931912946] Master Proxy is started...
[MasterProxy@masterproxyclient1931912946] sending identity to ActorSelection[Anchor(akka.tcp://master@127.0.0.1:3000/), Path(/user/master)]
[MasterProxy@masterproxyclient1931912946] Connected to [akka.tcp://master@127.0.0.1:3000/user/master]
Submit application succeed. The application id is 4
[ClientContext] Shutting down system client1931912946
```
Most of them are not needed for client and the application id is actually printed to console.

stanleyxu2005:
Set default logging level to `INFO` is okay. `ERROR` as default logging level is too high. For production environment, the admin will choose their preferred logging level. I prefer to reject this PR. I think one more important thing is that we log messages in proper logging level. E.g. not use warning for an error.

whjiang:
I agree with @stanleyxu2005. We shall keep log level as INFO. But, we need to change some message to DEBUG level to reduce logs shown. And having a cleaning but meaningful log for client is quite important. 

So, @manuzhang, please re-implement your PR in this direction.

manuzhang:
It's not all client log. MasterProxy logs may not be meaningful for client but they are also in other interactions with master. There are also akka logs. Closed this for now till we find a better solution.

PR #1990, opened 2016-03-04
fix #1989: add confirmation dialog for kill app and restart app

### LINKED ISSUE #1989: Add confirmation before doing irreversible operations
Propose to have a confirmation for irreversible operations, e.g. kill app restart app, so that we won't interrupt running applications by accidental click.

stanleyxu2005:
preview
Update 1: The buttons are now "OK" and "Cancel".


kkasravi:
+1 as long as we're consistent with 'Yes' 'No, Thanks' on other confirmation dialogs

PR #1995, opened 2016-03-06
[fix:1994]unified external package name

darionyaphet:
In external modules , I found HBase's package is difference with other's .

I try to update it package name from io.gearpump.external.hbase to io.gearpump.streaming.hbase , make external package names are the same :dancer:

### DIFF
--- a/external/hbase/src/main/scala/io/gearpump/streaming/hbase/HBaseSink.scala
+++ b/external/hbase/src/main/scala/io/gearpump/streaming/hbase/HBaseSink.scala
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.gearpump.external.hbase
+package io.gearpump.streaming.hbase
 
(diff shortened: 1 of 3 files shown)


whjiang:
+1.

manuzhang:
I wonder whether the HBase package is streaming specific or more general @huafengw ?

huafengw:
I think what we want is a general connector, not only for streaming applications. 
However, the DSL part is actually for streaming, so we should discuss how to reorganize the package names in external connector projects.

darionyaphet:
The package named with `io.gearpump.streaming` looks like a part of `streaming` module . Using another name maybe better ?

clockfly:
@darionyaphet 

This class also supposed to used by some non-streaming case. May not changing it for now?

darionyaphet:
@clockfly I think it's OK and closed this pull request . 

If we want to change package name I will create a issue at [GEARPUMP](https://issues.apache.org/jira/browse/GEARPUMP)

PR #2012, opened 2016-03-23
fix #1641, add exactly-once it

### LINKED ISSUE #1641: add exactly-once spec to integration-test
@stanleyxu2005 please help to add a HDFS docker image like that for Kafka

manuzhang:
1. add exactly-once integration test using `MessageCountApp` example .
2. add `sequenceiq/hadoop-docker:2.6.0` image which is required for checkpoint storage.
3. `MessageCountApp` is turned into `KafkaSource ~> Count ~> KafkaSink` pipeline such that we are able to feed input and verify output through Kafka.
4. `CountProcessor` counts the number of input messages and output the count after checkpoint. If the final count is correct on system failures, then exactly-once message delivery is guaranteed.
5. `PersistentTask` periodically sends a `CHECKPOINT` message to try committing checkpoint. The benefits are 
        - avoid checking `upstreamMinClock` on each input message
        - able to check `upstreamMinClock` and try committing checkpoint even if there no more input messages
6. checkpoint is committed synchronously which will introduce performance impact. I'd like to fix it in the following up PR.
7. it's hard-coded to output state after checkpoint. Another follow up is to provide users with a trigger API.


clockfly:
+1

PR #2014, opened 2016-03-24
GEARPUMP-24, refactor DataSource API

### LINKED ISSUE #2013: refactor DataSource API
This is the current DataSource API 

```scala

trait DataSource extends java.io.Serializable {

  /**
   * open connection to data source
   * invoked in onStart() method of [[io.gearpump.streaming.source.DataSourceTask]]
   * @param context is the task context at runtime
   * @param startTime is the start time of system
   */
  def open(context: TaskContext, startTime: Option[TimeStamp]): Unit

  /**
   * read a number of messages from data source.
   * invoked in each onNext() method of [[io.gearpump.streaming.source.DataSourceTask]]
   * @param batchSize max number of messages to read
   * @return a list of messages wrapped in [[io.gearpump.Message]]
   */
  def read(batchSize: Int): List[Message]

  /**
   * close connection to data source.
   * invoked in onStop() method of [[io.gearpump.streaming.source.DataSourceTask]]
   */
  def close(): Unit
}
```

which has several issues

1. `read` returns a scala list of Message which is unfriendly to Java DataSources. Same for `Option` parameter in `open`
2. the number of read messages may not be the same as the passed in `batchSize` which leaves uncertainty to users (users may access out of boundary list positions)
3. to return a list an extra buffer could be needed in `read` (e.g. `KafkaSource`) which is not best for performance

manuzhang:
please refer to #2013 for the problem description.

### DIFF
--- a/streaming/src/main/scala/io/gearpump/streaming/source/DataSource.scala
+++ b/streaming/src/main/scala/io/gearpump/streaming/source/DataSource.scala
@@ -29,10 +29,14 @@ import io.gearpump.{TimeStamp, Message}
  * {{{
  *  GenStringSource extends DataSource {
  *
- *    def open(context: TaskContext, startTime: Option[TimeStamp]): Unit = {}
+ *    def open(context: TaskContext, startTime: TimeStamp): Unit = {}
  *
- *    def read(batchSize: Int): List[Message] = {
- *      List.fill(batchSize)(Message("message"))
+ *    def advance(): Boolean = {
+ *      true
+ *    }
+ *
+ *    def read(): Message = {
+ *      Message("message")
  *    }
  *
  *    def close(): Unit = {}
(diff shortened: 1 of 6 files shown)


clockfly:
Looks you removed batching. Generally batching have performance advantage, have you done some micro-benchmark on this?

manuzhang:
jira link https://issues.apache.org/jira/browse/GEARPUMP-24

streaming/src/main/scala/io/gearpump/streaming/source/DataSource.scala
```diff
@@ -29,10 +29,14 @@ import io.gearpump.{TimeStamp, Message}
  * {{{
  *  GenStringSource extends DataSource {
  *
- *    def open(context: TaskContext, startTime: Option[TimeStamp]): Unit = {}
+ *    def open(context: TaskContext, startTime: TimeStamp): Unit = {}
  *
- *    def read(batchSize: Int): List[Message] = {
- *      List.fill(batchSize)(Message("message"))
+ *    def advance(): Boolean = {
```
whjiang: I don't get why we need this API?

Can a read() automatically step forward the cursor? If yes, maybe a better naming function for this is "hasData()".

manuzhang: if no `advance` is called in between, multiple `read` will return the same record

manuzhang:
keep it open till we finish migrating to the new repo

PR #2017, opened 2016-03-25
fix #1988, upgrade akka to akka 2.4.2

### LINKED ISSUE #1988: Add cross build and upgrade to akka 2.4.2 and off akka-stream-experimental

clockfly:
**Changes:**
1. Upgrade akka to akka 2.4.2, 
2. Upgrade akka-data-replication to akka-distributed-data-experimental 2.4.2
3. Upgrade akka-http-testkit-experimental to akka-http-testkit 2.4.2.
4. Upgrade scala from 2.11.5 to 2.11.8
5. Drop support for scala 2.10 as akka 2.4.2 no longer support it.
6. ~~Drop scalajs support, as it is not used, and we probably won't use it. Pure javascript may be more mature for us for frontend.~~
7. Disable project "gearpump-experiments-akkastream" temporary, will add it back using akka-http-stream 2.4.2 in another PR.
8. Drop the hotfix #1816. As the issue is fixed in akka 2.4.2
9. Drop support for JDK6 and JDK7, as akka 2.4.2 requires JDK8. And the upgrade from JDK6 to JDK8 should be smooth.
10. Scala 2.12 is not supported as it is not GA yet.
11. Update the travis build script with JDK8.

**Compatibility**
- [x] Make sure old gearpump job jar can run on new Gearpump version. 
Simple verification shows it is compatible. Still need more thorough test.

**Performance**
- [x] Make sure there is no performance regression, by updating from akka 2.3.12 to akka 2.4.2

Update: Scalajs is not removed in this PR.


manuzhang:
the fix (https://github.com/akka/akka/pull/19876) for #1816 is merged 3 days ago. Are you sure it is fixed in akka 2.4.2 ?

core/src/main/resources/geardefault.conf
```diff
@@ -575,6 +576,14 @@ akka {
   actor {
     provider = "akka.remote.RemoteActorRefProvider"
 
+    ## Doesn't warn on Java serializer usage
+    ##
+    ## Most of our streaming message are using custom serializer, wit a few
```
manuzhang: typo `wit` ?

clockfly:
@manuzhang 

akka/akka#19876 is about AbstractBoundedNodeQueue, we use AbstractNodeQueue

manuzhang:
Ok, I'll do some performance verification for this PR.

clockfly:
@manuzhang  Thanks for helping in performance verification.

kkasravi:
> `6. Drop scalajs support, as it is not used, and we probably won't use it. Pure javascript may be more mature for us for frontend.`

-1. We should attempt to keep this intact since 'serverless' deployment is a viable option for gearpump to leverage as part of distinguishing itself from other frameworks. https://www.syncano.io/, AWS lambda, IBM OpenWhisk are examples of serverless frameworks.
A good article is http://underscore.io/blog/posts/2016/03/21/serverless-scale-summit.html

clockfly:
@kkasravi  Thanks, I will looks into the links.

kkasravi:
+1 on other changes noted above (other than 6.)

clockfly:
@kkasravi 

Add the scalajs back. However, there are still some performance issue. We are trying to figure out whether it is a akka 2.4 bug or not.

stanleyxu2005:
Lgtm +1 and yes +1 for keeping scalajs

huafengw:
```
[error] (gearpump-servicesJS/*:update) sbt.ResolveException: unresolved dependency: org.scala-js#scalajs-compiler_2.11.8;0.6.4: not found
```

clockfly:
The performance issue is caused by #2020, when akka 2.4.2 upgrade netty from 3.9.3 to 3.10.3.

clockfly:
The "hang" issue is caused by netty version 3.10.3, in #2020.

After fixing this issue, we still see performance degradation of 20%. Still trying to figure out why.

clockfly:
More experiment on this:

When downgrading the netty version to 3.8.0.Final (same version as akka 2.3.9), the performance doesn't drop in noticeable way.
And http://akka.io/news/2015/09/30/akka-2.4.0-released.html, says that it is safe to downgrade netty version to 3.8.0.

So, we will use netty 3.8.0.Final, will work out a new commit soon.

clockfly:
**Performance verification:**
Env:
4 machines, 16 workers, 16 executors, SOL benchmark (48 upstream tasks, 48 downstream tasks), message size: 100 bytes.

Result:
About 17.1 million message/second.
Baseline: 18 million message/second.
screenshot from 2016-03-28 21-12-13

**Compatibility verification:**
By http://doc.akka.io/docs/akka/2.4.0/project/migration-guide-2.3.x-2.4.x.html?_ga=1.94913044.9280276.1458718679, seems akka take care the compatibility very well.

Also manual verification shows the new version is compatible with job jars compiled against old Gearpump version.

clockfly:
Can someone recheck the patch?

manuzhang:
+1. Shall we report the netty issue to akka community ?

clockfly:
@manuzhang 

Since we use our customized netty transport, instead of built-in akka netty transport. Netty transport may not have the same problem like this one.  

This netty issue is triggered when using channel.isWritable to do flow control. I did a rough search in akka code base, only find the akka UDP transport ( https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/transport/netty/UdpSupport.scala)

I think it is not easy to trigger this issue in Akka transport directly.

PR #2018, opened 2016-03-25
fix #2006, Try streaming kmeans on Gearpump

### LINKED ISSUE #2006: Try streaming kmeans on Gearpump
There is a document about streaming kmeans in Spark (https://databricks.com/blog/2015/01/28/introducing-streaming-k-means-in-spark-1-2.html), I think we can try to implement it on Gearpump. Here is my processor topology on Gearpump:
kmeans

The `Source Processor` will produce points by time, then broadcast the point to the `Distribution Processor`. The number of tasks of the `Distribution Processor` is k, where each task save one center and the corresponding points. When `Distribution Processor` receives a point from `Source Processor`, it will calculate the distance of this point to its center, and then send the distance along with the point and its `taskId` to the `Collection Processor`. When `Collection Processor` receives the distance from `Distribution Processor`, it will accumulate the number of current points, determine if it's time to update center, choose the smallest distance and then send the point along with its corresponding `Distribution Processor` taskId by broadcast partitioner. When `Distribution Processor` receives the result message, task with the corresponding `taskId` will accumulate the point. If `Distribution Processor` receives that it's time to update center, then all the tasks will update its corresponding center.

This procedure is streaming and the center of cluster will change by time.

pangolulu:
I will give UT then.

### DIFF
--- a/examples/streaming/streamingkmeans/src/main/scala/io/gearpump/streaming/examples/streamingkmeans/RandomRBFSource.scala
+++ b/examples/streaming/streamingkmeans/src/main/scala/io/gearpump/streaming/examples/streamingkmeans/RandomRBFSource.scala
@@ -1,0 +1,117 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.gearpump.streaming.examples.streamingkmeans
+
+import io.gearpump.util.LogUtil
+import io.gearpump.{TimeStamp, Message}
+import io.gearpump.streaming.source.DataSource
+import io.gearpump.streaming.task.TaskContext
+import org.slf4j.Logger
+
+import scala.util.Random
+
+object RandomRBFSource {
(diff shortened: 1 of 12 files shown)


manuzhang:
@pangolulu great work! Could you add a README to illustrate how to run this example and what's the expected behaviors ?

examples/streaming/streamingkmeans/src/main/resources/application.conf
```diff
@@ -0,0 +1,6 @@
+gearpump {
+  serializers {
+    "io.gearpump.streaming.examples.streamingkmeans.InputMessage" = ""
+    "io.gearpump.streaming.examples.streamingkmeans.ResultMessage" = ""
+  }
+}
```
huafengw: Try to change this config file name to geardefault.conf, then I think user does not need to put the conf file to folder ```gearpump/conf``` any more.

examples/streaming/streamingkmeans/src/main/scala/io/gearpump/streaming/examples/streamingkmeans/RandomRBFSource.scala
```diff
@@ -0,0 +1,117 @@
...
+  private val LOG: Logger = LogUtil.getLogger(classOf[RandomRBFSource])
+}
+
+/**
+ * RandomRBFGenerator generates via radial basis function.
+ * Reference by https://github.com/huawei-noah/streamDM
+ */
+class RandomRBFSource(k: Int, dimension: Int) extends DataSource {
+
+  class Centroid(center: Array[Double], classLab: Int, stdev: Double) {
```
manuzhang: case class ?

manuzhang:
@pangolulu how to test the model ?

pangolulu:
@manuzhang Now we only can use this app to get the cluster centroids over time, but it does not support evaluate the cluster centroids in streaming mode. You can write the cluster centroids into file system, and evaluate it offline using some cluster evaluation metrics.

manuzhang:
@pangolulu From my limited ML knowledge, I don't think it is useful if we cannot test our model. It will be great if our user can know

1. what do the train data and test data look like ? distributions?
2. how do we train our model ? 
3. what's the performance of the model against some baseline ? (SAMOA prints that in log, I think)

manuzhang:
please close this and open a pr at https://github.com/apache/incubator-gearpump/pulls

Issue #1977, opened 2016-02-24
REST API for submitting DAG


karol-brejna-i:
1. 
Couldn't find description of endpoint for submitting GP application (http://www.gearpump.io/releases/latest/dev-rest-api.html)

2. 
I've noticed that the arguments ('arguments' field in submit form) are passed as URL encoded value which causes problems for longer values
Example: 
http://demo.gearpump.io/api/v1.0/master/submitapp?args=param1%20param2%20param3%20fdsafads%20param5

Can we pass them as field in the form? (So we avoid request size limit problem)

kkasravi:
@karol-brejna-i 

```
1. Couldn't find description of endpoint for submitting GP application (http://www.gearpump.io/releases/latest/dev-rest-api.html)
```
We haven't formally documented this REST endpoint. See #1849.

```
2. I've noticed that the arguments ('arguments' field in submit form) are passed as URL encoded value which causes problems for longer values
Example: 
http://demo.gearpump.io/api/v1.0/master/submitapp?args=param1%20param2%20param3%20fdsafads%20param5

Can we pass them as field in the form? (So we avoid request size limit problem)
```
My recommendation is we include them in the JSON payload so we avoid the application/x-www-form-urlencoded mime-type. @stanleyxu2005 may have a preference.

stanleyxu2005:
I think changing it as a form field would be a better choice. +1

kkasravi:
In MasterService.scala we need to use formFields instead of parameters see http://doc.akka.io/docs/akka/2.4.2/scala/http/routing-dsl/index.html
For submit.js we need to provide a data field I believe to get the args sent as a form-field - https://github.com/danialfarid/ng-file-upload

@stanleyxu2005 would you be able to make these changes? Otherwise I'll make the changes and submit PR. I believe @karol-brejna-i needs the changes sometime Monday.

stanleyxu2005:
The `ng-file-upload` I'm using has some difficulty to understand form field together with upload files. What concrete use case are you facing that cannot pass args in the url?

kkasravi:
Hi @stanleyxu2005 

It does look like the latest version will help based on documentation I see under the `Upload Service`

https://github.com/danialfarid/ng-file-upload

```
  /*
  Specify the file and optional data to be sent to the server.
  Each field including nested objects will be sent as a form data multipart.
  Samples: {pic: file, username: username}
    {files: files, otherInfo: {id: id, person: person,...}} multiple files (html5)
[...]

PR #2023, opened 2016-03-29
fix GEARPUMP-2, Define REST API to submit job jar

clockfly:
The Github/Apache Jira integration is not complete yet, here is the jira link:
https://issues.apache.org/jira/browse/GEARPUMP-2

The original discussion can be found at: https://github.com/gearpump/gearpump/issues/1977

Changes:
1. Document how to submit an application jar by REST api.
2. Allow user to compose the config that need to be submitted in javascript. This is done by adding an additional field "configstring".

### DIFF
--- a/services/dashboard/services/restapi.js
+++ b/services/dashboard/services/restapi.js
@@ -118,9 +118,9 @@ angular.module('dashboard')
         },
 
         /** Submit an user defined application with user configuration */
-        submitUserApp: function(files, formFormNames, executorNum, args, onComplete) {
+        submitUserApp: function(files, fileFieldNames, executorNum, args, onComplete) {
           return self._submitApp(restapiV1Root + 'master/submitapp',
-            files, formFormNames, executorNum, args, onComplete);
+            files, fileFieldNames, executorNum, args, onComplete);
         },
 
         /** Submit a Storm application */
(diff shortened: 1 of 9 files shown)


services/dashboard/services/restapi.js
```diff
@@ -129,13 +129,16 @@ angular.module('dashboard')
...
+        _submitApp: function(url, files, fileFieldNames, executorNum, args, onComplete) {
           var upload = Upload.upload({
-            url: url + params,
+            url: url,
             method: 'POST',
             file: files,
-            fileFormDataName: formFormNames
+            fileFormDataName: fileFieldNames,
+            fields: {
+              "executornumber": executorNum,
```
stanleyxu2005: Maybe the correct name would be `executorCount`. And please update all occurrence, if the rename makes sense to you.

services/jvm/src/main/scala/io/gearpump/services/MasterService.scala
```diff
@@ -188,6 +194,22 @@ class MasterService(val master: ActorRef,
...
     }
   }
+
+  private def mergeConfig(configFile: Option[File], configString: String): Option[File] = {
+    if (configString == null || configString.isEmpty) {
+      configFile
+    } else {
+      configFile match {
+        case Some(file) =>
+          Files.write(file.toPath, ("\n" + configString).getBytes(UTF_8), APPEND)
```
stanleyxu2005: config file is HOCON and config string is JSON. Does appending really work?

services/dashboard/services/restapi.js
clockfly: It make sense, thanks.

services/jvm/src/main/scala/io/gearpump/services/MasterService.scala
clockfly: config string is also in HOCON.

And it works even if we directly combine HOCON + JSON.

stanleyxu2005:
Lgtm +1

clockfly:
@karol-brejna-i 

The doc style has been fixed.

whjiang:
Are we still using github/gearpump/gearpump repo?

clockfly:
@whjiang 

The code import to Apache git is still in process (blocked at Apache infrastructure), it is tracked by https://issues.apache.org/jira/browse/INFRA-11435

clockfly:
Before the code import is complete, we still use the current repo.

PR #2024, opened 2016-03-30
GEARPUMP-3, Define REST API to add/remove worker instances, which allow us to scale out in YARN

clockfly:
Motive:
In YARN environment, it will convenient if there is a REST API which allow us to start new worker instance. This allows us to scale out with more machines.

Jira ticket: https://issues.apache.org/jira/browse/GEARPUMP-3


stanleyxu2005:
Lgtm +1

PR #2029, opened 2016-04-04
GEARPUMP-10, Downgrade netty from Netty 4 to Netty 3.8 cause the OAuth2 authentication failure

clockfly:


stanleyxu2005:
Lgtm +1

PR #2032, opened 2016-04-11
GEARPUMP-11, fix code style

clockfly:
Patch: https://github.com/gearpump/gearpump/compare/master...clockfly:fix_style.patch


clockfly:
**Changes:**

1. Fix most compiler warnings (except scala  inline warning)
2. Enable java doc building for project core, streaming, and all external projects like kafka.
3. Fix Scaladoc and Javadoc compilation errors when updating java to jdk8, which has a more strict javadoc check.
4. Add scalastyle check, the file is scalastyle-config.xml. 
5. Enforce scalastyle check and scaladoc/javadoc check when submitting a PR.
6. Don't allow using scala postfix syntax, for example "3 seconds" is not allowed now, please use 3.seconds instead.
7. Disable using "scala.collection.JavaConversions", as it does automatic conversion between java and scala, and encourage unnecessary conversions.
8. Remove all unicode chars.
9. Limit the line length to 100 chars.
10. Use standard javadoc syntax. which is 
  ```
General description (In third person, Cap-first-letter. Instead of "Do", use "Does". Discourage using "will", "we" )

More Description

@param a Cap-letter description1
@param a Cap-letter description2
@param a Cap-letter description3
@return a Cap-letter description4
  ```
11. Discourage using of special scaladoc syntax. (As it is not friendly when generating javadoc from scala source code)
12. The importing statement should be grouped:

  ```scala
java imports
---empty line---
scala imports
---empty line---
third party imports
---empty line---
io.gearpump._ imports
  ```
13. Update sbt to sbt 0.13.11, as it fixed many bugs.
14. Update Akka to Akka 2.4.3 as it fixed a Akka-http warning issue (which keep printing "error connection not closed in 1 minutes")
15. Update "com.softwaremill.akka-http-session" to "0.2.5"
16. Separate scalajs compiling with scala compiling. (Scalajs need to be compiled separately)
17. Add more scaladoc/javadoc comments for public class and methods.
18. Update bijectionVersion to 0.8.0

**About code style**
We will use https://github.com/databricks/scala-style-guide as the foundation, with a few exceptions:
1. We still allow using ! and ? when sending akka message.
2. We still allow using System.currentTimeMillis

**TODO:**
Still need to fix the visibility issue. All class inside Gearpump which is not explicitly documented in gearpump.io should be considered private. We need to enforce it in code.

clockfly:
This PR is mostly about fixing code style. In order to get code style fixed, it has to fix some additional problems or doing some library upgrade.

manuzhang:
Thanks for taking care of this tedious but important work.

A question about "jvm.sbt" (I'm not able to comment on the hidden diffs since there are too many change files)

```
// Treat warnings as erros.
scalacOptions in(ThisBuild) ++= 
  // Show more detailed warnings.
  Seq("-unchecked", "-feature", "-deprecation", "-Yinline-warnings",
    // Disable postfix, and other advanced scala features,
    "-language:-dynamics", "-language:-postfixOps", "-language:-reflectiveCalls", "-language:-existentials"
  )
```

Which flag is used to "treat warnings as errors ? (p.s. typo "erros")

clockfly:
@manuzhang 

"treat warnings as errors" is NOT enabled, as the scala @inline warnings  are commonly seen, and cannot be suppressed. 

You can view the patch file: https://github.com/gearpump/gearpump/compare/master...clockfly:fix_style.patch

It is a complete diff.

manuzhang:
How about some other import rules from https://github.com/databricks/scala-style-guide#imports:

1. Avoid using wildcard imports, unless you are importing more than 6 entities, or implicit methods. Wildcard imports make the code less robust to external changes.
2. Always import packages using absolute paths (e.g. scala.util.Random) instead of relative ones (e.g. util.Random).

stanleyxu2005:
This is a very useful and important PR. +1
i'm curious have you changed everything with IntelliJ or using some static check tool.

clockfly:
@manuzhang 

Most obvious code style problems are fixed. With these exceptions:
1. Most monadic chaining problems are not fixed. Modifying chaining need deep inspection on code logic.
2. Symbolic Methods of built-in scala libraries are NOT fixed.. 
3. Implicit definition that are NOT obvious and safe to remove are NOT fixed..
4. Exception Handling. We still use try catch(ex Exception) for now. It is safe for us,  as we removed all "return" statement.
5. User API fixes like varargs (It is not easy and requires deep inspection)
6. Importing fix. We also allows user to import wildcard elements from companion object.

However, for new code, we should follow the code style at https://github.com/databricks/scala-style-guide#imports, with modification like this:

1. Still allow using ! and ?
2. Still allow using System.currentTimeMillis
3. Still allow importing fields defined in companion object.

kkasravi:
+1 though I think monadic chaining and implicit conversions are ok in a number of circumstances. Same question as @stanleyxu2005 - is there a way to import these rules into intellij?

clockfly:
@kkasravi 

The scalastyle-config.xml can be imported to Intellij. Steps:
1. Copy scalastyle-config.xml to .idea/scalastyle_config.xml
2. Config Intellij Inspection at: Settings/Inspections/Scala/Code Style/Scala style inspection.

PR #2124, opened 2019-01-01
Remove distribute service example

manuzhang:
The example no longer works with latest dependencies. It's archived at https://github.com/apache/incubator-gearpump/tree/master/examples/distributeservice


build.sbt
```diff
@@ -30,7 +30,6 @@ lazy val aggregated: Seq[ProjectReference] = Seq[ProjectReference](
   packProject,
   complexdag,
   distributedshell,
-  // distributeservice,
```
domsj: should also remove `lazy val distributeservice = ...` on line 183 below?

manuzhang:
@domsj  Thanks, merged

